PR #71, opened 2019-11-07
Bugfix/fix unit tests #69

### LINKED ISSUE #69: Unit tests on fresh install of master don't work
Doing a clean install of the repository currently causes the unit tests to fail - I believe this is due to packages mismatching/updating beneath our feet. I believe this is a problem due to the fact we are currently using `lerna bootstrap --hoist` which installs all the packages in the `node_modules` at the top level. Whilst this decreases redundancy, this seems to be causing issues and lerna itself does say that hoisting can break tooling. However, certain packages still need to be hoisted due to things breaking if `datagateway-common` and other packages don't share the same library.

Acceptance criteria:
- [ ] A clean install no longer has failing unit tests

louise-davies:
## Description
We now only hoist packages we need to hoist, namely react and react-router. Plus there are some changes to the lock files and fixing of some tests.

Additionally, due to the addition of Travis, this also fixes some issues that were causing Travis builds to fail.

## Testing instructions
- [x] Review code
- [x] Clone a fresh repo and switch to this branch and try installing + running the tests

## Agile board tracking
Fixes #69 and part of #73


esl-lewis:
On a fresh install when running tests from inside the /packages directories all of them except datagateway-search pass. This will be resolved in issue #74 . 
However running npm test from higher in the directory in the datagateway folder fails with a lerna error: "'lerna' is not recognized as an internal or external command,
operable program or batch file.

GoelBiju:
Like @esl-lewis mentioned npm test works fine inside the datagateway-table directory, however not at root level.

louise-davies:
@GoelBiju @esl-lewis This was due to the lerna package being overwritten because it wasn't explicitly being hoisted alongside react and react-router. I have now hoisted all the top-level dependencies as well as extracted out said config to the lerna.json file. This also resolves the issue with the install command "failing" - I think this was also due to lerna being uninstalled during the installation process.

GoelBiju:
@louise-davies I have tested the branch out on a fresh clone of the repository and the unit tests are working fine now.

Issue #88, opened 2019-11-19
datagateway-download needs a settings file


louise-davies:
Currently there is no setting file which means there is no way of supplying things like URLs and facility names other than hardcoded. Currently, datagateway-download doesn't use Redux (and I don't think it will ever have to), so a config object will be needed and perhaps we can use [React Context API](https://reactjs.org/docs/context.html) to supply it to child components.

Acceptance criteria:
- [x] datagateway-download has a settings file that can be loaded and used to configure the plugin

GoelBiju:
@louise-davies this can be worked on once #48 has been completed. Edited your issue to add in a link to the React Context API for explanation on its usage.

GoelBiju:
@louise-davies would it be worth adding an `accessMethods` array or object to the `datagateway-download-settings` file?

louise-davies:
Yes, it's where it needs to be defined. See https://github.com/icatproject/topcat/blob/master/yo/app/config/topcat.json.example#L594 to see that this is where TopCAT defines it. It also defines an IDS url per download type, which is definitely used by diamond (globus downloads are handled by a separate IDS).

GoelBiju:
- [ ] Add in support for access methods in `datagateway-download-settings.json` and accept custom idsUrl for submitting and making downloads,

- [ ] Replace all the remaining hardcoded settings,

- [ ] Add in unit tests for ConfigProvider.

GoelBiju:
Tests have been updated to reflect the addition of the `datagateway-download-settings.json` in https://github.com/ral-facilities/datagateway/commit/ff6355e1dbb2df62208fe838b282a8ce588a7caa. The `ConfigProvider` tests still need to be added and there are currently warnings stating due to not wrapping the `DownloadConfirmDialog` with the `ConfigProvider` as the settings are not passed on. The tests work fine without it however the console warnings persist, if we wrapped the `DownloadConfirmDialog` with the `ConfigProvider` the [`setProps`](https://github.com/ral-facilities/datagateway/blob/master/packages/datagateway-download/src/downloadConfirmation/downloadConfirmDialog.component.test.tsx#L328) call would fail as it is not the root component.

GoelBiju:
All remaining unit tests for ConfigProvider have been added in https://github.com/ral-facilities/datagateway/commit/e47a8d3dd73def09d5a93d587cad4bca1006e14c.

GoelBiju:
[...]

PR #89, opened 2019-11-19
Feature/download cart #47

### LINKED ISSUE #47: [datagateway-download] Create "cart" view
This is a large issue that may get split up into smaller issues.

Recreate the TopCAT download cart. Unlike TopCAT which has it as a modal, we should instead have it as a separate page accessed via the menu (`Data/Download`). I think we shall need two "tabs" - one for the download cart and one for the download status page.

Cart mockup (based on current TopCAT):

cart

When the user clicks download, they should then be taken to the download confirmation page

confirm download

Acceptance criteria:
- [ ] Have a download cart that displays the datafiles/datasets the user has selected
- [ ] The user can remove individual cart items or empty the whole cart
- [ ] When the user clicks download cart they are taken to the confirm download page
- [ ] The cart displays info on the total number of files and size of files in the cart, with maximums for Diamond

louise-davies:
## Description
Adds the download cart view that shows the user the items they currently have in their cart and gives them the opportunity to remove items or to proceed and download their cart.

Currently, the URLs required (topcat and datagateway-api) are not configurable since there is no settings file yet - instead this is being tracked in a separate issue: #88

## Testing instructions

- [x] Review code
- [x] Check Travis build

## Agile board tracking
Closes #47


esl-lewis:
Some minor UI suggestions, I expected more feedback from the remove dataset button. Maybe it could flash a different colour when pressed.

esl-lewis:
I ran the general npm tests for both dg-download and dg-common, these passed fine. But when I ran cypress interactively a few tests failed- more specifically when trying to access a row by aria label in 'should load correctly and display cart items' and 'should be able to sort cart items by name and type'. 
I assume this is due to the authentication issue we discussed earlier and 'npm test' works fine so I have approved anyway.

packages/datagateway-download/src/downloadCart/downloadCartApi.ts
```diff
@@ -0,0 +1,190 @@
...
+        items: `${entityType} ${entityId}`,
+      },
+    })
+    .then(() => {})
+    .catch(error => {
+      log.error(error.message);
+    });
+};
+
+// TODO: caching???
```
esl-lewis: Do you want this in with this commit?

louise-davies:
@esl-lewis 

- sort by Size
This could probably be implemented, the only reason it wasn't is 1. because TopCAT doesn't and 2.  because this value is dynamically loaded and so there would have to be some handling of disabling until everything has attempted to load / dealing with failed cases. I'll see how simple it is to add.

- details panels
This one is more tricky - we aren't actually loading the information on the entities here - the TopCAT cart stores the entity type and entity name, and we query for the size, but any other information would also have to be queried. This could be done, but I would rather wait for user feedback/make it a separate feature request before performing work for a function that might not be wanted by users.

- UI feedback upon item removal
I'll have a go at delaying the removal for a bit so that I can give more visual feedback

louise-davies:
@esl-lewis 
Here's a demo of adding a small delay and changing the colour of the icon when the item is being removed - does this feel better?

remove-cart-color-change

packages/datagateway-download/src/downloadCart/downloadCartApi.ts
louise-davies: I've removed this comment, we think caching in TopCAT was a solution for when getSize calls were expensive - so we'll only look into caching if we observe performance issues.

louise-davies:
@esl-lewis I've implemented the changes you suggested in your last review - could you please review this again?

louise-davies:
@esl-lewis This uses the `formatBytes` method, which currently always rounds to 2 decimal places. There's a trade-off of accuracy and information overload. For that specific figure, rounding is a concern since it is presenting a limit to the user and so displaying inaccurate information may cause issues in that the UI may show as being less than the limit but the actual number is greater than. It's not a difficult issue to fix if we get user feedback that we need variability on how many decimals/significant figures we should round to.

esl-lewis:
@louise-davies fair enough, I guess you wouldn't want to display inaccurate information.

PR #102, opened 2019-12-16
Improve e2e tests #97

### LINKED ISSUE #97: Improve stability and concurrency of e2e tests
Currently, the e2e tests are a little unstable, especially the new cart features (since these modify data on the remote server). Perhaps, rather than having the add to cart tests in each table test it is worth extracting them out into one spec file, so the tests that modify data are ran all in one go and we can control the concurrency issues a little better.

Perhaps adding the ability to retry tests once or twice in CI is also worth it.

Additionally, it might be worth separating the e2e job on Travis into multiple jobs for each plugin as this would increase speed and improve the ability to retry just one set of e2e tests. Caution must be made though on making sure the tests don't cause each other to fail when ran in parallel.

Acceptance criteria:
- [ ] Builds on Travis are more stable
- [ ] Potentially builds on Travis are faster

louise-davies:
## Description
I added longer waits to cart related actions, and separated the e2e tests for cart selection to their own spec file, as this simplifies the rest of the other tests and collects all of the "flaky" tests into one file. I still get occasional failures both locally and on CI for reasons I can't fathom - these are occurring less often than before, but to help with this I have added a retry plugin to Cypress and the tests in `cartSelection.spec.ts` are retried after they fail, so this should stabilise the Travis build.

Additionally, I have split the e2e tests into three separate jobs on Travis which allows them to run in parallel - which makes them faster. Also, due to the linting and unit tests being in their own stage, the e2e tests will only run after those two jobs have passed, which means builds will fail earlier if there are linting or unit test problems.

I have reran the build for this branch multiple times and not had the build fail, which hopefully indicates a more stable build.

## Testing instructions
- [x] Review code
- [x] Check Travis build

## Agile board tracking
Closes #97


louise-davies:
@esl-lewis I think this may have been due to a server hiccup or something. If this is a persistent problem then we can investigate when the issue occurs, but otherwise there isn't much for me to go off of.

PR #109, opened 2020-01-27
Add disableGlobal parameter to createGenerateClassName in packages (#108)

### LINKED ISSUE #108: Versions of MaterialUI clash when plugins are loaded in parent application
**Description**
Due to two instances of material-ui running when a plugin is loaded onto the parent application, there is a clash of styling that occurs, resulting in a poorly styled page. 

The main reason is due to the parent application being upgraded to `material-ui v4` and with both the parent and plugin applications using the same `material-ui` major versions there is a clash with the way the class names are being generated, mentioned by @louise-davies .

In order to solve this all plugins can be forced to use the `disableGlobal` set to true when calling `createGenerateClassName` to get deterministic styles. See MaterialUI [Advanced](https://material-ui.com/styles/advanced/#with-material-ui-core) for further information on this.

**Acceptance criteria:**

- [ ] `material-ui` versions do not clash and the styling on the parent application with two instances of material-ui running is correct.

GoelBiju:
## Description
Adds in the `disableGlobal` parameter for the `createGenerateClassName` function used for `StylesProvider` in all of the packages in datagateway.

## Testing instructions

- [x] Review code
- [x] Check Travis build

## Agile board tracking
Closes #108

### DIFF
--- a/packages/datagateway-download/src/App.tsx
+++ b/packages/datagateway-download/src/App.tsx
@@ -8,6 +8,7 @@ import {
 
 const generateClassName = createGenerateClassName({
   productionPrefix: 'dgwd',
+  disableGlobal: true,
 });
 
 class App extends Component<{}, { hasError: boolean }> {
--- a/packages/datagateway-download/src/App.tsx
+++ b/packages/datagateway-download/src/App.tsx
@@ -8,7 +8,11 @@ import {
 
 const generateClassName = createGenerateClassName({
   productionPrefix: 'dgwd',
-  disableGlobal: true,
+
+  // Only set disable when we are in production and not running e2e tests;
+  // ensures class selectors are working on tests.
+  disableGlobal:
+    process.env.NODE_ENV === 'production' && !process.env.REACT_APP_E2E_TESTING,
 });
 
 class App extends Component<{}, { hasError: boolean }> {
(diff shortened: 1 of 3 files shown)


GoelBiju:
> The e2e tests in datagateway-table are failing - this is due to some tests using class selectors. The sort tests can probably remove the assertions on classes - these are mostly asserting that the icon has changed, but we can still test that the column is sorted separately. The only other class selector I think is the Material Pickers day - this might be a little trickier to select without a consistent classname but it's probably possible.
> 
> Alternatively, we could always only add the disableGlobal property when process.env.NODE_ENV === `production` && !process.env.REACT_APP_E2E_TESTING. This would allow for us to select on classnames in e2e tests.

@louise-davies in https://github.com/ral-facilities/datagateway/pull/109/commits/43936cdf6adb10dcd82cdc8c483a238f649354a4 I have based the value of disableGlobal to be the check to see if we are in `production` and not e2e testing as you mentioned. I am not sure if we need to edit the tests since they only fail due to this one parameter.

GoelBiju:
> @GoelBiju it's fine that the tests haven't been modified - I was just giving options. The disabling for e2e tests solution is the easiest one ;)

@louise-davies it definately seemed like the quickest fix for the issue :)

PR #115, opened 2020-02-13
Use & parse scigateway:token

### LINKED ISSUE #111: Implemenent scigateway:token changes
Now that we have `scigateway:token` working nicely in SciGateway, the plugins should change to be able to utilise this token effectively.

Acceptance criteria:
- [x] Plugins parse the JWT body for things such as the ICAT session ID
- [ ] If plugins fail to contact ICAT due to authentication issues, they send an InvalidToken message to SciGateway

louise-davies:
## Description
Add utility functions that handle the parsing of `scigateway:token` and also a utility function that handles error handling so that we can invalidate tokens on authentication error responses.

## Testing instructions
- [ ] Review code
- [x] Check Travis build
- [ ] Set up `scigateway` with `scigateway-auth` and verify everything works fine in "prod" mode - note this is best to be done over HTTP (otherwise the plugins need to be ran over HTTPs, and thus `datagateway-api`...), which will require ral-facilities/scigateway-auth#43

## Agile board tracking
Closes #111

### DIFF
--- a/packages/datagateway-common/src/handleICATError.test.ts
+++ b/packages/datagateway-common/src/handleICATError.test.ts
@@ -1,0 +1,92 @@
+import { AxiosError } from 'axios';
+import * as log from 'loglevel';
+import handleICATError from './handleICATError';
+import { AnyAction } from 'redux';
+
+jest.mock('loglevel');
+
+describe('handleICATError', () => {
+  let error: AxiosError;
+  let events: CustomEvent<AnyAction>[] = [];
+
+  beforeEach(() => {
+    events = [];
+
+    document.dispatchEvent = (e: Event) => {
+      events.push(e as CustomEvent<AnyAction>);
+      return true;
+    };
+
+    error = {
+      isAxiosError: true,
+      config: {},
+      response: {
+        data: {},
+        status: 500,
+        statusText: 'Internal Server Error',
+        headers: {},
+        config: {},
+      },
+      name: 'Test error name',
+      message: 'Test error message',
+    };
+  });
+
+  it('logs an error and sends a notification to SciGateway', () => {
+    handleICATError(error);
+
+    expect(log.error).toHaveBeenCalledWith('Test error message');
+    expect(events.length).toBe(1);
+    expect(events[0].detail).toEqual({
+      type: 'scigateway:api:notification',
+      payload: { severity: 'error', message: 'Test error message' },
+    });
+  });
+
+  it('just logs an error if broadcast is false', () => {
+    handleICATError(error, false);
+
+    expect(log.error).toHaveBeenCalledWith('Test error message');
+    expect(events.length).toBe(0);
+  });
+
+  it('sends an invalidate token message to SciGateway on 403 response', () => {
+    if (error.response) error.response.status = 403;
+    handleICATError(error);
+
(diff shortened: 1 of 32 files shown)


packages/datagateway-common/src/handleICATError.ts
```diff
@@ -0,0 +1,40 @@
...
+          payload: {
+            severity: 'error',
+            message: error.message,
+          },
+        },
+      })
+    );
+  }
+  if (
+    error.response &&
```
esl-lewis: Currently the token is being invalidated on specific 403 messages which makes sense, I wonder if it should be broadened to invalidate on general error messages in case someone is trying to get in and manages to break the system in a new way.

yarn.lock
```diff
@@ -11722,7 +11732,7 @@ react-lifecycles-compat@^3.0.4:
   resolved "https://registry.yarnpkg.com/react-lifecycles-compat/-/react-lifecycles-compat-3.0.4.tgz#4f1a273afdfc8f3488a8c516bfda78f872352362"
   integrity sha512-fBASbA6LnOU9dOU2eW7aQ8xmYBSXUIWr+UmF9b1efZBazGNO+rcXT/icdKnYm2pTwcRylVUYwW7H1PHfLekVzA==
 
-react-redux@^7.1.0:
+react-redux@^7.1.0, react-redux@^7.1.3:
```
esl-lewis: looks like two versions of react-redux are locked here, possibly redundant

packages/datagateway-common/src/handleICATError.ts
louise-davies: We don't want to invalidate token on general errors because what if the user gets a server error - should that log them out? Limiting it to only invalidate the token on specific, invalid session id related errors means the user is prompted to re-login only when they have to.

yarn.lock
louise-davies: They're not being locked - this basically means two packages are requesting different ranges that can be resolved to the same version (i.e 7.1.3). I don't know why this updated though since I didn't touch anything - I think it's because common has a more recent dependency? I'll downgrade the version in common to match the others and remove this from the lockfile.

packages/datagateway-common/src/handleICATError.test.ts
```diff
@@ -0,0 +1,92 @@
...
+      },
+      name: 'Test error name',
+      message: 'Test error message',
+    };
+  });
+
+  it('logs an error and sends a notification to SciGateway', () => {
+    handleICATError(error);
+
+    expect(log.error).toHaveBeenCalledWith('Test error message');
```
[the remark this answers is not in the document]
louise-davies: This message has been defined for this test - essentially it's a mock error. In reality this will be passed a real axios error object which will contain more useful error information

packages/datagateway-common/src/handleICATError.ts
esl-lewis: understood!

packages/datagateway-common/src/handleICATError.test.ts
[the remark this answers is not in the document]
esl-lewis: As @louise-davies notes-these are test error messages, not to send to the user. I will close this conversation.

PR #116, opened 2020-02-19
Prevent incorrect URLS from fetching data

### LINKED ISSUE #59: [datagateway-table] Wrong URLs retrieve data
If a user edits the URL - then since the request generally only uses the previous entity ID to query then the user can grab data that is not associated at all with that hierarchy.

e.g. if the user browses, they go to `/browse/investigation/1/dataset/1/datafile`
but if they then change the url to `/browse/investigation/1/dataset/2/datafile` - it displays data from dataset of ID 2, when this dataset doesn't belong to investigation of ID 1.

TopCAT resolves this by checking each and every ID in the queries, which works for us up to the Investigation level. The problem will be ISIS's FacilityCycle and Instrument levels. (Proposals are trivially easy as they don't require a join, so can just filter by both name and ID)

Perhaps we send "preflight"-esque requests that manually check that the previous relations all make sense before allowing the user to see their data (e.g. for datafile, manually check that the investigation_id of the dataset it belongs to matches the investigation ID in the URL)? And forcibly display an "invalid url" message instead of the table? This will need to be specific to the configured tables since it is so closely linked to the hierarchy.

Acceptance criteria:
- [x] Users can't see incorrect data by manipulating the URL

louise-davies:
## Description
This PR creates a `withIdCheck` HOC that waits for the result of a promise, and if the result is true then renders the component it was passed. This is then used in `PageTable` for specific routes that supply specific "id check" functions that check the component IDs in the URL are valid to the ID being requested.

The tables affected are:
- `DatafileTable` - previously, the investigation ID could be modified and not actually contain the dataset of the datafiles we're displaying
- `ISISDatafileTable` - previously, the investigation ID, instrument id or facility id could be modified and not actually contain the dataset of the datafiles we're displaying
- `DLSDatafileTable` - previously, the investigation ID or proposal name could be modified and not actually contain the dataset of the datafiles we're displaying
- `ISISDatasetTable` - previously, the instrument id or facility id could be modified and not actually contain the investigation of the datasets we're displaying
- `DLSDatasetTable` - previously, the proposal name could be modified and not actually contain the investigation of the datasets we're displaying

(essentially, any url with more than 1 id in the URL was affected. ISISInvestigation table is unaffected due to the request already having to check both the instrument id + the facility cycle id)

Screenshot of page whilst loading:
ID check pending page

Screenshot of wrong URL page:
Wrong URL page

## Testing instructions
- [x] Review code
- [x] Check Travis build

## Agile board tracking
Fixes #59


agbeltran:
Note: check if possible to identify what part of the URL is wrong and include the correct URL at the lowest level of the hierarchy in the error message (so that the user knows where in the breadcrumbs to go back to).

PR #141, opened 2020-03-09
Feature/disable download type #120

### LINKED ISSUE #120: Check whether a download type is disabled in DownloadConfirmation
Admins can disable download types and so we need some way of stopping a user from submitting to a disabled download type and displaying a message to the user to describe why it is disabled.

TopCAT currently does this after the user clicks OK - but this is potentially confusing behaviour and perhaps a better way is telling the user before they try to submit their download. We would still need to be able to "select" a download in order to display the disabled reasoning message - so perhaps we should allow the option to be selected and display the message and disable the submit button.

See relevant TopCAT code: https://github.com/icatproject/topcat/blob/master/yo/app/scripts/controllers/download-cart.controller.js#L110

Acceptance criteria:
- [x] A user is not able to submit downloads to a download type that is disabled
- [x] A message is displayed to the user explaining that the download type is disabled or a custom message set on the download type status itself.
- [x] Ideally a better UX solution than TopCAT in that the user can tell that a download type is disabled before clicking the OK button.

GoelBiju:
## Description
Check to the status of an access method/download type within the download confirmation dialog before the user selects. In the event an access method is not available an appropriate message will be shown to the user and downloading via that method will not be allowed.

## Testing instructions

- [x] Review code
- [x] Check Travis build

## Agile board tracking
Closes #120


packages/datagateway-download/src/downloadConfirmation/downloadConfirmDialog.component.tsx
```diff
@@ -119,40 +136,74 @@ const DownloadConfirmDialog: React.FC<DownloadConfirmDialogProps> = (
...
+          [item.method]: {
+            ...settings.accessMethods[item.method],
+
+            // Set disabled to be true by default.
+            disabled: true,
+            message: '',
+          },
+        }),
+        {}
+      )
```
louise-davies: This code is a little unclear. We essentially want to take the `settings.accessMethods` object, and add the default `disabled` and `message` properties to each object right? I like functional programming and all that, but I think an easier to read and probably more efficient solution would just be a good ol' for...in loop

```
const defaultStatusMethods = {...settings.accessMethods};
for (const key in defaultStatusMethods) {
    defaultStatusMethods[key] = {...defaultStatusMethods[key], disabled: true, message: ''}
}
```

packages/datagateway-download/src/downloadConfirmation/downloadConfirmDialog.component.tsx
```diff
@@ -119,40 +136,74 @@ const DownloadConfirmDialog: React.FC<DownloadConfirmDialogProps> = (
...
+        message: string;
+      }
+    ][]
+  >([]);
+  const [isSorted, setIsSorted] = React.useState(false);
+  const [methodsUnavailable, setMethodsUnavailable] = React.useState(false);
 
+  // Size and two-level.
   const { totalSize } = props;
   const { isTwoLevel } = props;
```
louise-davies: you can destruct like this:

```
const { totalSize, isTwoLevel } = props;
```

packages/datagateway-download/src/downloadConfirmation/downloadConfirmDialog.component.tsx
```diff
@@ -161,15 +212,130 @@ const DownloadConfirmDialog: React.FC<DownloadConfirmDialogProps> = (
...
+    setSortedMethods(sorted);
+    setSelectedMethod(sorted[0][0]);
+    setIsSorted(true);
+  }, [statusMethods, setSortedMethods, setSelectedMethod, setIsSorted]);
+
+  React.useEffect(() => {
+    async function getStatus(): Promise<void> {
+      let statusErrors: string[] = [];
+      Promise.all(
+        Object.entries(statusMethods).map(([method]) =>
```
louise-davies: Why are you using `Object.entries` instead of `Object.keys` when you are just using the key in the `map`?

packages/datagateway-download/src/downloadConfirmation/downloadConfirmDialog.component.tsx
```diff
@@ -161,15 +212,130 @@ const DownloadConfirmDialog: React.FC<DownloadConfirmDialogProps> = (
...
+          getDownloadTypeStatus(method, {
+            facilityName: settings.facilityName,
+            downloadApiUrl: settings.downloadApiUrl,
+          })
+        )
+      ).then(methodStatus => {
+        // Loop through all the current access methods and match that
+        // to the status information we received for each.
+        Object.entries(statusMethods).forEach(([method], index) => {
```
louise-davies: Again, an `Object.entries` that just uses the key

packages/datagateway-download/src/downloadConfirmation/downloadConfirmDialog.component.tsx
```diff
@@ -161,15 +212,130 @@ const DownloadConfirmDialog: React.FC<DownloadConfirmDialogProps> = (
...
+    async function getStatus(): Promise<void> {
+      let statusErrors: string[] = [];
+      Promise.all(
+        Object.entries(statusMethods).map(([method]) =>
+          getDownloadTypeStatus(method, {
+            facilityName: settings.facilityName,
+            downloadApiUrl: settings.downloadApiUrl,
+          })
+        )
+      ).then(methodStatus => {
```
louise-davies: Since this is a `Promise.all`, this object is an array right? So comprises of multiple `methodStatus`. I know that status can be plural as well as singular but it would probably be clearer if it was statuses.

packages/datagateway-download/src/downloadConfirmation/downloadConfirmDialog.component.tsx
```diff
@@ -161,15 +212,130 @@ const DownloadConfirmDialog: React.FC<DownloadConfirmDialogProps> = (
...
+            // Push the method type to display an error.
+            statusErrors.push(method);
+          }
+        });
+
+        // Broadcast errors depending on the result of the status requests.
+        if (statusErrors.length < Object.keys(statusMethods).length) {
+          for (const method of statusErrors) {
+            broadcastError(
+              `Access method ${method.toUpperCase()} is currently unavailable. If required, use an alternative method.`
```
louise-davies: This error message could give a user the wrong idea. The access method might not be "currently unavailable" - what actually happened is that we could not ascertain the status of the access method. This error message should imply that the error was in fetching the status and not say whether the access method is actually enabled or disabled.

packages/datagateway-download/src/downloadConfirmation/downloadConfirmDialog.component.tsx
```diff
@@ -161,15 +212,130 @@ const DownloadConfirmDialog: React.FC<DownloadConfirmDialogProps> = (
...
+            );
+          }
+        } else {
+          setMethodsUnavailable(true);
+          broadcastError(
+            'Download access methods are unavailable. Please try again later.'
```
louise-davies: Again, it's more that the status of the access methods are unavailable - not the download methods themselves.

packages/datagateway-download/src/downloadConfirmation/downloadConfirmDialog.component.tsx
GoelBiju: @louise-davies I have simplified the code as you have shown in https://github.com/ral-facilities/datagateway/pull/141/commits/7facd345a991e683a4264bf7860b3f17406c16c3.

PR #162, opened 2020-03-18
Feature/rename datagateway table #127

### LINKED ISSUE #127: Rename package datagateway-table to datagateway-dataview
As we are adding other views, in addition to table view, we are renaming the package ```datagateway-table``` to ```datagateway-dataview```

GoelBiju:
## Description
Rename `datagateway-table` to `datagateway-dataview` as we will be providing multiple views of the data and not just a table view.

## Testing instructions

- [x] Review code
- [x] Check Travis build

## Agile board tracking
Closes #127


GoelBiju:
> Do we want to keep the name of the top level state as `dgtable`? These usually match the plugin name, but `dgdataview` is pretty long...

@louise-davies I think having it as `dataview` would be the most appropriate since it still isn't extremely long.

PR #179, opened 2020-04-14
Feature/overflow tooltip #67

### LINKED ISSUE #67: [datagateway-table] Evaluate handling overflowed table items for accessibility/readability
Currently, when we hover over a table item that is too long it displays the overflow in a long line. Keiran raised this is difficult to read, so we should evaluate whether this is needed or if there are better ways to display this

- Tooltip instead of overflow?
- Line breaks in overflow?
- Do resizable columns resolve this issue entirely?
- Are people happy with opening the details panel?

Acceptance criteria:
- [ ] Decision on how to display overflow rows

tswinb:
## Description
Addresses #67 by adding the hover tooltip component to table cells, and removes existing overflow code.

## Testing instructions

- [x] Review code
- [x] Check Travis build

## Agile board tracking
Closes #67


tswinb:
At the moment, I've written it so the tooltip appears on every cell - but could try using the percentageWidth parameter in ArrowTooltip to choose which a width when tooltip will first appear.

GoelBiju:
@tswinb you might want to remove the comments I have made in the component. My inital plan was to not have specific parameters all in one component but have indivual components which wrap around the original ArrowTooltip have provide it. Since you are using it then it's fine.

packages/datagateway-common/src/arrowtooltip.component.tsx
```diff
@@ -0,0 +1,146 @@
+import React, { useEffect } from 'react';
+import Tooltip, { TooltipProps } from '@material-ui/core/Tooltip';
+import { makeStyles, createStyles, Theme } from '@material-ui/core/styles';
+
+const arrowGenerator = (
```
louise-davies, line 5: If this code has been migrated/copied over to `common` - can we remove the version in `dataview` and instead import it from `common` there? Otherwise we have code repetition 😬

packages/datagateway-common/src/table/table.component.tsx
```diff
@@ -60,12 +60,10 @@ const styles = (theme: Theme): StyleRules =>
...
-      '&:hover': {
-        overflow: 'visible',
-        zIndex: 10000,
-        position: 'absolute',
-        backgroundColor: theme.palette.grey[200],
-      },
+      display: 'block',
+      whiteSpace: 'nowrap',
+      overflow: 'hidden',
+      textOverflow: 'ellipsis',
```
louise-davies, line 66: I'm pretty sure all of these styles are already being applied by the `Typography` component via the `noWrap` property

GoelBiju:
> It works well on overflowed items, however, even if an item isn't overflowing, the tooltip will still show. Do we want this behaviour? Might be a little annoying for users...
> 
> I know @GoelBiju did some work on this with the breadcrumbs - perhaps the code there can be moved to common and shared between the components as a generic "if an item is overflowing, show a tooltip, otherwise don't" component?

You are right @louise-davies, this is an issue which I wasn't sure about. The initial thought was to detect if the element which the Tooltip wraps is overflowing or not but it would be have to be dynamic. I have mentioned this to @tswinb and I think maybe we have to pass in a prop (maybe change how the percentageWidth works) to dynamically update the maximum width at the which the tooltip is disabled for. 

I guess in this case, it would be the current width of the cell. Maybe pass in the width of the cell to the tooltip component and as long as the content inside stays within that width, don't show the tooltip, if it exceeds it then you will have to. Although there might need to be some CSS done to allow for full width of the content to be exposed with `innerWidth`?

tswinb:
@GoelBiju @louise-davies I've managed to resolve it with 
`if (
          tooltipElement.current.offsetWidth <
          tooltipElement.current.scrollWidth
        ) {
          setTooltipVisible(true);
        } else {
          setTooltipVisible(false);
        }` (only show tooltip if the visible width is smaller than the full scroll width).

GoelBiju:
@tswinb I just noticed a bug with the ArrowToolip. If you decrease the page zoom by scrolling out (in Chrome with ctrl + scroll wheel), the component eventually produces an error saying that it `Cannot read property 'offsetHeight' of null`. I noticed this as it was present on the CardView component, and when using the `maxEnabledHeight`.

The solution for this seems to be to just check that `tooltipElement.current` is not null by adding a `tooltipElement.current &&` to the [percentageWidth check](https://github.com/ral-facilities/datagateway/blob/bf66c69bd9c43b6a307a45257d6c47b25977d166/packages/datagateway-common/src/arrowtooltip.component.tsx#L98) and to the [maxEnabledHeight check](https://github.com/ral-facilities/datagateway/blob/bf66c69bd9c43b6a307a45257d6c47b25977d166/packages/datagateway-common/src/arrowtooltip.component.tsx#L107). A check may also have to be added for the `offsetWidth` and `scrollWidth` [here](https://github.com/ral-facilities/datagateway/blob/bf66c69bd9c43b6a307a45257d6c47b25977d166/packages/datagateway-common/src/arrowtooltip.component.tsx#L115). 

It might be worth removing the `&& tooltipElement.current !== null` [initially](https://github.com/ral-facilities/datagateway/blob/bf66c69bd9c43b6a307a45257d6c47b25977d166/packages/datagateway-common/src/arrowtooltip.component.tsx#L92) as well if we are checking it later on.

louise-davies:
> would a delayed entry be appropriate on the tooltip? It would be shown slightly later (maybe 500ms) so that it is not instantly shown to a user who may not be interested in reading it in full.

Yeah, I think a slight delay would be good so that it only shows if a user is actively hovering and not just passing over.

PR #227, opened 2020-06-02
Feature/search settings #125

### LINKED ISSUE #125: Add settings file to datagateway-search
Currently the URLs in datagateway-search are hardcoded - this should be changed so they can be changed via a settings file.

Acceptance criteria:
- [x] datagateway-search reads it's configuration information from a settings file
- [x] URLs for APIs are retrieved from settings.

tswinb:
## Description
datagateway-search reads its configuration information from a settings file. Load URLs, facilityName and ui-strings folder into state from settings file. 

## Testing instructions

- [x] Review code
- [x] Check Travis build

## Agile board tracking
resolves #125

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -27,3 +27,4 @@ yarn-error.log*
 
 **/public/settings.json
 **/public/datagateway-table-settings.json
+**/public/datagateway-search-settings.json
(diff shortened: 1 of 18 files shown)


GoelBiju:
A small warning/error spotted with @tswinb, when using `datagateway-search`, is the errors produced when loading the investigation, dataset and datafiles tabs. Material-UI mentions that the values provided do not exist e.g. none/investigation/dataset/datafile. 

This does not affect how the search works and is just error because of how we do not want to render any tabs (and not hide them) initially. Since we optionally render them based on if they are selected, it could be that between renders they do not exist and this could be why material-ui does not find the `Tab` component with the value it should expect.

Issue #84, opened 2019-11-18
[all?] Internationalisation


louise-davies:
The breadcrumbs code review brought up that entities were being pluralised using the letter 's' - this is obviously English specific and should be handled by an internationalisation library/component. Additionally, `datagateway-table` currently hard codes its strings - this can also be fixed by an internationalisation library/component

Acceptance criteria:
- [ ] Internationalisation library is used to pluralise entity names
- [ ] Internationalisation library is used to provide alternate language settings for strings

GoelBiju:
These are two libraries to possibly be used for this particular issue: [i18next](https://www.i18next.com/) and [polyglot.js](https://airbnb.io/polyglot.js/).

louise-davies:
Additionally there's [`react-intl`](https://github.com/formatjs/react-intl)

PR #286, opened 2020-06-25
Internationalisation

louise-davies:
## Description
This PR adds `react-i18next` to `datagateway-dataview` and `datagateway-download`, which allows for text in the app to be sourced from a JSON file and thus configured at will. It also helps with some things like pluralisation in `datagateway-download`.

Note that most of the changes aren't really visible, since if everything is working correctly then the strings should just load and not look any different to before. The only exception is that I didn't make DLS/ISIS specific sections to the translation file, so they share the same default words (e.g. DLS used to say "beamline" instead of "instrument")

## Testing instructions
- [x] Review code
- [x] Check Travis build

## Agile board tracking
part of #84 (still need to do it for `datagateway-search`)


packages/datagateway-dataview/src/pageContainer.component.test.tsx
```diff
@@ -14,6 +14,21 @@ import { createLocation } from 'history';
 import { MemoryRouter } from 'react-router';
 
 import PageContainer from './pageContainer.component';
+// import { I18nextProvider } from 'react-i18next';
```
tswinb, line 17: Did you mean to keep these comments in?

packages/datagateway-dataview/src/breadcrumbs.component.tsx
```diff
@@ -248,12 +251,9 @@ class PageBreadcrumbs extends React.Component<
         base: {
           entityName: baseEntityName,
 
-          // TODO: This display name requires internationalisation,
-          //       as currently it adds an 's' to the word disregarding
-          //       the current language the application is being served in.
-          displayName:
-            `${baseEntityName}`.charAt(0).toUpperCase() +
-            `${baseEntityName}s`.slice(1),
+          displayName: this.props.t(`breadcrumbs.${baseEntityName}`, {
```
tswinb, line 254: There's something going wrong with the breadcrumb label for facility cycles
image

packages/datagateway-download/src/downloadTab/downloadTab.component.tsx
```diff
@@ -94,23 +96,39 @@ const DownloadTabs: React.FC = () => {
         textColor="primary"
         centered
       >
-        <Tab label="Cart" aria-label="Cart tab" {...a11yProps(0)} />
-        <Tab label="Downloads" aria-label="Downloads tab" {...a11yProps(1)} />
+        <Tab
+          label="Cart"
```
tswinb, line 100: Should the tab labels be also 'translated' as well as the aria-labels?

louise-davies: Totally missed this, yes they should be!

packages/datagateway-dataview/src/pageContainer.component.test.tsx
louise-davies: oops

PR #435, opened 2020-09-24
Feature/display search numbers #419

### LINKED ISSUE #419: UI improvement: show count of search results
**Description**:
Currently we are not showing count of search results. Add count for each entity type.

**Acceptance criteria**:
- [ ] Show count of search results

patrick-austin:
## Description
Added badges to each tab of the search view to show the number of search results of that type.

## Testing instructions

- [x] Review code
- [x] Check Travis build
- [x] Check the badge number is accurate for all tabs after searching
EDIT:
- [ ] Check alignment for different browser font sizes

## Agile board tracking
closes #419


louise-davies (commented):
Works well, just some style comments. The badge numbers are shown as a superscript - I wonder if showing them inline makes more sense as currently it looks like they're being squared etc.! Additionally, I think the badges themselves should be a different colour to the background to make them stand out a bit more as badges not just numbers (not the default orange! maybe white background with blue text?)

louise-davies:
Looking better, I prefer the new colour, but it still seems as though they're slightly misaligned. 

image

Like the badge needs to be a little higher?

louise-davies (commented):
Looking better, I prefer the new colour, but it still seems as though they're slightly misaligned. 

image

Like the badge needs to be a little higher?

patrick-austin:
Replaced the hardcoded alignment in `px` with `em` and `ch` (font height / character width). Not sure what the root cause of the differing alignmnet was (the aligment looked good locally, but my node modules were out of wack at the time, or could just be different browser settings?).
Badge should be in line with the label, and label+badge should appear centrally within each tab irrespective of the current font size. One thing worth noting is the line height was (for me at least) set to 1.75, but the line height unit (`lh`) is an "experimental" feature so I couldn't use it to dynamically set the vertical alignment. So there's still a factor of 1.75 hardcoded into the badge's `top` positioning.

louise-davies (approved):

Issue #426, opened 2020-09-22
Add loading indicator to tables


louise-davies:
**Description**:
Currently, there is no loading indicator on the tables, so users don't know if there is data attempting to be loaded or if there is just no data.

**Acceptance criteria**:
- [ ] A loading indicator is added to the table

GoelBiju:
@louise-davies just as a comment to this, there is a [loading indicator](https://github.com/ral-facilities/datagateway/blob/feature/create-card-view-%23128/packages/datagateway-dataview/src/page/pageContainer.component.tsx#L279-L291) on the `PageContainer` (on the card-view branch) which uses a `LinearProgress` component to be shown when data is being loaded when cards are toggled. Not sure if a similar approach should be used for tables (if that is what you had in mind).

PR #165, opened 2020-03-20
Feature/create card view #128

GoelBiju:
## Description
Create a generic **CardView** component to show data within Cards in a Grid for varying entities (investigations/datasets).

Additionally, implements an *InvestigationCardView*/*DatasetCardView* to illustrate usage of the CardView component.

The pull request adds in the following ISIS card views:

- instruments,
- facility cycles,
- investigations,
- datasets.

The pull request adds in the following DLS (DIAMOND) card views:

- proposals,
- visits,
- datasets.

## Testing instructions

- [ ] Review code
- [ ] Check Travis build
- [ ] Review changes to test coverage

## Agile board tracking
Closes #128, closes #395

PR #438, opened 2020-09-29
Feature/table loading indicator #426

patrick-austin:
## Description
Added a `LinearProgress` loading indicator to the datagateway plugins in line with the one being added in #165 (but without the `Route` element or depenency on card view).

## Testing instructions

- [x] Review code
- [ ] Check Travis build
- [x] Check the loading bar only appears when table data is yet to load in the datagateway plugins

## Agile board tracking
connect to #426


packages/datagateway-download/package.json
```diff
@@ -20,8 +21,12 @@
     "react": "^16.13.1",
     "react-dom": "^16.11.0",
     "react-i18next": "^11.7.0",
+    "react-redux": "^7.1.0",
```
louise-davies, line 24: Why does `datagateway-download` now require redux? This adds a massive overhead. The loading state in `datagateway-common` will not indicate anything useful here I think. If a loading indicator is needed in download, then an alternative solution needs to be thought of (e.g. do it on the individual table components, so that the loading boolean can be passed in that way)

louise-davies (changes requested):
I wonder whether it's easier to perform this logic in the central `table` component, since it's already passed a `loading` prop to know when to enable select boxes, so perhaps that is a better way to go about it, especially since `datagateway-download` doesn't have redux?

packages/datagateway-download/package.json
patrick-austin: Have moved the loading bars from `downloadTab` to `downloadCartTable` and `downloadStatusTable`, where `dataloaded` is used to perform the logic. This removes the need for redux to access `loading` (as we do in search and dataview).

louise-davies (approved):

PR #434, opened 2020-09-23
Feature/add cart to dataview #415

patrick-austin:
## Description
Added an icon for the download cart that includes the number of items currently in it on the dataview plugin. Clicking the cart will link to /download.

## Testing instructions
- [x] Review code
- [ ] Check Travis build
- [x] Review changes to test coverage
- [x] Check icon appears and that clicking links to /download
- [x] Check the badge with the number of items in the cart displays and is accurate

## Agile board tracking
connect to #415

Issue #430, opened 2020-09-23
Extend tour


agbeltran:
**Description**:
The first pass of the tour was done some time ago and does not include the latest features. The tour needs to be extended to provide guidance on the features not yet covered (e.g. #429). 

**Acceptance criteria**:
- [ ] Tour extended to include all the new features

patrick-austin:
The tour was extended, and the functionality to add more steps is in place. The text may need reviewing or new steps adding as features are added, but unless there's anything that currently needs adding I'm closing this issue.

PR #442, opened 2020-10-02
add plugin specific tour steps #430

patrick-austin:
## Description
Have replaced `helpText` with `helpSteps` which allows the tour to target specific elements within a plugin. This requires /ral-facilities/scigateway#443 code to display. Added classNames to elements to be targetted by the tour. If the element cannot be found it is just ommited from the tour, so only elements curretly displayed will be part of the tour. This also means elements introduced in other PRs (e.g. cart/search icons in #434) have the help text specified here but this won't display until that element is added and given the correct className.

Some of the help text might be unnecessary (i.e. it explains something obvious) but I wanted to make sure there was at least one element for each plugin as a proof of concept.

## Testing instructions

- [ ] Review code
- [ ] Check Travis build
- [ ] Check help for the navbar displays for all plugins
- [ ] Check that in dataview the breadcrumbs, results (and icons if present) have help text
- [ ] Check that in search the tabs have help text
- [ ] Check that in download both tabs have help text

## Agile board tracking
connect to #430


louise-davies:
note to self: at lesley to review text once screenshots are created for tour steps

packages/datagateway-dataview/src/App.tsx
```diff
@@ -77,7 +77,36 @@ const registerRouteAction = {
...
+        content: 'Navigate to Datagateway Search',
+      },
+      {
+        target: '.tour-dataview-cart-icon',
+        content:
+          'Displays the current number of items in the download cart, and navigates to Datagateway Download',
+      },
+      {
+        target: '.tour-dataview-card-view',
+        content: 'Toggle between list and card view',
```
VKTB, line 107: This help step does not get displayed, I think it is because the element does not have the class specified in `target`.

VKTB (changes requested):
Just a couple of minor things that I noticed but looks good otherwise.

Not sure how much of an issue this is but I noticed that the final help steps of the _DataView_ and _Search_ tours display _Next_ instead of a _Last_ button. For the _DataView_ I think that this is because the `helpStep` is present but it cannot find the class element provided in `target`, meaning that the final help step that it displays is not technically the final. Not sure why this is happening in _Search_ though.

packages/datagateway-dataview/src/App.tsx
patrick-austin: Good catch, when I started this the card view hadn't been merged into master so I couldn't add the classname to the toggle as it wasn't there. But it's been merged now so I'll add that in.

patrick-austin:
> Not sure how much of an issue this is but I also noticed that the final help steps of the _DataView_ and _Search_ tours display _Next_ instead of a _Last_ button. For the _DataView_ I think that this is because the `helpStep` is present but it cannot find the class element provided in `target`, meaning that the final help step that it displays is not technically the final. Not sure why this is happening in _Search_ though.

Had a look into this, and you're right that it's because the `helpStep` is registered for the tour but it can't find the target, but it's happening in both plugins.

Because only the elements for the current plugin are in the DOM, all steps belonging to a different plugin will silently fail when the tour runs. This isn't a problem in itself, as only displaying help for elements on screen and not forcing a user onto a different plugin makes sense (to me at least). If one of those failed steps happened to be the last step to be registered, then "Last" will never appear because the step never runs. This could also happen on the search plugin because the tabbed pane only appears after searching and so the target element is present by default.

This can be solved in SG by removing steps from the tour based on whether the target can be found prior to starting the tour. We already do a similar check to remove help for the plugins in the sidebar when a user isn't logged in (as they won't appear for them). I've pushed this fix to the corresponding SG PR /ral-facilities/scigateway#443.

VKTB (approved):

louise-davies (approved):

Issue #459, opened 2020-11-04
Tests for cardView changes


patrick-austin:
**Description**:
As it stands, any tests that were failing as a result of cardView have been fixed (or skipped), but coverage in unit tests is low, and cypress tests in table rather than card view. Coverage should be increased, and cypress tests written. See #165.
- [x] DG-common unit tests
- [x] DG-dataview unit tests: changes to existing tests
- [x] DG-dataview unit tests: new tests for cardView features
- [x] Cypress tests

**Acceptance criteria**:
- [ ] Acceptable level of codecov through unit tests
- [ ] e2e tests for cardView cover the same features as existing table tests, in addition to cardView only features

patrick-austin:
Also, try and reduce warnings/loggings from the e2e tests. This is making debugging actual failures less clear.

patrick-austin:
Two areas that I haven't been able to hit with the unit tests are `Select` and `Route` elements that use the `render` function. It seems like since the updgrade to MUI that came with card view, the menu `Select` creates doesn't get picked up by the wrapper, so can't interact with it. I also wasn't able to get the code inside the `render` function to run, so files that use it a lot (`pageCard`/`pageContainer`) have low coverage.

PR #474, opened 2020-11-12
Feature/improve cv tests #459

patrick-austin:
## Description
Unit tests for card view, along with some minor changes to prevent errors and logging in tests.

e2e tests are not included (yet) as there are other issues that are a higher priority to work on now.

## Testing instructions
- [x] Review code
- [x] Check Travis build
- [x] No functional changes expected

## Agile board tracking
connect to #459


patrick-austin:
Disabled the e2e test for card/investigations to get Travis to pass. While it was passing locally, it kept failing for Travis. The errors it gets are the same infrequent errors that affect the other e2e tests (if you don't wait for requests, then components will be incorrectly sorted, but if you do, sometimes it will timeout waiting for them) but for whatever reason are far more frequent for this particular test. Maybe due to the high number of investigations compared to datasets/files?

VKTB (approved):
Looks good! I tried my best with the unit tests.

VKTB:
Just noticed a note that I made in my notebook about using constants for the `aria-label` values rather than hard coded strings. For example the `Start Date date filter to` has been used as a `aria-label` value 20 times in 10 different files in the code base so if it changed for some reason in future then changes in lots of places will need to be made.

patrick-austin:
> 
> 
> @patrick-austin Just found my note that I made in my notebook about using constants for the `aria-label` values rather than hard coded strings. For example the `Start Date date filter to` has been used as a `aria-label` value 20 times in 10 different files in the code base so if it changed for some reason in future then changes in lots of places will need to be made. Same with any other values.

That's a good point, I've had a quick look and in some (vast minority of) places we do set the `aria-label` using a translated string from the res file so it should be possible to do it like that? @louise-davies has mentioned accesibility considerations before, I don't know if there's an existing plan for this, or if we should make an issue dedicated to making sure everything has an appropriate (non-hardcoded, translated) aria label?

At any rate it looks like we over 1000 hardcoded labels, so it's probably outside the scope of this PR so I'll merge it as is.

PR #478, opened 2020-11-18
Fix date search in datagateway-search for only one date

### LINKED ISSUE #464: Fix date search for only 1 date
**Description**:
Currently date searching with only a start or end date via datagateway-search doesn't work - this is because we only specify the one limit and lucene requires both be set (and the one the user didn't specify needs to be set at a ridiculous lower or upper bound)

**Acceptance criteria**:
- [ ] Date search works with only 1 of start or end date

VKTB:
## Description
Sets a date limit to the date that is not provided so that only the results before/ after the provided date are displayed.

## Testing instructions
- [x] Review code
- [x] Check Travis build
- [x] Navigate to _Datagateway Search_, tick all three boxes (_Investigation_, _Dataset_ and _Datafile_), pick a start date, click the _SEARCH_ button, click on the _INVESTIGATION_ tab that is present in the results table and check that only results after the picked date are displayed. Check that this is the case in the _DATASET_ and the _DATAFILE_ tab.
- [x] Repeat the above step but picking only an end date this time.

## Agile board tracking
Closes #464

### DIFF
--- a/packages/datagateway-search/src/search/searchButton.component.test.tsx
+++ b/packages/datagateway-search/src/search/searchButton.component.test.tsx
@@ -235,7 +235,7 @@ describe('Search Button component tests', () => {
           query: {
             target: 'Datafile',
             lower: '201311110000',
-            upper: '210001012359',
+            upper: '9000012312359',
           },
           sessionId: null,
         },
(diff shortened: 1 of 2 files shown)


louise-davies (changes requested):
Looks good - my only comment is that the start and end date "limits" could be more extreme; see TopCAT: https://github.com/icatproject/topcat/blob/master/yo/app/scripts/controllers/search-results.controller.js#L17

It's not unreasonable for 1984 to be a possible date, and if our software is still in use in 80 years I'd be surprised but it's not totally impossible ;)

louise-davies (approved):
Looks good and works well :)

Issue #460, opened 2020-11-04
Sorting issues for certain fields


patrick-austin:
**Description**:
Have seen issues with the following fields when sorting, even when that field is populated:
- facilityCycle (ISIS):
  - Name: no effect on order in card/table
  - Description: no effect on order in card/table
- visit/investigation (DLS):
  - Beamline/Instrument: removes entries in card/table 
- investigation (ISIS):
  - instrument: removes entries in card/table

Has issues, but the field is unpopulated (so maybe we just don't handle empty fields properly when sorting?):
- investigation (ISIS):
  - DOI: removes entries in card/table
- investigation (Generic):
  - RB Number: removes entries in card/table
  - Instrument: removes entries in table view only
- dataset (Generic)
  - Datafile Count: removes entries, cardView only

Need to look into this more, work out what needs to be done.

**Acceptance criteria**:
- [ ] ?

louise-davies:
Sorting by instrument is an issue with the API that is being tracked via ral-facilities/datagateway-api#122

Sorting by datafileCount should be disabled as this is a calculated value (like size).

Facility cycles should work - we need to check this in the API, as well as the other ones.

patrick-austin:
Covered by ral-facilities/datagateway-api#122:
- [ ] Visit/investigation beamline/instrument
- [ ] ISIS investigation DOI: we're currently trying to get the DOI by `STUDYINVESTIGATION.STUDY.PID`, which gives no results and also doesn't sort. Could get it straight from the investigation (DOI is a field in ICAT schema for investigations) but if storing it on the study is more realistic I'll leave it as is

Will address:
- [ ] RB number in isn't a field in the ICAT schema, but is in our `app.types` interface, so attempting to sort by it always fails. I was under the impression RB was a different term for VISIT_ID (which we already display in the investigation table). Will disable sorting for now, but could also just remove RB from `app.types` (assuming that because it doesn't appear in the schema we'll never get a returned value for it)?
- [ ] Datafile count should be disabled in card view like it is in table view and DLS datasets

Can't diagnose:
- [ ] FacilityCycle name/description sends the requests to the API successfully (unlike the other issues where the request 500s), and can't see any reason why it would work for dates but not the name/description. We do join a lot of entities for the request, so maybe it's related to ral-facilities/datagateway-api#122 and won't be a problem when running against an API that's had that fix?

sam-glendenning:
Currently, a couple of issues are still known about. 

[...]

PR #499, opened 2020-12-01
add StudyHierarchy for ISIS data, closes #480

### LINKED ISSUE #480: Create new ISIS hierarchy
**Description**:
In the ISIS demo, it was suggested that perhaps a more intuitive/user friendly hierarchy would be `Instrument -> Investigation (via RB Number) -> Dataset -> Datafile`. Additionally, they thought that perhaps it should be togglable like card view is.

This hierarchy should be easier to implement, and should just require configuring the ISIS investigations table to display the correct data and for RB numbers to be more prominent, and for a toggle to switch between which table is linked to from the Instruments table. Trying to keep both functionalities in the Investigations table might be a bit tricky though, or a bit of a mess with the toggle boolean.

**Acceptance criteria**:
- [ ] New ISIS hierarchy implemented
- [ ] Users can switch between old and new ISIS hierarchy

patrick-austin:
## Description
Introduce alternate hierarchy for ISIS data:
- Instrument > Study > Investigation > Dataset > Datafile

This can be accessed by navigating to `/browseStudyHierarchy/...` rather than `/browse/...` which will include facility cycles as before. Navigation within each hierarchy will only link to other pages with the same browse prefix, so if the plugin is registered with the link `/browse(StudyHierarchy)/instrument` then it will only allow the user to navigate to the corresponding pages (though in reality providing a valid web address will work irrespective of which plugin was linked.

This is done with routing, and the addition of a prop to the existing tables/CVs which determines which address to link to next.

Also changed "Visit ID" to "RB Number" for ISIS views and added translations to CV labels.

Note that sorting for the new study table/CV doesn't work for the same reason as mentioned on #460 and ral-facilities/datagateway-api#122. The latter says this should be fixed, but maybe the api url the dev instance points to (http://scigateway-preprod.esc.rl.ac.uk:5000) doesn't have the fix yet?

## Testing instructions

- [ ] Review code
- [ ] Check Travis build
- [ ] Check navigation between entities works for both hierarchies

## Agile board tracking
connect to #480


packages/datagateway-dataview/public/res/default.json
```diff
@@ -35,6 +36,13 @@
     "end_date": "End Date",
     "start_date": "Start Date"
   },
+  "studies": {
+    "name": "Name",
+    "rb_number":"RB Number",
```
louise-davies, line 41: I'd prefer this to match the ICAT field name - rb_number isn't a field - it's generally the `name` field

packages/datagateway-dataview/src/views/card/isis/isisInvestigationsCardView.component.tsx
```diff
@@ -195,52 +218,52 @@ const ISISInvestigationsCardView = (
...
-        label: 'Description',
+        label: t('investigations.details.summary'),
         dataKey: 'SUMMARY',
         filterComponent: textFilter,
       }}
       information={[
         {
           icon: <Fingerprint />,
-          label: 'Visit ID',
+          label: t('investigations.rb_number'),
```
louise-davies, line 228: The correct field name for Visit ID is `visit_id` (see old labels) - "RB number" is actually the name field, but again this shouldn't be a translation key name, instead it should be the value of the name key.

packages/datagateway-dataview/src/views/card/isis/isisStudiesCardView.component.tsx
```diff
@@ -0,0 +1,251 @@
...
+      }}
+      description={{
+        label: t('facilitycycles.description'),
+        dataKey: 'STUDY.DESCRIPTION',
+        filterComponent: textFilter,
+      }}
+      information={[
+        {
+          icon: <Fingerprint />,
+          label: t('studies.rb_number'),
```
louise-davies, line 161: Again, would prefer the translation key to match the dataKey

packages/datagateway-dataview/src/views/card/isis/isisStudiesCardView.component.tsx
```diff
@@ -0,0 +1,251 @@
...
+        content: (studyInvestigation: StudyInvestigation) =>
+          tableLink(
+            `/${pathRoot}/instrument/${instrumentId}/${instrumentChild}/${studyInvestigation.STUDY.ID}/investigation`,
+            studyInvestigation.STUDY.NAME,
+            view
+          ),
+        filterComponent: textFilter,
+      }}
+      description={{
+        label: t('facilitycycles.description'),
```
louise-davies, line 154: Some of the labels here all refer to facilitycycles, not studies

packages/datagateway-dataview/src/views/card/isis/isisStudiesCardView.component.tsx
```diff
@@ -0,0 +1,251 @@
...
+    ),
+  fetchData: (allIds: number[], offsetParams: IndexRange) =>
+    dispatch(
+      fetchStudies({
+        offsetParams,
+        additionalFilters: [
+          {
+            filterType: 'where',
+            filterValue: JSON.stringify({
+              INVESTIGATION_ID: { in: allIds },
```
louise-davies, line 204: This is a clever way to add instrument context to a study search! Again, unsure if this is going to be the final way we fetch study info but it's fine for now

packages/datagateway-dataview/src/views/table/isis/isisInvestigationsTable.component.tsx
```diff
@@ -185,7 +211,7 @@ const ISISInvestigationsTable = (
         },
         {
           icon: <FingerprintIcon />,
-          label: t('investigations.visit_id'),
+          label: t('investigations.rb_number'),
```
louise-davies, line 214: Revert this change

packages/datagateway-dataview/src/views/table/isis/isisMyDataTable.component.tsx
```diff
@@ -205,7 +205,7 @@ const ISISMyDataTable = (
         },
         {
           icon: <FingerprintIcon />,
-          label: t('investigations.visit_id'),
+          label: t('investigations.rb_number'),
```
louise-davies, line 208: Revert

packages/datagateway-dataview/src/views/table/isis/isisStudiesTable.component.tsx
```diff
@@ -0,0 +1,217 @@
...
+          cellContentRenderer: (props: TableCellProps) =>
+            tableLink(
+              `/${pathRoot}/instrument/${instrumentId}/${instrumentChild}/${props.rowData.ID}/investigation`,
+              props.rowData.STUDY?.NAME
+            ),
+          filterComponent: textFilter,
+        },
+        {
+          icon: <FingerprintIcon />,
+          label: t('studies.rb_number'),
```
louise-davies, line 129: Again, the study RB number is the name. Studies don't have visit ids, since one study could contain multiple visits

louise-davies (changes requested):
Looks good overall, mostly just some confusion on what RB numbers map to in ICAT. However, I have noticed that if you click on a Study, what shows up in the breadcrumb isn't the same name as the study I just clicked on. e.g. clicking on `STUDY 375` in the table/card view takes me to `STUDY 7` in the breadcrumbs. Could you investigate what's going on here?

Again, it should be noted that this view requires study objects to exist, so would require creation of studies for every investigation currently in ISIS ICAT. We will need to discuss whether this is the most sensible option later.

Additionally, we need to think about which URL we're hosting this on. (probably worth opening issues to address these two points)

patrick-austin:
The breadcrumb issue was due to using `STUDYINVESTIGATION.ID` for the link in the studies table rather than `STUDYINVESTIGATION.STUDY.ID` which is what we were (correctly) doing in card view. So it wasn't an issue with the breadcrumbs, but the actual navigation. 
I'll revert my mistaken attempts to include the RB Number, and use it as the string for `NAME`/`investigations.isis.name` instead (on ISIS tables/CVs), and include the `PID`/DOI as a field to display.

packages/datagateway-dataview/public/res/default.json
```diff
@@ -72,6 +82,9 @@
         "label": "Publications",
         "reference": "Reference"
       }
+    },
+    "isis": {
+      "name":"RB Number"
```
louise-davies, line 87: I'm not exactly a fan of this "isis" section header - the idea would be that the facility would configure `investigation.name` to be either `RB Number` in ISIS, or `Name` for DLS.

packages/datagateway-dataview/src/views/table/isis/isisStudiesTable.component.tsx
```diff
@@ -114,28 +114,28 @@ const ISISStudiesTable = (
...
       columns={[
         {
-          icon: <TitleIcon />,
+          icon: <FingerprintIcon />,
           label: t('studies.name'),
           dataKey: 'STUDY.NAME',
           cellContentRenderer: (props: TableCellProps) =>
             tableLink(
-              `/${pathRoot}/instrument/${instrumentId}/${instrumentChild}/${props.rowData.ID}/investigation`,
+              `/${pathRoot}/instrument/${instrumentId}/${instrumentChild}/${props.rowData.STUDY.ID}/investigation`,
```
louise-davies, line 122: When I load the table and click on the first instrument, when loading the study table I now get an error: `props.rowData.STUDY is undefined` which I presume stems from here. However,  if I load the page directly (e.g. reloading the page once I get the error) it works fine. This error only occurs in Table view, but if I navigate via Card view then switch to Table view then the data loads correctly with no error. It seems as if the data isn't being requested when the route changes?

I think it might be because of the two useEffects required for fetching data in this table.

louise-davies (changes requested):

packages/datagateway-dataview/public/res/default.json
patrick-austin: Ah OK, the original issue mentioned making RB numbers more apparent or something like that so I thought I needed to change something to say "RB Number" as part of the change, but it being configurable by the user makes sense. Provided we're displaying the `<entity>.name` for relevant tables/views it should be OK then?

packages/datagateway-dataview/src/views/table/isis/isisStudiesTable.component.tsx
patrick-austin: When navigating directly to studies table `data` is empty, and when swapping from cardview to table we already have `data` populated with `StudyInvestigation`s. In both cases `props.rowData.STUDY` can be called safely.

When navigating from instruments `data` is populated with `Instrument`s which obviously don't have `STUDY` on them, and so it errors (`clearTable` is called straight away but the error still occurs before it can reset `data`).

We don't see the issue with card view as due to the pagination logic, the link isn't rendered straight away, only after we've called `fetchData` once. All other tables use `<entity>.ID` so this isn't an issue normally, as the new table can still access an ID from the old `data` for however long the split second is before `clearTable` hits.

Rather than fundamentally change anything to do with routing or clearing, just changing it to `STUDY?.ID` will ensure we don't error while waiting between the location change and `clearTable`.

louise-davies (dismissed):
The conditional check fixed the problem I was having so this looks good to go 👍

patrick-austin:
Have updated tests to use translated strings, and the routing to use the new ID checks so re-review needed.

louise-davies (approved):
still looks good 👍

PR #510, opened 2020-12-10
disable DF count and RB sorting #460

patrick-austin:
## Description
Disables sorting on fields which we don't have on the entity, which resulted in rejected requests to the API and no data in the table. Doesn't address the other aspects of #460 

## Testing instructions
- [x] Review code
- [ ] Check Travis build
- [x] Check we can't sort by Datafile Count in generic dataset card view
- [x] Check we can't sort by RB Number in generic investigation card/table view

## Agile board tracking
connect to #460

### DIFF
--- a/packages/datagateway-dataview/src/views/card/datasetCardView.component.tsx
+++ b/packages/datagateway-dataview/src/views/card/datasetCardView.component.tsx
@@ -168,6 +168,7 @@ const DatasetCardView = (props: DatasetCVCombinedProps): React.ReactElement => {
           icon: <ConfirmationNumber />,
           label: 'Datafile Count',
           dataKey: 'DATAFILE_COUNT',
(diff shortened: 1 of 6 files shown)


VKTB:
Looks good, however it is still possible to sort by Datafile Count in generic dataset table view (not card view) but results in the expected rejected requests. Should this sorting be disabled in the table view too @patrick-austin ?

patrick-austin:
> 
> 
> Looks good, however it is still possible to sort by Datafile Count in generic dataset table view (not card view) but results in the expected rejected requests. Should this sorting be disabled in the table view too @patrick-austin ?

You're absolutely right, that should be disabled as well. I must have missed this when creating the issue, I've added that in the latest commit.

VKTB (approved):

PR #511, opened 2020-12-10
include lucene data in investigation requests #508

### LINKED ISSUE #508: select all doesn't take search into account
**Description**:
When using the select all checkbox in the search plugin, all investigations are added to the cart, not the ones that are displayed in the table. This doesn't affect datasets or files.

search all
cart all

You can also see this as when individually selecting each row, the select all checkbox appears as a dash (indicating only some items selected).

search individual

**Acceptance criteria**:
- [ ] The select all checkbox only adds the entities present in the search results

patrick-austin:
## Description
Added `luceneData` to `fetchCount` and `fetchAllIds` in the investigation tab of the search plugin, like we do for the other entities. 

## Testing instructions
- [x] Review code
- [x] Check Travis build
- [x] Check that adding investigations with the select all button in the search plugin only adds the investigations dispalyed in the table
  - This information isn't directly displayed in the search plugin, so use dataview or download to check what's in the cart (or check the state of `cartItems`.

## Agile board tracking
closes #508


VKTB (approved):
Works as expected. I am guessing that this case is not really possible to test using a unit or integration test given that it is not possible to access cart details from within the search plugin?

patrick-austin:
> 
> 
> Works as expected. I am guessing that this case is not really possible to test using a unit or integration test given that it is not possible to access cart details from within the search plugin?

Had a think about this, while it would be difficult to test the contents of the cart directly, the bug was also affecting how the select all checkbox worked which can be easily tested. Without the fix, selecting all individual rows would leave the select all box in the "indeterminate" state (see the original issue screenshots) as the select all box was adding/removing all entities not just those shown in the table. By testing that adding individual rows causes the select all checkbox to be ticked, we can indirectly test that the select all box is only adding/removing the data displayed in the table. I've added this test to all 3 tables for completeness.

patrick-austin:
Was waiting for the Cypress upgrade so I could set retries on the `search` E2E tests, this should also fix the failing tests on master (`card/studies` wasn't compatible with the new version) and (hopefully) make some other tests more reliable with their `wait`s.

VKTB (approved):

PR #521, opened 2020-12-16
#476 - fix Table duplication bug

### LINKED ISSUE #476: Duplicate datasets/files caused by sidebar
**Description**:
On pre-prod, can repeatedly duplicate the datasets and datafiles shown in dataview by opening and closing the side navigation bar. This only happens in table view, not card. "Verifying URL" briefly appears before the duplicate entries. Sorting or applying filters seems to reduce the number of duplicates, but not consistently. Checked all tables and it doesn't happen for datasets in "Generic", just DLS and ISIS. Happens for datafiles in all 3.

As part of #461 there was mention of sending duplicate requests, if this wasn't an issue in the past then maybe that's related? Haven't looked into this properly yet.

**Acceptance criteria**:
- [ ] Duplicates don't appear.

image

louise-davies:
## Description
Fixed duplication bug by extracting out "Safe" table components to non-inline components + memoizing them - preventing React from remounting them when a rerender is requested.

Additionally, added memoization to id check functions - this was initially a way to try and fix the bug but I kept it in even though it didn't contribute towards the fix as it does improve performance (see code comment)

And added id checking to the card views that require it

## Testing instructions

- [x] Review code
- [ ] Check Travis build
- [x] Mount dataview in scigateway and check that datafile tables do not get duplicate data when the sidebar opens/closes

## Agile board tracking
Closes #476


patrick-austin (approved):
Looks good, couldn't trigger the duplication anymore.

e2e test were failing for the datafile tables because we were waiting on requests that I assume due to the memoization we no longer send, so I removed them.

PR #522, opened 2020-12-16
map currentTab to state #479

### LINKED ISSUE #479: Switching tabs doesn't switch data type in search
**Description**:
When you switch tabs in datagateway-search, it doesn't always switch the data to what should be displayed in  the table (i.e. switching from investigation to datasets still shows the investigations in the datasets table).

I noticed this in the demos, so to recreate: first search for just investigations, then tick datasets and research, then switch tabs. These steps might not be necessary for replicating the bug, but it happened in both demos with the same steps so should be a fairly stable bug replication method.

**Acceptance criteria**:
- [ ] Switching between entity type tabs correctly switches the data displayed

patrick-austin:
## Description
Attempt to fix bug when switching search tabs. Changes to the order we set `datafileTab` etc. in state, and map the value of the tab selector to state. These modifications seemed to stop the bug for me, but as I couldn't reliably trigger it more work may be needed.

## Testing instructions
For information on how to cause the bug, see #479 

- [x] Review code
- [x] Check Travis build
- [x] While serving the plugin through SG, verify that only datasets are displayed in their tab when swapping from the investigation tab

## Agile board tracking
closes #479


louise-davies (dismissed):
I've deployed this to preprod and can't reproduce the issue after multiple attempts, so I think we'll call this one fixed :)

patrick-austin:
Resolved merge conflicts, but now needs re-review.

louise-davies (approved):

PR #530, opened 2021-01-05
Adjust height of download tables to prevent scrolling #523

### LINKED ISSUE #523: Download cart display requires scrolling
**Description**:
The display of the download cart requires to scroll to see the action buttons at the bottom of the page, even when the list of download entities is very short. This happens both when the menu is open and when the menu is closed. See screenshots.

**Acceptance criteria**:
- [ ] The download cart display shows the action buttons at the bottom of the page without requiring scrolling

Screenshot 2020-12-18 at 10 53 41

Screenshot 2020-12-18 at 10 55 39

patrick-austin:
## Description
Adjusted the heights in in the download tables (cart and history) to account for other elements on the page. Have also added an somewhat arbitrary `minHeight` of 230px which should allow for the headers and 4 rows. Prevented the elements below the cart table from starting a new line.

Result should be that the table takes up all the free space, but no more than that and so the only scroll bar is the one within the table (not for the window), unless you have a window shorter than around 500px, in which case the minimum table height is reached and a window scroll bar may appear.

## Testing instructions
- [x] Review code
- [x] Check Actions build
- [x] Check no (window) scroll bar appears on either table at reasonable (i.e. >500*500px) browser sizes
- [x] Check there's no empty space when served through SG, or that there's 64px spare in standalone mode (for the SG app bar)
- [x] Check that short windows (<500px) do not completely remove the table, but it remains at its minimum height and introduces a scroll bar for the window
- [x] Check that the information below the cart table doesn't create new lines for (somewhat) narrow windows (my browser wouldn't go narrower than 400px, below that it might force it onto a newline but windows wider than 400px should be OK)

## Agile board tracking
closes #523


VKTB (commented):
Looks good, however we might want to reduce the heights even more as there is a scrollbar when deploying SciGateway with the footer section. The PR that implements the footer section is not yet merged so it's best to pull the changes for that branch and test it that way.

The _Download Cart_ button goes onto a new line when the screen width goes below 350px but not sure if this is too much of an issue as only iPhones below iPhone 6 have smaller screen widths (not sure about Android phones). The text starts to create a new line when the screen width goes below 325px but again, I don't think that this will be an issue.

packages/datagateway-download/src/downloadCart/downloadCartTable.component.tsx
```diff
@@ -207,7 +207,16 @@ const DownloadCartTable: React.FC<DownloadCartTableProps> = (
...
         )}
         <Grid item>
-          <Paper style={{ height: 'calc(100vh - 150px)' }}>
+          {/* Table should take up page but leave room for: SG appbar, 
+              SG footer, tabs, table padding, text below table, and buttons
+              (respectively). */}
+          <Paper
+            style={{
+              height:
+                'calc(100vh - 64px - 24px - 48px - 48px - 3rem - (1.75 * 0.875rem + 12px)',
```
VKTB, line 216: There is still a scroll bar as shown in the screenshot below. I think you will need to subtract a few more pixels from `100vh` like I did in my #535 PR to fix this. You will notice in there that I subtracted `180px` despite the total pixels for the 4 components (AppBar, Footer, cart icon and card toggle) being `174px`. I could not understand why this was happening.

image

packages/datagateway-download/src/downloadStatus/downloadStatusTable.component.tsx
```diff
@@ -200,7 +200,14 @@ const DownloadStatusTable: React.FC<DownloadStatusTableProps> = (
         </Grid>
       )}
       <Grid item>
-        <Paper style={{ height: 'calc(100vh - 110px)' }}>
+        {/* Table should take up page but leave room for: SG appbar, SG footer,
+            tabs,table padding, and text above table (respectively). */}
+        <Paper
+          style={{
+            height: 'calc(100vh - 64px -24px - 48px - 48px - (1.75rem + 40px))',
```
VKTB, line 207: Not sure if I tested this and then forgot to mention it in my tests review, but the height of the status table appears to have been drastically reduced, see screenshot below.
image

VKTB (changes requested):

packages/datagateway-download/src/downloadCart/downloadCartTable.component.tsx
patrick-austin: Ah I see, that's my fault for not actually testing this myself and relying on unit tests; I assumed adding in the size of the footer was such a simple change it wouldn't break anything...

Had a look into what causes the extra space, and it's the [padding from the routing component](https://github.com/ral-facilities/scigateway/blob/master/src/routing/routing.component.tsx#L25) which is 30px. 

I'll up the padding here to 30, and since you've changed it in dataview I'll have a look at `search` and change the height there as part of this PR as well.

packages/datagateway-download/src/downloadStatus/downloadStatusTable.component.tsx
patrick-austin: I think `calc()` doesn't work with "negative" pixel values, so `- 24px` is fine but `-24px` is breaking it, my fault for not actually testing the change (again).

packages/datagateway-download/src/downloadCart/downloadCartTable.component.tsx
VKTB: Does the same test criteria applies for the **search** plugin changes - _Check no (window) bar appears at reasonable (i.e. >500*500px) browser sizes_? Because the vertical scroll bar appears at 820px window width. Also I noticed (but don't think that it's as a result of the changes that you have made) that the **search** plugin table displays scrollbars at 558px window height but the records can't be seen when you try to scroll down. Reducing the window height below this results in the table height reducing and eventually getting removed.

patrick-austin: I've made a few more changes, so now the search box/table have a minimum width of 600px (584 + 16 padding), which is the xs/sm screen breakpoint. Below that you will get a horizontal scrollbar. For me nothing jumps onto a new line, but I suspects that's dependent on browser settings. For example when running the plugin standalone there's quite a lot of space to play with, but serving it through SG chooses a different font which is wider and so there's no spare pixels in the layout. Equally increasing the font size (my browser was set to 16) will probably force newlines.

In terms of height, I added a minimum height to the search tables like I did for the downloads, so it should always display 3-4 rows at one time. This now means the vertical scrollbar appears sooner (for me, 662 px when served through SG).

While there's still the potential for this to be broken when upping the browsers font size, I can't think of much more we can do without fundamentally changing how we're laying things out.

image

VKTB: This looks good now, thanks for making the changes.

VKTB (approved):
Looks good and changes work as expected.

Issue #123, opened 2020-02-28
Improve column resize behaviour


louise-davies:
The column resize behaviour is unintuitive due to the fact that it uses flex ratios instead of pixel widths. Investigations should be made into making this feature more intuitive.

Acceptance criteria:
- [ ] Columns can be resized more intuitively.

PR #531, opened 2021-01-05
Disable text (and flex) wrapping for (date) headers #526

### LINKED ISSUE #526: Stop icon and title elements of date column filter from wrapping
**Description**:
When narrowing the browser window, while in the table view, the date column filter title breaks to the next line and as a result the _To_ input/ date picker element gets hidden.

image

**Acceptance criteria**:
- [ ] _To_ input/ date picker element can be seen when the browser window is narrow

patrick-austin:
## Description
Disabled text wrapping (i.e. between words) for all table headers, and disable flex wrapping (between icon and text) for date headers.

This should ensure that all headers take up a maximum of 3 lines (icon, text, selector; or for dates icon and text, from selector, to selector). This is consistent with the header height (110px) which can fit three lines worth of content. This height is static, so we can't can't automatically expand the header vertically for narrow windows.

narrow

Note that horizontal clipping still occurs, and it is possible to cut off all of the text for date selectors in very narrow windows (as they must now share a line with the icon). This will in turn prevent sorting for that field as the text cannot be clicked on. If that's an issue, possible solutions would be to either increase the minimum column size (currently 70px) or allow sorting when clicking on the icon of a header (not just the text as it is currently). The user can also simply manually resize the date columns to be wider as before. 

very narrow
resized

## Testing instructions

- [x] Review code
- [x] Check Actions build
- [x] Check that the table header selector cannot be vertically clipped in table view

## Agile board tracking
closes #526


VKTB (commented):
Changes look good and works as expected.

>Note that horizontal clipping still occurs, and it is possible to cut off all of the text for date selectors in very narrow windows (as they must now share a line with the icon). This will in turn prevent sorting for that field as the text cannot be clicked on. If that's an issue, possible solutions would be to either increase the minimum column size (currently 70px) or allow sorting when clicking on the icon of a header (not just the text as it is currently).

As you say, this is only an issue with the data selectors in very narrow windows. The text selectors are fine because they appear under the icons. I had a look and changing the minimum column width of the date selectors only to `100px` may fix this issue (see screenshot below) but I am not sure on the knock on effects that this may have. 
image

I think it will be best to get the views of others.

patrick-austin:
> I think it will be best to get the views of others.

The default width of 70px was set in #123 / #327 by @louise-davies so she might have an opinion? I suppose a benefit of having a smaller minimum size is that it gives more leeway when resizing another column to be wider. Something else that could be done as well / instead is reducing the padding between columns (currently 16px from each for a total seperation of 32px) either to a smaller static value, or having it as a fraction of the screen width?

patrick-austin:
Have reduced the padding on data columns to 8px either side on sm and xs screens (<960px), this should trigger before hitting the minimum column size which I've increased to 84px. So at its narrowest it should display the first two letters of the header to allow sorting.

VKTB (approved):
Looks good

PR #540, opened 2021-01-18
Add exclude text filter #484

### LINKED ISSUE #484: Add exclude text filter
**Description**:
Sylvie mentioned that an exclusive text filter would be useful for users in Diamond. In and of itself this should be simple, but then some functionality to be able to select between the inclusive and exclusive filters would be needed. This could then be extended in the future to other types of text filters.

**Acceptance criteria**:
- [ ] Users can filter by exclusive text (e.g. title NOT containing "dog")
- [ ] Users can switch between different text filter types.

VKTB:
## Description
Modifies the Text Column Filter component so that users can select an include or exclude filter type in the text filters. The exclusive filter was requested by Diamond and was primarily for the for dataview, but it makes sense to add the functionality to all of the instances of the Text Column Filter component in all the three plugins. There is now a Settings icon in the text filter that users can click on to select the filter type that they want. The placeholder text in the text filter changes from _Include..._ to _Exclude..._ and vice versa - depending on the filter type selected. The include filter displays all the records that contain the filter value, whereas the exclude filter displays all the records that do not contain the filter value. The views are not filtered if a filter type is selected while the filter value is empty.
image

Please note that the `renders data columns correctly` unit test in `src/table/table.component.test.tsx` is failing due to the actual value not matching the expected value.
```
 FAIL  src/table/table.component.test.tsx (35.592s)
  ● Table component › renders data columns correctly

    expect(received).toEqual(expected) // deep equality

    Expected: "Test 1"
    Received: "Test 1​"

      67 |         .first()
      68 |         .text()
    > 69 |     ).toEqual('Test 1');
         |       ^
      70 |
      71 |     expect(
      72 |       wrapper

      at Object.<anonymous> (src/table/table.component.test.tsx:69:7)
```
Visually, the actual and expected values look the same however it was discovered that the actual value contains an extra character. I spent some time investigating but I was unable to find what is causing this. If anyone knows or can spot anything obvious, then please let me know. I have disabled this test so that I can get the e2e tests to run.

## Testing instructions
- [x] Review code
- [ ] Check Actions build
- [x] Settings icon present in text filters
- [x] The default filter type is _Include_
- [x] Menu shows upon clicking on the Settings icon
- [x] Selecting a filter type from the menu changes the placeholder text in the text filter accordingly i.e. selecting _Exclude_ changes the placeholder text to _Exclude..._
- [x] Selecting a filter type while the filter value is empty (nothing is inputted in the text filter) does not filter the table/ card view.
- [x] Filtering a column by a specific value while the include filter is selected results in the table/ card view component displaying records that contain that value (Check this is the case in all table and card views in dataview, search and download).
- [x] Filtering a column by a specific value while the exclude filter is selected results in the table/ card view component displaying records that do not contain that value (Check this is the case in all table and card views in dataview, search and download).

## Agile board tracking
closes #484


louise-davies (approved):
Looks good and works well for me - just have some conflicts to resolve first

VKTB:
The CI build fails because `datagateway-common` cannot be compiled(?) as a result of the `objectChanged` function in `index.tsx` failing. I think this is happening because this PR has changed the definition of `Filter`. I had a look but did not get too far, @patrick-austin would you be able to help with this, I can see that you wrote this function?

patrick-austin:
> 
> 
> The CI build fails because `datagateway-common` cannot be compiled(?) as a result of issues with `objectChanged` in `index. I think this is happening because this PR has changed the definition of `Filter`. I had a look but did not get too far, @patrick-austin would you be able to help with this, I can see that you wrote this code?

So `objectChanged` had three main cases, strings and numbers which could be directly compared, arrays of strings which we have to iterate over, and anything else was a `DateFilter` where we check startDate and endDate. Now, we don't have strings/numbers as options for FilterType, and since the new `TextFilter` type is an object it's being cast as a `DateFilter` by the `objectChanged` function and this is what's stopping it from compiling.

As far as I can see we'd either need to add a check for the type of object we're dealing with at the end of objectChanged and treat `TextFilter`s and `DateFilter`s seperately, OR have a more generic check of the object where we just iterate over its keys without ever casting it or specifying what those keys are. But I haven't actually run the code so any solution might need to be more involved. Let me know if you want me to rewrite it or if you'd prefer to handle it, I don't mind either way.

VKTB:
@patrick-austin Thank you for your explanation and ideas. I tried writing a generic check to cover both `TextFilter` and `DateFilter` but I kept on getting TypeScript errors. As a result, I wrote the logic in `8027eaa` that does separate checks. I can't say I like the changes that I did, but given that it compiled and all tests passed. I assume that it meets the desired outcome.

patrick-austin:
> 
> 
> @patrick-austin Thank you for your explanation and ideas. I tried writing a generic check to cover both `TextFilter` and `DateFilter` but I kept on getting TypeScript errors. As a result, I wrote the logic in `8027eaa` that does separate checks. I can't say I like the changes that I did, but given that it compiled and all tests passed. I assume that it meets the desired outcome.

@VKTB I've checked out the branch and had quick test, while the `objectChanged` function is compiling fine it's no longer preventing the duplicate requests when a filter changes and I can't immediately see a reason why not. Given that this doesn't affect the exclude filter functionality and I don't know how long it would take me to get to the bottom of this, would you be happy to merge this as is and then I'll create an issue/branch/PR for fixing the `objectChanged` function to handle the new TextFilters? In terms of reviewing this PR I've not looked at anything before @louise-davies' approving review but the commits since then look fine so I can approve it now?

packages/datagateway-common/src/state/actions/index.tsx
```diff
@@ -158,8 +159,17 @@ const objectChanged = (
...
+          parsedEntry.startDate !== (stateEntry as DateFilter).startDate) ||
+        ('endDate' in parsedEntry &&
+          parsedEntry.endDate !== (stateEntry as DateFilter).endDate)
+      ) {
+        changed = true;
+      } else if (
+        ('value' in parsedEntry &&
+          parsedEntry.value !== (stateEntry as TextFilter).value) ||
+        ('type' in parsedEntry &&
+          parsedEntry.type !== (stateEntry as TextFilter).value)
```
patrick-austin, line 172: Should be `parsedEntry.type !== (stateEntry as TextFilter).type)`

patrick-austin (changes requested):

packages/datagateway-common/src/state/actions/index.tsx
VKTB: Oops. Good catch, thanks.

patrick-austin (approved):
I'm happy with the changes since @louise-davies' approving review, and looks like it works to me.

VKTB:
Thanks @patrick-austin. @louise-davies are you happy for this to be merged?

Issue #537, opened 2021-01-12
Filters visually persist between tables


patrick-austin:
**Description**:
Filter (but not sort) values visually persist between tables if breadcrumbs are used to navigate and the tables share a field (but the table contents isn't actually filtered):

<img width="960" alt="datafile filter" src="https://user-images.githubusercontent.com/61705287/104325871-fed96680-54e0-11eb-903d-2e0bf3ffa618.png">
<img width="960" alt="browser back" src="https://user-images.githubusercontent.com/61705287/104325880-00a32a00-54e1-11eb-8851-28357d6aa806.png">
It also seems to occur when using browser back buttons, however that may be fixed by (or at least affected by) the changes to querying as part of #461. So probably best to wait until that PR is merged before trying to fix this. I originally noticed this on that branch, but am making this a seperate issue to prevent further increasing the size of that PR.

**Acceptance criteria**:
- [ ] Navigating via any means (table links, breadcrumbs, browser) ensures that the filters' appearances matches the URL/table contents
- [ ] Compatible with upcoming changes to querying from #461

Issue #461, opened 2020-11-04
TODOs from cardView


patrick-austin:
**Description**:
There were a number of TODOs left in the cardView PR, and unticked boxes from #128 that might not have been addressed. They may vary in terms of importance, and some might not be actioned in the end. It may be easiest to create seperate issues as these are scoped/worked on?
- [x] **useMemo vs useEffect**: `datasetCardView` (but probably applies to multiple places). My understanding is useMemo triggers less renders, but I'm not sure what all the pros/cons are
- [ ] **Refactor add/remove buttons**: `datasetCardview` (but applies elsewhere). We have the same button defined in multiple places, would be more convenient to style it once and re-use by importing
- [x] **Duplicate count and data requests**: `cardView`. One of the "main items to fix", caused by multiple useEffects interfering with each other
- [x] **Filter panel selected items**: `cardView`. The card only filter panel starts "closed" even when filters are in effect due to re-rendering
- [x] **No results display**: `cardView`. When there are no results, we still display pagination, sort/filter panel
- [x] **Separate queries for card/table**: `dgcommon.reducer`. Card and table have distinct queries (i.e. filters and sorting), so the results and their order can be different between card/table view. This was intended in case the queries needed seperate handling, but it might be more intuitive to share sorting etc. between views.
- [x] **Advanced filters in small window**: from #128. Couldn't replicate this, but that might be due to pagination being set to both (or it could have been fixed already?)
- [x] **Rename filterTable to filterData**:  from #128. Both terms are used in the code, with the action `filterTable` being called via `pushPageFilter`, and `filterData` being the possibilities for the card only filtering in the state. It seems to work as is, so this could be an outdated comment?
- [ ] **Move elements to common**: from #128. Specifically `card` and `cardView`, related to #420. 
- [x] **Narrow window considerations**: from #128. When too narrow to display cards and filters/sorts side by side, the components should adjust to use the full width (lots of whitespace currently)
- [x] **UI while fetching**: from #128. Disable filters/sort/pagination/result # while fetching results?
- [x] **Query and browser navigation**: from #128. Navigating back/forwards using browser buttons doesn't apply the filters/sort from the url query (until manually refreshing).

**Acceptance criteria**:
- [x] Scope potential TODOs, and decide what's a priority

patrick-austin:
- [ ] **Timing of of `fetchDownloadCart` in `pageContainer`**

[...]

PR #543, opened 2021-01-22
Feature/cardview query issues #461

patrick-austin:
## Description
Various modifications to how requests for data are sent, with the intent of resolving the following issues:
- Duplicate requests to fetch the count and data in card view
- Sorting and common filters not persisting between card and table views
- Browser navigation not applying query information

The changes themselves:
- Unified `filters` and `sort` into the `query` object, which makes it easier to watch for and update changes to `location.search` (i.e. `?view=...&filter=...`, not the currently unused `query.search`) in a single `pushQuery` action
- Changed `savedView` to `savedQuery` which similarly contains all the relevant values
- When swapping views, sort and non-"custom" filters are kept in the current query (values which only affect or can be altered in card view like `page` or `results` are still "saved" as before)
- Put routing of card/tables into one file, and included the view in the url/state for datafiles (which always display as a table) so that we still have access to it when using breadcrumbs navigation
- Load query when either `location.path` or `locatation.search` changes (before we just did it on `path`)
- Logic to prevent triggering `useEffects` when "updating" arrays and objects which haven't _actually_ changed 
- Only fetching `count` and `AllIds` when filters change
- Changed card view's "loading" logic to depend on a state boolean controlled by the sending/recieving of requests rather than logic in the component itself
- Reduced the number of props for card view by using `query` rather than `sort`, `page`, `results`...
- Reduced the use of `React.useState` in cardView to track things which are already mapped to the dgcommon state, and the functions needed to update them
- Calculate any page changes _before_ sending requests for data with outdated indices
- Replaced anonymous callbacks for fetching data/counts with memoized version to prevent needless re-renders
- Include `?view=...` in links between entities so we can visit card views directly rather than visiting a table _then_ loading the view and pushing to a new URL
- Changes to the tests so they expect or wait for the right (number of) actions
- Fixed a typo in studies which was causing too much data to be fetched
- Small changes to semi-related things which appeared to be wrong (e.g. setting `loading: false` when sending a request as opposed to when it succeeds/fails)

Things not addressed:
- `dgcommon.loading` is set to `false` when the first request succeeds/fails, which is often premature as we have multiple requests at once
- #537 
- Aspects of #461 relating to refactors

## Testing instructions
As a larger PR it may be beneficial to review this "in person", who (@louise-davies @VKTB @MRichards99) was intending to have a look at this? I can arrange a meeting for VSCode liveshare, screensharing or just discussion for whenever works best for the people involved?

- [x] Review code
- [x] Check Actions build

Card view:
- [x] Forcing a reset to `page: 1` does not result in duplicate requests
- [x] No "flickering" of UI or premature message about filters on loading

For both card and table views:
- [x] Studies now only have ~1 investigation displayed rather than ~10 (should be taking the STUDY.ID into account not just that of the INSTRUMENT
- [x] Navigating to new entities only causes single requests
- [x] Navigating with breadcrumbs only causes single requests (especially between datafiles and another entity in CV)
- [x] Filtering causes single requests for count (IDs) and data
- [x] Sorting causes only single requests for data
- [x] Swapping views persists common filters and sort, but NOT CV specific values (e.g. page and TYPE_ID)
- [x] Using browser navigation causes the query in the URL to be applied to the table/cards
- [x] Functionality otherwise unaffected

## Agile board tracking
connect to #461


louise-davies (approved):
Looks good, all the manual tests you specified worked and I tested the "what happens if a user refilters for something with the exact same number of results" use case we spoke about in the meeting and the new loadCount boolean does cause correct behaviour in that case :)

PR #545, opened 2021-01-27
Display banner to users that they need to login in my data view

### LINKED ISSUE #481: Display banner to users that they need to login in my data view
**Description**:
Due to the autologin functionality, one of the only reasons for users to login may be the My Data view. However, if a user navigated to this whilst logged in anonymously they will have no data, which may cause confusion if they forget they haven't logged in. The My Data pages should thus detect if a user is autologged in/logged in anonymously, and display a message to remind the user that they will need to login to see their own data.

Autologin code is in SciGateway and it would be best to take a look there to see how the process is handled before tackling this issue.

**Acceptance criteria**:
- [ ] Message of some sort is displayed to the user if they attempt to look at My Data whilst logged in automatically/as anon

VKTB:
## Description
The My Data pages detect if a user is autologged in/logged in anonymously, and display a message to remind the user that they will need to login to see their own data. The banners are displayed when the DataGateway-dataview plugin is deployed together with SciGateway but due to a bug in SciGateway, they are displayed multiple times. For example, when you first access a My Data page, two banners will displayed, then three banners in addition to the two will be displayed if you revisit the page, and so on. I opened the [Prevent duplicate banners from displaying multiple times](https://github.com/ral-facilities/scigateway/issues/542) issue in SciGateway to address this bug.

## Testing instructions
- [x] Review code
- [x] Check Actions build
- [x] Deploy DataGateway-dataview together with SciGateway
- [x] Navigate to the My Data pages and check that the banner is displayed (note that the banner will be displayed multiple times as stated above)

## Agile board tracking
closes #481


VKTB:
There is a fix for the failing DataGateway Search e2e test in #543, so I will merge master into this branch once that PR is merged.

patrick-austin (commented):
While testing I was able to get the warning to display while logged in using `jwt`:
image
I managed this by first auto logging in using `icat` as my `auth-provider`, which is the only provider to set `autoLogin` in the browser's storage. Then switching to `jwt` without clearing local storage will leave `autoLogin` as true (since the `jwtAuthProvider` functions never touch it). 
So the warning _can_ display when logged in, but I don't think this would ever be an issue in real usage as the auth provider is only set in `settings.json` which users wouldn't have access to change. If we (or whoever was running a SciGateway instance) did decide to change from `icat` to `jwt` while the site was live then existing users would have to clear their local storage to get rid of the `autoLogin` boolean, or if we though this might happen we could put some code into `jwtAuthProvider` which sets `autoLogin` to be false/removes it from local storage. 
Either way the workaround would be simple and it's so unlikely that I don't think it's a problem and can be merged as is once the failing e2e test gets the fix from the other branch.

VKTB:
The failing e2e is passing now @patrick-austin

patrick-austin (dismissed):

VKTB:
@patrick-austin Needs a re-review as I needed to resolve some conflicts.

patrick-austin (approved):

PR #548, opened 2021-02-03
Give routing dummy URL while navigating #537

### LINKED ISSUE #537: Filters visually persist between tables

patrick-austin:
## Description
When navigating the hierarchy using breadcrumbs/browser buttons, if there was a query set (e.g. filters applied) then requests were sent with the new query for the old entity (for browser navigation) or the old query with the new entity (for breadcrumbs). A side effect of this was the visual appearance of the filters being outdated.

To prevent this, have modified the `location` we pass to the router so that while we are in the process of navigating we use a dummy url so that the next entity table/card isn't loaded until the query is updated. This prevents the router and query in state getting out of sync. This also meant we need to pass the view in table links to ensure we always update the query with everything we need after the location changes.

## Testing instructions

- [x] Review code
- [x] Check Actions build
- [x] Ensure that swapping card/table view works without additional causing duplicate requests
- [x] Ensure that normal navigation (table/card links) works in card/table views with and without queries, without causing duplicate requests
- [x] Ensure that navigation using breadcrumbs works in card/table views, with and without queries, without causing duplicate requests
- [x] Ensure that navigation using browser back buttons works in card/table views with and without queries, without causing duplicate requests

## Agile board tracking
closes #537

### DIFF
--- a/packages/datagateway-common/src/state/reducers/dgcommon.reducer.test.tsx
+++ b/packages/datagateway-common/src/state/reducers/dgcommon.reducer.test.tsx
@@ -186,6 +186,14 @@ describe('DGCommon reducer', () => {
         sort: { NAME: 'asc' },
         filters: { NAME: 't' },
       },
+      savedQuery: {
+        view: 'card',
+        search: 'searchTwo',
+        page: 2,
+        results: 2,
+        sort: { NAME: 'desc' },
+        filters: { NAME: 'c' },
+      },
     };
 
     const updatedState = DGCommonReducer(state, clearTable());
@@ -196,14 +204,8 @@ describe('DGCommon reducer', () => {
       loading: false,
       downloading: false,
       error: null,
-      query: {
-        view: 'table',
-        search: 'searchOne',
-        page: 1,
-        results: 1,
-        sort: { NAME: 'asc' },
-        filters: { NAME: 't' },
-      },
+      query: initialState.query,
+      savedQuery: initialState.savedQuery,
     });
   });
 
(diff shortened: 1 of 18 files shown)


VKTB (approved):
Works as expected. One thing that I noticed however when toggling from table to card view while in `/browse/investigations`, is that two `datagateway_common:fectch_filter_request` requests are sent. Is this expected?

patrick-austin:
> 
> 
> Works as expected. One thing that I noticed however when toggling from table to card view while in `/browse/investigations`, is that two `datagateway_common:fectch_filter_request` requests are sent. Is this expected?

Yeah we send one for `TYPE_ID` and one for `FACILITY_ID` so it's not the same "duplication" as requesting the count, IDs or data twice.

VKTB:
>Yeah we send one for TYPE_ID and one for FACILITY_ID so it's not the same "duplication" as requesting the count, IDs or data twice.

Ah yes, you are right, that's different.

patrick-austin:
@VKTB Need a re-review as I had conflicts with the addition of the TextFilter in the imports.

VKTB (approved):

PR #549, opened 2021-02-04
Refactor cardView and filter warning on tables #461

patrick-austin:
## Description
Refactoring of cardView, cardEntity and advancedFilter into common so we can re-use them in the future. As a lot of the cardView file was implicitly tested by other files in dataview (e.g. investigationCardView) this means creating a dedicated test file in common to keep the coverage up. Also needed to move the material-ui/lab package to common, and add in i18 for translations.

Also added the filter warning that displays in card view to table view, and moved cart fetching from didUpdate to didMount (as we were only calling it once anyway.

Haven't refactored the add to cart / download buttons yet as that's dependent on #547 

## Testing instructions
- [x] Review code
- [x] Check Actions build
- [x] Check that the message about filters displays in table view when applying a filter which results in no data being found (and that it doesn't show otherwise)
- [x] Check that we still fetch the card when viewing data (should see the first two investigations checked by default provided the cart hasn't been modified and the request succeeds)
- [x] Check cardView still displays as it used to (no unintended effects of the refactor) 

## Agile board tracking
connect to #461


VKTB (approved):
Looks good and works as expected.

patrick-austin:
@VKTB conflict merging needs a re-review

VKTB (approved):

PR #552, opened 2021-02-09
Translate download status strings, update filters and tests #486

### LINKED ISSUE #486: Add ability to "translate" download availabilities
**Description**:
In the download status table, the "status" column is just the value stored in TopCAT but capitalised normally. Chris however said it might be useful to "translate" these values to more meaningful ones (eg. "on tape" rather than "archived"). TopCAT I think did this statically, and more consideration is needed on how you can filter the values when the values in the table don't match the values in the database (you'd need to "translate" the partial strings somehow instead). Probably more discussion is needed and more looking into what TopCAT does.

**Acceptance criteria**:
- [ ] Ability for an admin to configure more meaningful availability labels

patrick-austin:
## Description
Added translation strings for the possible values of `Download.status` based on what is currently done in topcat (https://github.com/icatproject/topcat/blob/master/yo/app/languages/lang.json#L375). This also involved changing how we handle casing in the filters; before we displayed the `status` (which was always uppercase in the data) with standard capitalisation, and forced any filter input to upper case for the string comparisons. Now, since translation and casing is handled by `res/default.json`, we make both the filter input and displayed string lower case when comparing them. Functionally this should behave the same i.e. the case of the filter is ignored.

## Testing instructions

- [x] Review code
- [x] Check Actions build
- [x] In the download status table, check filtering for display text that appears in the status field filters correctly irrespective of input casing (e.g. available, AVAILABLE, AvAiL etc. all work)
- [x] In the download status table, check filtering for text of the original `status` value doesn't work, for example we map `COMPLETE` to `Available`, so (include) filtering for `complete` should not return anything 

## Agile board tracking
closes #486

### DIFF
--- a/packages/datagateway-download/src/downloadStatus/downloadStatusTable.component.tsx
+++ b/packages/datagateway-download/src/downloadStatus/downloadStatusTable.component.tsx
@@ -4,7 +4,7 @@ import { Grid, Paper, IconButton, LinearProgress } from '@material-ui/core';
 import {
   Table,
   Order,
-  Download,
+  FormattedDownload,
   TextColumnFilter,
   TableActionProps,
   DateColumnFilter,
@@ -35,7 +35,7 @@ const DownloadStatusTable: React.FC<DownloadStatusTableProps> = (
       | { value?: string | number; type: string }
       | { startDate?: string; endDate?: string };
   }>({});
-  const [data, setData] = React.useState<Download[]>([]);
+  const [data, setData] = React.useState<FormattedDownload[]>([]);
   const [dataLoaded, setDataLoaded] = React.useState(false);
 
   const { refreshTable, setRefreshTable, setLastChecked } = props;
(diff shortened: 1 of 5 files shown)


packages/datagateway-download/src/downloadStatus/downloadStatusTable.component.tsx
```diff
@@ -237,15 +262,15 @@ const DownloadStatusTable: React.FC<DownloadStatusTableProps> = (
...
-                },
+                // cellContentRenderer: (props: TableCellProps) => {
+                //   if (props.cellData) {
+                //     const status: string = props.cellData;
+                //     return (
+                //       status.substring(0, 1).toUpperCase() +
+                //       status.substring(1).toLowerCase()
+                //     );
+                //   }
+                // },
```
VKTB, line 273: Should this code be removed?

patrick-austin: Yes it should, thanks.

VKTB (approved):

PR #559, opened 2021-02-11
Anon login for all DG cypress tests #554

### LINKED ISSUE #554: Use anonymous login for E2E tests
**Description**:
Currently tests involving the download cart can interfere with each other because they share a login. Using anonymous login with different sessions would prevent this.

**Acceptance criteria**:
- [ ] Tests use anonymous login
- [ ] Can run tests/app concurrently without sharing a cart

patrick-austin:
## Description
Changed the cypress login command to use anonymous login, which allows the carts used in concurrent E2E to be treated seperately, which should reduce errors in E2E tests.

The url is currently hardcoded in `cypress/support/commands.js` as we were already doing this in `downloads` and we don't have a dedicated setting for authUrl  (`8000` isn't accessible outside the RAL network so fails on integration tests?). We were previously using the apiUrl to authenticate in `search` and `dataview`.

## Testing instructions

- [x] Review code
- [x] Check Actions build
- [x] To manually test, run `yarn e2e:interactive` in two dataview terminals and run the cartSelection test in both at the same time. All tests should pass on electron (using chrome I got errors when scrolling). For reference when running concurrent tests with the old login method I got 1-2 outright failures and ~5 tests passing with retries of the 48 in the cartSelection file.

## Agile board tracking
closes #554

### DIFF
--- a/packages/datagateway-dataview/cypress/support/commands.js
+++ b/packages/datagateway-dataview/cypress/support/commands.js
@@ -55,13 +55,21 @@ export const readSciGatewayToken = () => {
 };
 
 Cypress.Commands.add('login', () => {
-  cy.request('POST', 'http://scigateway-preprod.esc.rl.ac.uk:8000/login', {
-    mnemonic: 'anon',
+  cy.request({
+    method: 'POST',
+    url: 'https://scigateway-preprod.esc.rl.ac.uk:8181/icat/session',
+    body: {
+      json: JSON.stringify({
+        plugin: 'anon',
+      }),
+    },
+    form: true,
   }).then((response) => {
     const jwtHeader = { alg: 'HS256', typ: 'JWT' };
-    const payload = JSON.parse(parseJwt(response.body));
-    // Fake the username for my-data tests
-    payload.username = 'Robert499';
+    const payload = {
+      sessionId: response.body.sessionId,
+      username: 'Robert499',
+    };
     const jwt = jsrsasign.KJUR.jws.JWS.sign('HS256', jwtHeader, payload, 'shh');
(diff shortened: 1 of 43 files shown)


packages/datagateway-dataview/cypress/support/commands.js
```diff
@@ -54,26 +54,24 @@ export const readSciGatewayToken = () => {
...
-        payload,
-        'shh'
-      );
-
-      window.localStorage.setItem('scigateway:token', jwt);
-    });
+Cypress.Commands.add('login', () => {
+  cy.request({
+    method: 'POST',
+    url: 'https://scigateway-preprod.esc.rl.ac.uk:8181/icat/session',
```
louise-davies, line 60: It's probably worth adding a TODO to switch this back to the "old" method once we switch over to use the ICAT backend on preprod rather than DB backend, because then that will support anonymous login properly.

packages/datagateway-search/cypress/support/commands.js
```diff
@@ -54,26 +54,24 @@ export const readSciGatewayToken = () => {
...
-        payload,
-        'shh'
-      );
-
-      window.localStorage.setItem('scigateway:token', jwt);
-    });
+Cypress.Commands.add('login', () => {
+  cy.request({
+    method: 'POST',
+    url: 'https://scigateway-preprod.esc.rl.ac.uk:8181/icat/session',
```
louise-davies, line 60: Same as the other, a TODO is probably useful

louise-davies (approved):

louise-davies (approved):

VKTB (approved):

PR #567, opened 2021-02-15
investigationSearchTable links take facility into account #566

### LINKED ISSUE #566: Configure links in search plugin to match facility
**Description**:
Currently, links in the search table(s) link to `/browse/investigation/...`, the "generic" format for the entity hierarchy. For ISIS, DLS or other users with a different hierarchy, we'd want to be able to link to their respective url.

**Acceptance criteria**:
- [ ] Links in search can be made to point to the right hierarchy

patrick-austin:
## Description
Extract the hierarchy from the url of the search plugin and use it to format links in the investigationSearchTable with the correct hierarchy. `/search/dls` and `/search/isis` will give their respective links, anything else (e.g. `/search/data`) will default to the generic hierarchy link. ISIS relies on possibly undefined values, so if any can't be found then it will just render text rather than a link. When serving search through SG, would therefore expect the link registered to be changed from `/search/data` to `/search/isis` for example.

Also mocked `lodash.debounce` in the search unit tests like we do in the other plugins so we can test the text filters. This is unrelated to the issue, but the increased coverage made it easier to check branching statements I'd introduced were covered by unit tests.

## Testing instructions
To check validity of links, either serve both dataview and search through SG or run both dataview and search standalone and just modify the port number after clicking the link in search to check the url is valid in dataview.

- [x] Review code
- [x] Check Actions build
- [x] Check generic links go to valid generic url
- [x] Check DLS links go to valid DLS url
- [x] Check ISIS links go to valid ISIS url

## Agile board tracking
closes #566

### DIFF
--- a/packages/datagateway-search/src/__snapshots__/searchPageContainer.component.test.tsx.snap
+++ b/packages/datagateway-search/src/__snapshots__/searchPageContainer.component.test.tsx.snap
@@ -1,24 +1,7 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
 exports[`SearchPageContainer - Tests renders correctly at /search/data route 1`] = `
-<div
-  computedMatch={
-    Object {
-      "isExact": false,
-      "params": Object {},
-      "path": "/",
-      "url": "/",
-    }
-  }
-  location={
-    Object {
-      "hash": "",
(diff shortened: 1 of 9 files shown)


VKTB (approved):
Works as expected

PR #575, opened 2021-02-22
Use env.production to provide prefix for settings files #569

### LINKED ISSUE #569: Enable static files to be located at URL prefixes
**Description**:
@ajkyffin suggested that it would be nice to be able to host plugins at `/plugins/datagateway-x` but this would require static files (such as settings files & translation files) to be located at not `/` which is where they're currently expected. They are not currently configurable, and for the settings file it's particularly problematic because how to you configure where to load the settings file without the settings file!

I think one good solution is to use environment variables (we can use `.env` files - see here: https://create-react-app.dev/docs/adding-custom-environment-variables/) as these are injected at build time.

**Acceptance criteria**:
- [ ] Ability to change where static files are loaded from

patrick-austin:
## Description
Have added a check for an environment variable to set the directory used to load the settings file and `res/default.json` (these were the two files mentioned in the issue and I haven't seen any other files with hardcoded locations but I may have missed something). This is specified in `.env.production` which is used when building the plugin but not when testing or running in standalone (note that because the E2E tests use the build command they are also affected by the value in this file). The value can either be changed directly or overridden by a `.local` version of the environment file, as described here https://create-react-app.dev/docs/adding-custom-environment-variables/#what-other-env-files-can-be-used. If a value isn't available we default back to `/` so development should be unaffected.

As mentioned above this is injected at build time. I had a look into doing this at runtime instead, and you could do it by running a .js file to collect the values of any environment variables in the `start` or `prestart` commands, then overwriting another file in your build folder which in turn is added to `index.html`. The drawbacks of this are that you still need to know where your build folder is to overwrite, and the script that holds the environment variables would need to be added to SG's build folder/index rather than the index of the plugins, so SG is then containing the code for the plugins to a certain extent. That along with the fact it's just a more complicated way of doing it led me to go with the on build approach.

Also I orginally didn't fully understand the issue, so if at any point I've got what we need from this change wrong let me know.

## Testing instructions
- [ ] Review code
- [ ] Check Actions build

For each plugin:
- [ ] Change the value of `REACT_APP_<PLUGIN>_BUILD_DIRECTORY` in `.env.production` or add an `.env.production.local` file with a different value
- [ ] Check that when run standalone, the custom directory is ignored and the plugin loads the settings/strings from the default location
- [ ] Check that when building and serving through SG, the custom path is used. In practice this means it will fail to load unless you create a copy of the files at the specified location in SG's public folder (rather than DG's build folder).

## Agile board tracking
closes #569

### DIFF
--- a/packages/datagateway-dataview/.env.production
+++ b/packages/datagateway-dataview/.env.production
@@ -1,1 +1,1 @@
-REACT_APP_DATAVIEW_SETTINGS_PATH=/datagateway-dataview-settings.json
+REACT_APP_DATAVIEW_BUILD_DIRECTORY=/
--- a/packages/datagateway-dataview/.env.production
+++ b/packages/datagateway-dataview/.env.production
@@ -1,1 +1,1 @@
-REACT_APP_DATAVIEW_SETTINGS_PATH=/plugins/datagateway-dataview/datagateway-dataview-settings.json
+REACT_APP_DATAVIEW_SETTINGS_PATH=/datagateway-dataview-settings.json
--- a/packages/datagateway-dataview/.env.production
+++ b/packages/datagateway-dataview/.env.production
@@ -1,0 +1,1 @@
+REACT_APP_DATAVIEW_SETTINGS_PATH=/plugins/datagateway-dataview/datagateway-dataview-settings.json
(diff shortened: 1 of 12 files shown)


ajkyffin (approved):
It works now when I set `.env.production` to
```
REACT_APP_DATAVIEW_BUILD_DIRECTORY=/plugins/datagateway-dataview/
REACT_APP_DOWNLOAD_BUILD_DIRECTORY=/plugins/datagateway-download/
REACT_APP_SEARCH_BUILD_DIRECTORY=/plugins/datagateway-search/
```

PR #578, opened 2021-02-23
Don't include routing when datagateway-dataview element missing #574

### LINKED ISSUE #574: Do not send requests if dataview element is not present
**Description**:
Requests are still sent when the DataView is accessed while the site is under maintenance. Changes need to be made so that no requests are sent if the `datagateway-dataview` element is not present.

**Acceptance criteria**:
- [ ] Requests are not sent when DataView is accessed while site is under maintenance 
- [ ] No unintended effects of the changes

patrick-austin:
## Description
Add a check on the presence of an element with `id="datagateway-dataview"` for including the routing in PageContainer. The effect of this is to prevent us from sending requests for data unless the element is present, like how we do for fetching the cart. This means we'll no longer get requests while in maintenance mode (see ral-facilities/scigateway#565).

No functional change expected.

## Testing instructions

- [x] Review code
- [x] Check Actions build
- [x] Check datagateway loads successfully as it did before 

## Agile board tracking
closes #574

### DIFF
--- a/packages/datagateway-dataview/src/page/__snapshots__/pageContainer.component.test.tsx.snap
+++ b/packages/datagateway-dataview/src/page/__snapshots__/pageContainer.component.test.tsx.snap
@@ -48,20 +48,7 @@ exports[`PageContainer - Tests displays the correct entity count 1`] = `
       aria-label="container-table"
       item={true}
       xs={12}
-    >
-      <ViewRouting
-        loadedCount={false}
-        location={
-          Object {
-            "hash": "",
-            "pathname": "/",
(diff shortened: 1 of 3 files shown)


VKTB (approved):
Works nicely. Thanks for doing this - I was planning of making a start on it after I finish all the admin page work.

patrick-austin:
> 
> 
> Works nicely. Thanks for doing this - I was planning of making a start on it after I finish all the admin page work.

No problem, I saw the issue and didn't have much else to work on so thought it get it out of the way.

PR #603, opened 2021-03-01
Prevent logging in successful unit tests #601

### LINKED ISSUE #601: Prevent logging / warning during unit tests
**Description**:
Following recent package updates, we now get more errors/warnings/logging that doesn't fail unit tests, but can clog the terminal and make it harder to debug legitimate failures. 

**Acceptance criteria**:
- [ ] No logging for passing unit tests

patrick-austin:
## Description

- Add `__mocks__/react-i18next.jsx` to common so we can mock translations for refactored cardview components
- Explicitly mock `console.error` in one of the cardView tests. The error is for using a dynamic variable to set whether an Accordion (aka ExpansionPanel) is expanded by default. I've had a look at how these panels are controlled, and I don't think we want to change our functionality here, so just surpressed the warning in the tests.
- Add `item` to `Grid` in cardView (no functional change expected)
- Treat id checks as asynchronous in `pageContainer` tests like we do in `pageRouting`
- In lots of tests, move mocked `axios.get` into `beforeEach` and/or add in mocking for `axios.post` and `axios.delete` (corresponding to adding and removing to/from cart) to prevent unhandled request logging
- Reduce size of `searchPageTable` snapshot by using `shallow` instead of `mount`
- Change how we treat disabled tabs in search, instead of removing them and using `none` as a placeholding (which leads to errors about having a value that doesn't correspond to a tab) use a `Tab` set to `display: 'none'`. This should achieve the same functionality as before but without the warnings.
 
## Testing instructions

- [ ] Review code
- [ ] Check Actions build
- [ ] Check no logging in successful unit tests
- [ ] Check functionality of tabs in search is unaffected: unchecking an entity and researching should remove that tab from the UI and the default table dispalyed should be the first of the three entities that's checked.

## Agile board tracking
closes #601

### DIFF
--- a/packages/datagateway-common/src/card/cardView.component.tsx
+++ b/packages/datagateway-common/src/card/cardView.component.tsx
@@ -760,7 +760,7 @@ const CardView = (props: CardViewProps): React.ReactElement => {
                 })}
               </List>
             ) : (
-              <Grid xs={12} md={8}>
+              <Grid item xs={12} md={8}>
                 <Paper className={classes.noResultsPaper}>
                   <Typography align="center" variant="h6" component="h6">
                     {t('loading.filter_message')}
--- a/packages/datagateway-search/src/searchPageTable.test.tsx
+++ b/packages/datagateway-search/src/searchPageTable.test.tsx
@@ -5,26 +5,33 @@ import { MemoryRouter } from 'react-router';
 
 import SearchPageTable from './searchPageTable';
 
-import { mount as enzymeMount } from 'enzyme';
-import { createMount } from '@material-ui/core/test-utils';
+import { mount as enzymeMount, shallow as enzymeShallow } from 'enzyme';
+import { createMount, createShallow } from '@material-ui/core/test-utils';
 import configureStore from 'redux-mock-store';
 import thunk from 'redux-thunk';
 import { initialState } from './state/reducers/dgsearch.reducer';
 import { dGCommonInitialState } from 'datagateway-common';
 import { setCurrentTab } from './state/actions/actions';
+import axios from 'axios';
 
 describe('SearchPageTable', () => {
   let mount: typeof enzymeMount;
+  let shallow: typeof enzymeShallow;
   let state: StateType;
 
   beforeEach(() => {
     mount = createMount();
+    shallow = createShallow({ untilSelector: 'Paper' });
 
     state = JSON.parse(
       JSON.stringify({ dgsearch: initialState, dgcommon: dGCommonInitialState })
     );
 
     state.dgsearch.requestReceived = true;
+
+    (axios.get as jest.Mock).mockImplementation(() =>
+      Promise.resolve({ data: [] })
+    );
   });
 
   it('renders SearchPageTable correctly before request', () => {
@@ -45,7 +52,7 @@ describe('SearchPageTable', () => {
         datasetTab: true,
         datafileTab: true,
         investigationTab: true,
-        currentTab: 'none',
+        currentTab: 'investigation',
       },
       searchData: {
         investigation: Array(1),
@@ -54,7 +61,7 @@ describe('SearchPageTable', () => {
       },
     };
     const mockStore = configureStore([thunk]);
-    const wrapper = mount(
+    const wrapper = shallow(
       <Provider store={mockStore(state)}>
         <SearchPageTable store={mockStore(state)} />
       </Provider>
@@ -69,7 +76,7 @@ describe('SearchPageTable', () => {
         datasetTab: true,
         datafileTab: true,
         investigationTab: true,
-        currentTab: 'none',
+        currentTab: 'investigation',
       },
     };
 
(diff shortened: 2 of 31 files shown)


packages/datagateway-common/src/card/cardView.component.tsx
```diff
@@ -760,7 +760,7 @@ const CardView = (props: CardViewProps): React.ReactElement => {
                 })}
               </List>
             ) : (
-              <Grid xs={12} md={8}>
+              <Grid item xs={12} md={8}>
```
sam-glendenning, line 763: I assume `item` is always going to be true, else this code will not work. Safety alternative is `item={item}`

packages/datagateway-search/src/searchPageTable.test.tsx
```diff
@@ -5,26 +5,33 @@ import { MemoryRouter } from 'react-router';
...
+import axios from 'axios';
 
 describe('SearchPageTable', () => {
   let mount: typeof enzymeMount;
+  let shallow: typeof enzymeShallow;
   let state: StateType;
 
   beforeEach(() => {
     mount = createMount();
+    shallow = createShallow({ untilSelector: 'Paper' });
```
sam-glendenning, line 24: Is this shallow snapshot being rendered to the correct selector? As I'm unsure of its exact function, could use clarification from Louise/Victor.

sam-glendenning (commented):
Overall this change looks good to me and not noticing any unnecessary output in terminal when running tests. From a code point of view it looks fine for the most part, just added a couple of comments in files for clarification. 

`datagateway-search/src/table/investigationSearchTable.component.test.tsx` still makes reference to `instrumentId` and `facilityCycleId` on one occasion when rendering an `<InvestigationSearchTable />` and I'm not entirely sure if it is intended or not (in other parts of the file, these attributes are removed). 

To summarise, it seems good to me but I won't approve yet because I would like a final once-over from someone on Monday (likely either Louise or Victor) to make sure everything is correct as I'm still unclear on the overarching functionality of some of these tests. I'll organise this on Monday.

packages/datagateway-common/src/card/cardView.component.tsx
patrick-austin: `item` is a boolean prop for the MUI `Grid` component which defaults to false. This is shorthand for `item={True}`, you can see the same thing elsewhere with the `container` prop.

https://material-ui.com/api/grid/#props

packages/datagateway-search/src/searchPageTable.test.tsx
patrick-austin: Can't really remember this one, but from looking at the snapshot for this test rendering until `Paper` stops the individual `InvestigationSearchTable`, `DatasetSearchTable`, `DatafileSearchTable` from rendering. Maybe these were causing some excess logging, or maybe it was just to make the snapshot smaller? If there was a reason to have them in the snapshot it could be reverted or just make sure it's being covered by the tests that actaully mount.

patrick-austin:
> 
> 
> 
> `datagateway-search/src/table/investigationSearchTable.component.test.tsx` still makes reference to `instrumentId` and `facilityCycleId` on one occasion when rendering an `<InvestigationSearchTable />` and I'm not entirely sure if it is intended or not (in other parts of the file, these attributes are removed).

At a guess I'd say that having them only caused errors when we fully mounted, so I removed all of those but missed the `'renders details panel correctly'` because it's shallow and so doesn't cause errors, is near the bottom of the file and if I was just find/replacing `instrumentId="4" facilityCycleId="5"` the fact it has a newline would mean it wouldn't show up.

Ultimately it shouldn't have them, as these only apply to `ISISInvestigationsTable` not `InvestigationSearchTable`, but it doesn't look like it's logging any warnings by leaving them in.

packages/datagateway-common/src/card/cardView.component.tsx
sam-glendenning: I understand the shorthand, just needed to know it would be true in all circumstances. I'll give it a tick.

packages/datagateway-search/src/searchPageTable.test.tsx
VKTB: Can't see any logging when changing it back to `mount` however the snapshot gets bigger as `InvestigationSearchTable`, `DatasetSearchTable`, `DatafileSearchTable` are rendered. Looking at the coverage, all lines are covered when using `shallow` so I would just leave it as it is.

VKTB:
> > `datagateway-search/src/table/investigationSearchTable.component.test.tsx` still makes reference to `instrumentId` and `facilityCycleId` on one occasion when rendering an `<InvestigationSearchTable />` and I'm not entirely sure if it is intended or not (in other parts of the file, these attributes are removed).
> 
> At a guess I'd say that having them only caused errors when we fully mounted, so I removed all of those but missed the `'renders details panel correctly'` because it's shallow and so doesn't cause errors, is near the bottom of the file and if I was just find/replacing `instrumentId="4" facilityCycleId="5"` the fact it has a newline would mean it wouldn't show up.
> 
> Ultimately it shouldn't have them, as these only apply to `ISISInvestigationsTable` not `InvestigationSearchTable`, but it doesn't look like it's logging any warnings by leaving them in.

@sam-glendenning As @patrick-austin says, these references do not cause any issues but please feel free to remove them in case you think they can cause confusion in future. The rest of the changes look good to me.

sam-glendenning (approved):
Taken out those unnecessary props (prevents confusion later and keeps things clean and tidy) and all tests are happy. All looks good to me.

PR #660, opened 2021-04-14
Add Admin Download table

### LINKED ISSUE #613: Add TopCAT admin download table
**Description**:
Implement the admin download table functionality.

**Acceptance criteria**:
- [ ] Admin download table and all it's associated functionality added

VKTB:
## Description
Implements the Admin Download table from TopCat. It works differently to the normal Download Status table because it is expected that the database table behind this table will contain many records. Instead of fetching all the records on load, which is what the Download Status table does, it loads 50 records and then makes use of the `loadMoreRows` functionality to load more records (25) when the table is scrolled. It has actions to allow for downloads to be restored, paused, resumed and deleted. Some downloads could take long to restore, therefore the `status` (_Availability_) of the download can change during this process. Instead of the `status` automatically changing in the table, which will require a lot of polling and it's likely to be inefficient, I implemented the refresh functionality found in the Download Status table to allow admins to refresh the table and this will re-fetch and therefore update the status of the download items.
image

The changes so far can be reviewed while I am writing tests.

## Testing instructions
Can be tested with DataGateway-Download on its own or together with SciGateway but will require the changes from the [Display admin functionality from plugins](https://github.com/ral-facilities/scigateway/pull/679) PR. An Admin Download route also needs to be added to the `*-settings.json` file:
```JSON
...
    {
      "section": "Admin",
      "link": "/admin-download",
      "displayName": "Admin Download",
      "admin": true,
      "order": 0
    }
...
```
- [x] Review code
- [x] Check Actions build
- [x] Admin Download table displays when navigating to `/admin-download`
- [x] 50 rows are fetched on load and are displayed in the table
- [x] More rows (25 at a time) are loaded when the table is scrolled down
- [x] Column filtering and sorting work as expected
- [x] Actions work as expected
- [x] Refresh functionality works as expected

## Agile board tracking
closes #613


packages/datagateway-download/src/downloadApi.ts
```diff
@@ -177,6 +177,33 @@ export const fetchDownloads: (
...
+        queryOffset: !queryOffset
+          ? 'where download.isDeleted = false'
+          : queryOffset,
+      },
+    })
+    .then((response) => {
+      return response.data;
+    })
+    .catch((error) => {
+      // Should it be handling ICAT error or TopCat?
```
louise-davies, line 201: The `handleICATError` function does take into account TopCAT errors - it's just a helper function for all bits of the ICAT stack so remove this comment

packages/datagateway-download/src/downloadApi.ts
```diff
@@ -293,6 +320,72 @@ export const downloadDeleted: (
...
+  params.append('value', JSON.stringify(deleted));
+
+  return axios
+    .put(
+      `${settings.downloadApiUrl}/admin/download/${downloadId}/isDeleted`,
+      params
+    )
+    .then(() => {
+      // do nothing
+    })
```
louise-davies, line 350: I think you don't technically need the `then()` here but perhaps it's good to keep in to make it explicit that we aren't doing anything with the response

packages/datagateway-download/src/downloadStatus/adminDownloadStatusTable.component.tsx
```diff
@@ -0,0 +1,600 @@
...
+  const [filters, setFilters] = React.useState<{
+    [column: string]:
+      | { value?: string | number; type: string }
+      | { startDate?: string; endDate?: string };
+  }>({});
+  const [data, setData] = React.useState<FormattedDownload[]>([]);
+  const [dataCount, setDataCount] = React.useState<number>(-1);
+  const [dataLoaded, setDataLoaded] = React.useState(false);
+  const [refreshDownloads, setRefreshDownloads] = React.useState(false);
+  const [lastChecked, setLastChecked] = React.useState('');
```
louise-davies, line 53: I think we should initialise this to be the current time as the last checked UI looks weird without a timestamp on initial load.

packages/datagateway-download/src/downloadStatus/adminDownloadStatusTable.component.tsx
```diff
@@ -0,0 +1,600 @@
...
+                        <Restore />
+                      </IconButton>
+                    );
+                  },
+
+                  function DeleteButton({ rowData }: TableActionProps) {
+                    const downloadItem = rowData as FormattedDownload;
+                    const [isDeleting, setIsDeleting] = React.useState(false);
+
+                    return (
```
louise-davies, line 548: TopCAT hides this button if the download has already been deleted - is this possible? Otherwise we have a confusing button that does nothing

packages/datagateway-download/src/downloadStatus/adminDownloadStatusTable.component.tsx
```diff
@@ -0,0 +1,600 @@
...
+                    setSort(restOfSort);
+                  }
+                }}
+                data={data}
+                loading={!dataLoaded}
+                loadMoreRows={fetchMoreData}
+                totalRowCount={dataCount}
+                actionsWidth={100}
+                actions={[
+                  function RestoreButton({ rowData }: TableActionProps) {
```
louise-davies, line 380: I think in the TopCAT admin table there are two buttons, Restore/Delete and Pause/Resume. Whereas we have Delete and Restore/Pause/Resume buttons. Whether it makes sense to have a Pause/Resume button on a deleted download is a different discussion but that option exists in TopCAT. I think we should align more closely with what already exists, or perhaps consult Chris on what configuration of buttons he needs.

louise-davies (changes requested):
Looks good for the most part and I like how you fixed the problem with the long count query by setting the number to be very large until it comes back as it works very well. Just some small issues and perhaps some further thoughts needed on the specific operations of the buttons that can be a separate issue.

packages/datagateway-download/src/downloadApi.ts
VKTB: I agree. Other functions to the same so I will keep it as it is.

packages/datagateway-download/src/downloadStatus/adminDownloadStatusTable.component.tsx
VKTB: I used the normal Download Status table as an example when I created the buttons hence why _Restore_ and _Delete_ are two separate buttons. It looks like it is possible to hide buttons if `null` is returned by the action function. This is how it behaves with some of the changes that I made just now:
Admin Download table

is this how we want it to work @louise-davies ?

packages/datagateway-download/src/downloadStatus/adminDownloadStatusTable.component.tsx
VKTB: See my reply to your previous comment.

packages/datagateway-download/src/downloadStatus/adminDownloadStatusTable.component.tsx
louise-davies: @VKTB that looks good and looks more like what TopCAT had and I think makes more sense.

packages/datagateway-download/cypress/integration/adminDownloadStatus.spec.ts
```diff
@@ -0,0 +1,157 @@
...
+
+      cy.get('[aria-rowcount="1"]').should('exist');
+      cy.get('[aria-rowindex="1"] [aria-colindex="2"]').should(
+        'have.text',
+        '675eea67-bad7-4209-b54e-87f0d97eb50d'
+      );
+    });
+
+    it.skip('date between', () => {
+      // TODO: Test Requested Date fukter
```
VKTB, line 138: Ignore the spelling mistake but the _Requested Date_ column filter should be tested - I didn't have time. In addition to this, it would also be good to test the action buttons.

PR #662, opened 2021-04-15
Customisable homepage #431

### LINKED ISSUE #431: Customisable homepage?
**Description**:
I like the visual impact of the SciGateway homepage, although I'm a bit concerned that the Explore, Analyse, Record might be setting expectations a bit high at this stage - this is more of a long term goal I think.

Is the page customisable for each facility? As a minimum, for example, can we put the Diamond/ISIS logo on it and some words saying "Welcome to the Diamond/ISIS data archive".

**Acceptance criteria**:
- [ ] A checklist

sam-glendenning:
## Description
This represents the new homepage specific to DataGateway. Overall code management strategy, e.g. where the homepage should live, how it is loaded, etc. is still up for debate. Current assumption is:

1. Store homepage in datagateway-common/src
2. Import homepage into datagateway-dataview (specifically datagateway-dataview/src/page/pageContainer.component.tsx)
3. Register a new homepage route in the datagateway-dataview paths
4. In the PageContainer component, render only the <HomePage /> if location.pathname === paths.homepage

Assumption is to review and suggest changes for this PR until we are satisfied with where a homepage should live, where it should be loaded and how it should be loaded. This will hopefully set a standard for future homepages of new plugins.

## Testing instructions
- [ ] Review code
- [ ] Check Actions build
- [ ] Check homepage is built to match [this spec](https://github.com/ral-facilities/datagateway/issues/431#issuecomment-815986709) as described in the issue

## Agile board tracking
connect to [SciGateway #678](https://github.com/ral-facilities/scigateway/issues/678)
closes #431


packages/datagateway-common/package.json
```diff
@@ -75,7 +75,7 @@
     "prepare": "yarn tsc",
     "test:watch": "react-scripts test --env=jsdom --watch",
     "lint:js": "eslint --ext=tsx --ext=ts --ext=js --ext=jsx --fix ./src",
-    "tsc": "tsc -p tsconfig.lib.json",
+    "tsc": "tsc -p tsconfig.lib.json && cp -r src/images lib/",
```
sam-glendenning, line 78: This was to allow for images on the new homepage to be referred to by source code in lib, although it may not be needed... I'm not sure how dg-common exports images actually. Sometimes plugin images don't load properly on my local machine, so this is something I need to look into.

packages/datagateway-common/src/homePage/contents.tsx
```diff
@@ -0,0 +1,15 @@
...
+    'Browse, explore and visualise the experimental data produced at the facility.',
+  'discover-label': '2. Discover',
+  'discover-description':
+    'Search for the experimental data according to different criteria.',
+  'download-label': '3. Download',
+  'download-description':
+    'Retrieve the experimental data using a variety of download methods.',
+};
+
+export default homePageContents;
```
sam-glendenning, line 15: Might refactor this into a default.json file

packages/datagateway-common/src/homePage/homePage.component.test.tsx
```diff
@@ -0,0 +1,41 @@
...
+
+//   afterEach(() => {
+//     mount.cleanUp();
+//   });
+
+//   it('homepage renders correctly', () => {
+//     const wrapper = shallow(<HomePageWithoutStyles {...props} />);
+//     expect(wrapper).toMatchSnapshot();
+//   });
+// });
```
sam-glendenning, line 41: Commented out at the moment ahead of rewriting for the DG homepage. This test file is borrowed from the SciGateway homepage.

packages/datagateway-common/src/homePage/homePage.component.tsx
```diff
@@ -0,0 +1,146 @@
...
+  Grid,
+  createStyles,
+} from '@material-ui/core';
+import { StyleRules } from '@material-ui/core/styles';
+import Image from '../images/Jasmin4 _DSC7054.jpg';
+import DatagatewayLogoWithText from '../images/datagateway-logo-white.svg';
+import ExploreImage from '../images/explore.jpg';
+import DiscoverImage from '../images/discover.jpg';
+import DownloadImage from '../images/download.jpg';
+// import { UKRITheme } from './theming';
```
sam-glendenning, line 16: Will go back to. This wasn't working properly last I tried.

packages/datagateway-common/src/images/datagateway-logo-white.svg
```diff
@@ -0,0 +1,21 @@
...
+    <path d="M1080.9,143.32l24.69-71.09h10.07l-36.09,98.9-1.87,4.38q-7,15.39-21.49,15.39a26.22,26.22,0,0,1-7.19-1.1l-.07-7.73,4.84.47q6.87,0,11.13-3.4t7.23-11.68l4.14-11.41-31.88-83.82h10.24Z" transform="translate(-0.5 -0.5)" style="fill: rgb(255, 255, 255);"/>
+  </g>
+  <g id="b074a1e9-6de5-4314-8d6e-719a56511290" data-name="D amrk">
+    <path d="M170.21,29.79A100,100,0,0,0,.5,86.36L43.64,43.22A80,80,0,0,1,156.78,156.36L113.64,199.5A100,100,0,0,0,170.21,29.79Z" transform="translate(-0.5 -0.5)" fill="#1e5df8"/>
+    <path d="M160.12,99.79A60,60,0,0,0,57.78,57.37L.5,114.65A100,100,0,0,0,6.39,137L71.92,71.51a40,40,0,0,1,56.57,56.57L63,193.61a100,100,0,0,0,22.39,5.89l57.28-57.28A59.57,59.57,0,0,0,160.12,99.79Z" transform="translate(-0.5 -0.5)" fill="#1e5df8"/>
+    <path d="M114.35,85.65a20,20,0,0,0-28.29,0l-70,70a101,101,0,0,0,28.29,28.29l70-70A20,20,0,0,0,114.35,85.65Z" transform="translate(-0.5 -0.5)" fill="#1e5df8"/>
+  </g>
+</svg>
```
sam-glendenning, line 21: This is a new image that is the same as the standard DataGateway logo but with white text. The one with black text wasn't easily visible against the background image.

packages/datagateway-dataview/public/datagateway-dataview-settings.example.json
```diff
@@ -46,23 +46,29 @@
     }
   ],
   "routes": [
+    {
+      "section": "Homepage",
+      "link": "/datagateway",
+      "displayName": "DataGateway",
+      "order": 0
+    },
```
sam-glendenning, line 54: Homepage needs to be a registered route for at least one plugin for SciGateway to access it. It was decided to use dataview for this since there is currently no implementation of DataGateway that does not use dataview.

packages/datagateway-dataview/src/App.tsx
```diff
@@ -8,6 +8,7 @@ import {
   DGThemeProvider,
   listenToMessages,
   Preloader,
+  //  HomePage,
```
sam-glendenning, line 11: TODO: remove. This was for testing purposes.

packages/datagateway-dataview/src/page/pageContainer.component.tsx
```diff
@@ -477,51 +479,57 @@ class PageContainer extends React.Component<
...
-                  toggleCard={this.state.toggleCard}
-                  handleToggleChange={this.handleToggleChange}
-                />
-              )}
-            />
-          </Grid>
+    // Renders the homepage if the current location equals /datagateway
+    // TODO determine if this is the best way of doing things
+    if (this.props.location.pathname === paths.homepage) {
+      return <HomePage />;
```
sam-glendenning, line 485: May be a better way of rendering the homepage. This is very plugin-specific too. It was done this way because the entirety of the dataview package seems to revolve around the PageContainer component. Right now, it is essentially saying if we are on /datagateway, load the homepage, otherwise load the normal PageContainer page as default.

packages/datagateway-common/src/homePage/theming.tsx
```diff
@@ -0,0 +1,177 @@
...
+  return createMuiTheme(options);
+};
+
+// function mapThemeProviderStateToProps(
+//   state: StateType
+// ): { prefersDarkMode: boolean } {
+//   return {
+//     prefersDarkMode: state.dgcommon.darkMode,
+//   };
+// }
```
sam-glendenning, line 158: This all relates to dark mode theming, which is managed from the SciGateway perspective AFAIK. I've commented it out for now in case it is needed but as of now I don't see a place for it here.

sam-glendenning:
Current state: writing tests for new code and looking for thoughts on code and structure so far

packages/datagateway-common/src/homePage/homePage.component.tsx
```diff
@@ -0,0 +1,147 @@
...
+      color: theme.palette.text.primary,
+      paddingBottom: 20,
+    },
+    howItWorksGridItem: {
+      display: 'flex',
+      flexDirection: 'column',
+      alignItems: 'center',
+    },
+    howItWorksGridItemTitle: {
+      // color: (theme as UKRITheme).ukri.bright.orange,    // TODO add this back. Cannot find 'bright' of undefined
```
sam-glendenning, line 52: Is `UKRITheme` something that needs to be exported from dg-common in `index.tsx` to get this to work? FWIW this code works fine in SciGateway's homepage

packages/datagateway-common/src/homePage/homePage.component.tsx
sam-glendenning: Deciding to remove the theming file entirely and instead hard-code the only line that was being imported from it (the bright orange colour)

packages/datagateway-common/src/homePage/homePage.component.tsx
sam-glendenning: Leaving the theming file out as it wasn't worth importing the whole thing if only one line was taken from it

packages/datagateway-common/src/homePage/homePage.component.test.tsx
```diff
@@ -0,0 +1,35 @@
...
+      },
+    };
+  });
+
+  afterEach(() => {
+    mount.cleanUp();
+  });
+
+  it('homepage renders correctly', () => {
+    const wrapper = shallow(<HomePageWithoutStyles {...props} />);
```
VKTB, line 32: You are using `HomePageWithoutStyles` which is why Codecov warns that `createStyles` is not covered. Instead import the whole component by adding `import HomePage from './homePage.component';` at the top and pass the props to it like you are already doing - `const wrapper = shallow(<HomePage {...props} />);`.

packages/datagateway-common/src/homePage/homePage.component.test.tsx
```diff
@@ -0,0 +1,35 @@
...
+import { HomePageWithoutStyles, HomePageProps } from './homePage.component';
+
+describe('Home page component', () => {
+  let shallow;
+  let mount;
+  let props: HomePageProps;
+
+  beforeEach(() => {
+    shallow = createShallow({ untilSelector: 'div' });
+    mount = createMount();
```
VKTB, line 12: `mount` is not needed as the test uses `shallow`. This also means that the `afterEach` block can also be removed.

packages/datagateway-common/src/homePage/homePage.component.tsx
```diff
@@ -0,0 +1,145 @@
...
+        </Grid>
+      </Grid>
+    </div>
+  </div>
+);
+
+export const HomePageWithoutStyles = HomePage;
+export const HomePageWithStyles = withStyles(styles)(HomePage);
+
+export default HomePageWithStyles;
```
VKTB, line 145: If you make the changes that I suggested in the unit test file then that would mean that `HomePageWithoutStyles` will never be imported. I also don't think that `HomePageWithStyles` should be exported twice. You can refactor these four lines to the following:
```suggestion
export default withStyles(styles)(HomePage);
```

packages/datagateway-common/src/homePage/homePage.component.tsx
```diff
@@ -0,0 +1,145 @@
...
+  WithStyles,
+  Grid,
+  createStyles,
+} from '@material-ui/core';
+import { StyleRules } from '@material-ui/core/styles';
+import Image from '../images/Jasmin4 _DSC7054.jpg';
+import DatagatewayLogoWithText from '../images/datagateway-logo-white.svg';
+import ExploreImage from '../images/explore.jpg';
+import DiscoverImage from '../images/discover.jpg';
+import DownloadImage from '../images/download.jpg';
```
VKTB, line 15: As shown in the screenshot, these images apart from `explore.jpg` are not found. I can see them in the `images` directory though.
image

packages/datagateway-dataview/public/datagateway-dataview-settings.example.json
VKTB: The navigation drawer now contains a link that directs to the homepage when clicked on it. I think that it would be better not to have this. I already did something similar for the admin plugins as part of the [Display admin functionality from plugins](https://github.com/ral-facilities/scigateway/pull/679) PR in the `navigationDrawer.component.tsx` file,

packages/datagateway-dataview/src/page/pageContainer.component.test.tsx
```diff
@@ -445,4 +446,29 @@ describe('PageContainer - Tests', () => {
       wrapper.find(PageContainer).children().first().state('modifiedLocation')
     ).toEqual(newLocation);
   });
+
+  it('should return the homepage if current path equals homepage route', () => {
+    const homepageLocation = createLocation(paths.homepage);
```
VKTB, line 451: I like the idea of how you made use of `paths` here as values can be kept in one place and changes will not need to be made in multiple place if the value of `homepage` needs to be changed in future. I wonder if it's worth doing the same thing for the other tests of this file too?

VKTB (changes requested):

packages/datagateway-common/src/homePage/homePage.component.tsx
sam-glendenning: Yeah I wasn't sure if it was just me. Okay thanks for that feedback, clearly something is going on there

packages/datagateway-dataview/src/page/pageContainer.component.test.tsx
sam-glendenning: That's what I thought. Worth a refactor, I'd say

packages/datagateway-dataview/public/datagateway-dataview-settings.example.json
sam-glendenning: This is something I'll fix on the SciGateway side of things

packages/datagateway-dataview/public/datagateway-dataview-settings.example.json
```diff
@@ -46,24 +46,40 @@
...
+  "home-page": {
+    "title": "DataGateway",
+    "how-label": "How it works...",
+    "explore-label": "1. Explore",
+    "explore-description": "Browse, explore and visualise the experimental data produced at the facility.",
+    "discover-label": "2. Discover",
+    "discover-description": "Search for the experimental data according to different criteria.",
+    "download-label": "3. Download",
+    "download-description": "Retrieve the experimental data using a variety of download methods."
+  },
```
sam-glendenning, line 83: Homepage text now lives inside datagateway-dataview-settings.json

packages/datagateway-common/src/images/datagateway-logo-white.svg
sam-glendenning: This image is not actually needed because a logo with white text already existed! Will remove in a later commit.

sam-glendenning:
Changes have been made to datagateway-dataview-settings-example.json. Make sure to update your datagateway-dataview-settings.json when testing this.

packages/datagateway-common/src/homePage/homePage.component.tsx
sam-glendenning: This has, finally, been fixed!

packages/datagateway-common/package.json
sam-glendenning: This additional cp command has now been removed as it is no longer necessary

packages/datagateway-common/src/homePage/contents.tsx
sam-glendenning: Homepage text now lives in datagateway-dataview-settings.json

packages/datagateway-dataview/src/page/pageContainer.component.test.tsx
sam-glendenning: This has also been done

packages/datagateway-dataview/public/datagateway-dataview-settings.example.json
sam-glendenning: This has now been fixed in[ this commit](https://github.com/ral-facilities/scigateway/pull/704/commits/309ad488cfc0da6189170d013b3e8f85d70eaa45).

louise-davies:
@sam-glendenning There's a setting in datagateway-dataview called `pluginHost` which effectively provides what you have made a hacky solution for which is used by the register route code to provide paths to the logos. I think unifying and using that solution is both less hacky and more consistent

sam-glendenning:
@louise-davies thanks for the hint with "pluginHost", I did not realise that was what that setting is for. The latest commit incorporates its usage. 

Currently, "pluginHost" is not a mandatory setting but I decided to make it mandatory if a "homepage" field is set. This is because the homepage wouldn't display properly without it as the image srcs would not be able to be found. This is why [this line](https://github.com/ral-facilities/datagateway/blob/e8bd18f21c38c8ec6bd007ab3f715c0b27fc5cba/packages/datagateway-dataview/src/state/actions/index.tsx#L182) is as it is.

sam-glendenning:
Last commit addresses consistency and state issues. PR is ready for review again.

packages/datagateway-dataview/src/page/pageContainer.component.tsx
```diff
@@ -475,51 +485,64 @@ class PageContainer extends React.Component<
...
-          navigateToDownload={this.props.navigateToDownload}
+    if (this.props.location.pathname === paths.homepage) {
+      return (
+        <HomePage
+          {...this.props}
+          logo={this.props.pluginHostUrl + DGLogo}
+          backgroundImage={this.props.pluginHostUrl + BackgroundImage}
+          exploreImage={this.props.pluginHostUrl + ExploreImage}
+          discoverImage={this.props.pluginHostUrl + DiscoverImage}
+          downloadImage={this.props.pluginHostUrl + DownloadImage}
```
louise-davies, line 496: Can we do this using a `Switch` and a `Route` rather than manually via the redux store? I know we connect our router to the store but I think we should still do simple routing like this the `react-router` way. You'd then probably need to refactor the `else` into a standalone component to provide to a `Route` with no `path` param (basically catches all other routes). We need to refactor into a standalone component and not an inline render component because otherwise we'd have a lot of unnecessary rerenders.

louise-davies (changes requested):

packages/datagateway-common/src/homePage/homePage.component.tsx
```diff
@@ -58,122 +57,126 @@ const styles = (theme: Theme): StyleRules =>
...
 
-export type CombinedHomePageProps = HomePageProps & WithStyles<typeof styles>;
+type CombinedHomePageProps = HomePageProps & WithStyles<typeof styles>;
 
-class HomePage extends React.Component<CombinedHomePageProps> {
-  getString = (key: string): string => {
-    return this.props.res && this.props.res[key] ? this.props.res[key] : key;
+const HomePage = (props: CombinedHomePageProps): React.ReactElement => {
+  const getString = (key: string): string => {
+    return props[key] !== undefined ? props[key] : key;
```
louise-davies, line 82: We don't actually need this function any more as the `t` function handles this logic so it will always pass a `string`. Plus all the types of the props are `string` not `string | undefined` so I think it would just be easier to refer to them directly.

packages/datagateway-dataview/src/page/pageRouting.component.tsx
```diff
@@ -219,6 +227,42 @@ const SafeDLSDatasetsCardView = React.memo(
 );
 SafeDLSDatasetsCardView.displayName = 'SafeDLSDatasetsCardView';
 
+interface HomePageStateProps {
+  pluginHostUrl: string | undefined;
+}
+
+const SafeHomePage = React.memo(
```
louise-davies, line 234: This probably makes sense more to be named "Translated" HomePage rather than "Safe".

packages/datagateway-common/src/homePage/homePage.component.tsx
sam-glendenning: Ahh good spot, I had missed that

packages/datagateway-dataview/src/page/pageContainer.component.tsx
```diff
@@ -477,47 +478,56 @@ class PageContainer extends React.Component<
...
+            location={this.props.location}
+          />
+        ) : (
+          <>
+            <NavBar
+              entityCount={this.props.entityCount}
+              cartItems={this.props.cartItems}
+              navigateToSearch={this.props.navigateToSearch}
+              navigateToDownload={this.props.navigateToDownload}
+            />
```
sam-glendenning, line 493: Done like this to ensure the nav bar and a big footer don't render on the homepage (along with other unnecessary pageContainer components). Otherwise, we get something like in the image. Still not overly sure of this code structure. Thoughts, @louise-davies ?
Capture

louise-davies: It might make more sense to have the `HomePage` component back in the `PageContainer` component similar to the original logic and instead just add a `Switch` component rather than an `if` statement and two routes - one for the `HomePage` which has no `NavBar` and one catchall route for the layout stuff for the `NavBar` and thus the `PageRouting` component. Again, I would just say we would need to memoize the layout stuff + NavBar to optimise. Let me know if you'd like a pair programming session to explain this a little better @sam-glendenning

packages/datagateway-dataview/src/page/pageContainer.component.tsx
```diff
@@ -476,49 +478,58 @@ class PageContainer extends React.Component<
...
-          entityCount={this.props.entityCount}
-          cartItems={this.props.cartItems}
-          navigateToSearch={this.props.navigateToSearch}
-          navigateToDownload={this.props.navigateToDownload}
-        />
+      <SwitchRouting location={this.props.location}>
+        {/* Load the homepage */}
+        <Route exact path={paths.homepage} component={TranslatedHomePage} />
+        <Route
+          render={() => (
```
sam-glendenning, line 485: Thanks to use of the `render` function, the original functionality of pageContainer remains without being remounted all the time. This also removes the need for an entirely new component dedicated just to pageContainer, although if it helps readability and maintenance down the line, this can also be done.

sam-glendenning:
Apologies for dismissing the previous review, Github does not make it easy to re-request a review sometimes. Ready for review again, @louise-davies

louise-davies (approved):
I'm happy with the code changes and it works well

louise-davies (approved):

sam-glendenning:
@louise-davies was fixing conflicts that arose from merging [this PR](https://github.com/ral-facilities/datagateway/pull/547) into master. Most revolved around the `"pluginHost"` setting that both Patrick and I were testing. Overwrote my changes with his as they basically do the same thing.

louise-davies (approved):

Issue #659, opened 2021-04-14
Fix My Data e2e tests


louise-davies:
**Description**:
Currently, the myData tests fetch data for an anonymous user - which by definition doesn't make sense. This is currently fine on master due to a workaround, but this workaround is being removed as part of the switch to python-icat and so shouldn't be relied upon. The tests should login as a user which has data they are the user for - either we can do this by adding credential login back to that test, modifying the fake data + fake data generator script or via hacking the local storage token and manually changing the username.

**Acceptance criteria**:
- [ ] My Data table e2e tests run as a user that has data associated with it

sam-glendenning:
I just had a go at this for feasibility's sake. Modifying the generator script or stubbing response data from the API could be tricky on a test-by-test basis, especially when it comes to testing filters and sorting. Since those work by querying the API, each response would have to be stubbed just to make the test pass and that seems to defeat the spirit of a true e2e test. I think we would need to have some sort of permanent user with data linked to it that we login as while running these tests. But open to ideas

MRichards99:
I know Louise had some ideas about this but I don't know enough about them to give you guidance on what to do :(

Issue #658, opened 2021-04-14
Fix investigationSearchTable slow queries due to how facility cycles are fetched


louise-davies:
**Description**:
Investigating failing e2e tests with Matt on his JSON casing branch, we discovered that the inclusion of `facilityCycle` in the query is massively slowing down the request. Including `facilityCycles` means a lot of duplicate data since it fetches the whole list of facility cycles for each investigation. Instead, we should send a separate request to the API for the list of facility cycles (if we detect we're an ISIS facility), store this in the table as local state and use that for the link construction instead.

**Acceptance criteria**:
- [ ] Facility Cycles are fetched once per investigationSearchTable and not for each item in a table

Issue #94, opened 2019-11-28
Casing of JSON fields in datagateway-api is not consistent with ICAT API 


louise-davies:
Related to ral-facilities/datagateway-api#119

Since datagateway-api needs to change what it returns to match ICAT, so does datagateway-table.

Acceptance criteria:
- [ ] Once ral-facilities/datagateway-api#119 has been fixed, datagateway-table makes the same changes w.r.t. casing to access the API properly.

PR #663, opened 2021-04-16
Make DataGateway Work with API's ICAT Backend

MRichards99:
## Description
In a sentence, these changes allow DataGateway to use DataGateway API's Python ICAT backend, the backend to be used in production. The majority of these changes revolve around moving attribute names from SNAKE_CASE to camelCase, so many of these commits (especially a2c7fde001a4f10456aead012951db029737a495) are the result of find and replace. However, there foreign key attributes have been removed (e.g. `INVESTIGATION_ID` on `Dataset`) as the ICAT backend doesn't make use of foreign keys. 

There's one _very_ minor change in functionality - full names of users are now displayed as per b7f0642a3fa0492ca9f2948690e856351ed824c0.

I've left a couple of TODOs in (search for `TODO -` which will display mine) just to serve as reminders about a couple things. They should probably get removed before getting merged, but I wanted to keep them there until _everything_ passes etc.

On that note, there are still a few outstanding things on this branch:
- My data table views (both ISIS and Diamond) have issues on their e2e tests. There is an issue for this (#659)
- Lots of e2e tests on Search fail due to a slow query involving facilityCycles that are included in said query. Again, there is an issue for this (#658) (EDIT: this is fixed as of 26/05/21)
- On dataview, e2e tests on Diamond proposals (card & table view) fail - this is because ICAT Server 4.11.1 needs to be installed on preprod (an request to the API with a distinct filter on multiple fields which needs the newer version of ICAT Server)
- `cartSelection.spec.ts` fails due to issues with the my data views

The CI will inevitably fail on the e2e tests because the DB backend is on preprod, rather than ICAT backend. I've made some screenshots to show the current state of the tests:

Unit Tests for all components- all pass nicely:
all-unit-tests

DataView Card View e2e - all pass except for Diamond proposals (4.11.1) as above
dataview-e2e-batch1

DataView Table View e2e - all pass except for Diamond proposals (4.11.1) and mydata (issues above)
dataview-e2e-batch2

DataView Remaining e2e - all pass except for cartSelection (mydata)
dataview-e2e-batch3

Download e2e - all pass
download-e2e

Search e2e - lots of fails due to facilityCycles include (issue above)
search-e2e

## Testing instructions
To test this, the ICAT backend of the API should be used. Specifically, use `temp-distinct-client-handling-merge` branch on DataGateway API - this has a mixture of the new client handling solution and the fix for the distinct filter and is the branch that's most like the API that'll be run in production at time of writing. For client handling configuration, I recommend the following settings for testing. You should be able to get away with `client_pool_init_size: 3` but the below settings are the ones that have actually been tested with this frontend branch:
```json
"client_cache_size": 10,
"client_pool_init_size": 7,
"client_pool_max_size": 10,
```

As mentioned above, ICAT Server 4.11.1 is required for the API from now on.

- [ ] Review code
- [ ] Check Actions build
- [ ] {more steps here}

## Agile board tracking
connect to #94


packages/datagateway-download/src/downloadApi.ts
```diff
@@ -361,10 +361,11 @@ export const getDatafileCount: (
       .get<number>(`${settings.apiUrl}/datafiles/count`, {
         params: {
           where: {
-            DATASET_ID: {
+            'dataset.id': {
               eq: entityId,
             },
           },
+          include: 'dataset',
```
louise-davies, line 368: Adding this `include: dataset` causing this function to fail on Diamond prod - it's giving an error to do with ICAT not supporting multiple attributes? Regardless, the `where` on it's own works fine for the `python-icat` backend (although I presume not for the database backend)

Can we investigate this using DLS preprod to double check?

MRichards99: Does Diamond prod have ICAT Server 4.11.1? That might be why there's an issue

louise-davies: It doesn't no - in fact I forgot that currently lots of requests currently fail due to this! So it's not a DataGateway issue then - perhaps we might need to use an older version of datagateway-api for DLS if the ICAT 4.11.1 release is delayed

MRichards99: I could certainly find the most recent API commit not to require 4.11.1 if needed, but that'll mean the distinct filter will be broken on that version of the API, something which Diamond need I think?

louise-davies: This is more thinking of a worst case scenario in which we develop a work around in DataGateway to the issue (similar to what TopCAT does) then in that case we'd also need to drop the datagateway-api to a previous commit yeah. Nothing to worry about now though hopefully!

MRichards99: Yeah if we need to do something like that, just let me know and I'm sure we can get it sorted :)

MRichards99:
I removed the failing test for the landing page work (and put studyInvestigation's investigation to mandatory/removed the `?`). I've also added the fixes for the datafiles stuff, and made the suggested refactor - a3fe16a7a30c826e67481fb2c97997ba3fc476c5 for this :)

MRichards99:
CI shows the unit tests all pass, but I've run e2e tests locally to check them:

No changes for table view - my data still fail (existing issue for this), diamond proposals passed when I ran it in interactive mode, visits and the two failing investigations files were fixed in 9a8bc7dfcc055646375f2e5bab3a62145597a250.


Card view are all good
image

Same with landing pages
image

Other dataview tests pass too (only cart selection ones to fail are the ISIS my data table)
image

Download passes too
image

Search still fails but hopefully it'll be fixed with the existing issue

sam-glendenning:
> Search still fails but hopefully it'll be fixed with the existing issue

In the middle of trying to get this fix done, @MRichards99. Currently having problems with merge conflicts between my branch and yours, causing a bit of a headache

sam-glendenning:
> > Search still fails but hopefully it'll be fixed with the existing issue
> 
> In the middle of trying to apply this fix, @MRichards99. Currently having problems with merge conflicts between my branch and yours, causing a bit of a headache

MRichards99:
@sam-glendenning yeah I noticed the PR, great to hear it's close to being done! We can go through the merge conflicts together at some point if that helps :)

sam-glendenning:
> @sam-glendenning yeah I noticed the PR, great to hear it's close to being done! We can go through the merge conflicts together at some point if that helps :)

I'll keep trying to figure it out, if nothing else it's not a lot of code and I could just commit it directly to this branch without bothering with a PR for merging into this branch

louise-davies (approved):

PR #665, opened 2021-04-20
Fix query param updates causing freezing

louise-davies:
## Description
Fixes bug which caused freezing when card views were navigated between via the SciGateway nav drawer. This required a lot of investigation to what was going on and I presumed it was our code causing the problem. But I got stuck and started searching the related libraries (react-redux, react-router, connected-react-router etc.) for issues and as you can see I found the issue linked in a comment in the code. The reason the bug only occurs in deployment is because that's the only time the components completely unmount (when navigating via the navbar, SciGateway unmounts the plugins), and that was the condition for the bug.

There's no unit tests since this is effectively hotfixing an issue with an underlying library, and I don't think writing any tests for this is worth much time. Similarly, due to the bug only occurring when mounted in SciGateway, no e2e tests have been written

## Testing instructions
I deployed the new version on scigateway-preprod and you can look at that to see the bug no longer occurs. I can also redeploy master for comparison.

- [x] Review code
- [x] Check Actions build

### DIFF
--- a/packages/datagateway-dataview/package.json
+++ b/packages/datagateway-dataview/package.json
@@ -15,6 +15,7 @@
     "connected-react-router": "^6.9.1",
     "custom-event-polyfill": "^1.0.7",
     "datagateway-common": "0.1.0",
+    "history": "^4.10.1",
     "i18next": "^19.9.0",
     "i18next-browser-languagedetector": "^6.0.0",
     "i18next-http-backend": "^1.1.1",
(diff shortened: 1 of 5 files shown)


sam-glendenning (commented):
Looks fine as far as I can tell, as I've not been able to recreate the bug I can't test that it's now gone but I don't come across it on preprod. Only thing left is that codecov isn't happy with the diff, although I'm not sure how to test the lines it's complaining about. Can we merge without passing that test?

VKTB (commented):
Looks good but like Sam stated, Codecov isn't happy.

louise-davies:
@VKTB @sam-glendenning I'm an admin so I can merge with the failing codecov patch. There's no real way to make codecov happy for that code without modifying the code and writing kind of meaningless tests. Ideally the issue would be fixed upstream.

VKTB (approved):

sam-glendenning (approved):

Issue #696, opened 2021-05-20
Fix display of investigation.visitId label


agbeltran:
**Description**:
See screenshot

**Acceptance criteria**:
- [ ] A correct label is shown for investigation.visitId

<img width="1364" alt="Screenshot 2021-05-20 at 15 07 21" src="https://user-images.githubusercontent.com/1417033/119004666-a92d3300-b986-11eb-8856-889607430520.png">

PR #700, opened 2021-05-24
Fixing slow search queries on `bugfix/casing-json-fields-#94` branch #658

### LINKED ISSUE #658: Fix investigationSearchTable slow queries due to how facility cycles are fetched

sam-glendenning:
## Description
This issues a fix for slow queries in the investigationSearchTable that are being experienced on the `bugfix/casing-json-fields-#94` branch. As the linked issue describes, what was making the queries slow was that the entire list of facility cycles was being fetched with each query to then construct the link to each investigation. This resulted in a large amount of data being sent that wasn't necessary.

This fix fetches the entire list of facility cycles on component mount, stores that list in local state and uses that to construct the investigation links instead (ISIS investigations only). Facility cycles are no longer fetched in the investigation queries, which still take place.

## Testing instructions
- [ ] Review code
- [ ] Check Actions build

## Agile board tracking
Connect to Issue #94 and [PR #663](https://github.com/ral-facilities/datagateway/pull/663)
Closes #658


sam-glendenning:
As of 24/05/21, this is up to date with the branch it is based off of, [bugfix/casing-json-fields-#94](https://github.com/ral-facilities/datagateway/tree/bugfix/casing-json-fields-%2394). As issue #658 relates to this branch, I have made it the base to be merged into. Alternatively, as there's not much code to it, I can simply commit it directly to the base branch instead, although I thought this PR would be good for review purposes.

MRichards99:
Fixed the Search e2e tests so they now pass, this change now means some of the timeouts can be reduced too :)

image

MRichards99 (approved):
This fixes the issue mentioned in the original issue and the Search e2e tests now pass. I think @louise-davies mentioned some visit ID changes got reverted and it might be good for one of you to sanity check my e2e test changes, but I'm happy for this to get merged once those things happen :)

sam-glendenning:
Will wait on further info from Louise about changes I may have overwritten before proceeding.

packages/datagateway-search/src/table/investigationSearchTable.component.tsx
```diff
@@ -229,7 +254,7 @@ const InvestigationSearchTable = (
           filterComponent: textFilter,
         },
         {
-          label: t('investigations.visit_id'),
+          label: t('investigations.visitId'),
```
louise-davies, line 257: Can this change be reverted please - this undoes the fix to #696

PR #713, opened 2021-06-02
Configuring links for all search results #566

### LINKED ISSUE #566: Configure links in search plugin to match facility

sam-glendenning:
## Description
This branch is based on and builds on [Make DataGateway work with API's ICAT backend](https://github.com/ral-facilities/datagateway/pull/663) and [Fixing slow search queries on bugfix/casing-json-fields-#94](https://github.com/ral-facilities/datagateway/pull/700) PRs. 

This finishes off the issue that was originally addressed by Patrick but not completed. Investigation search results in DataGateway currently provide a link to the investigation itself but datasets and datafiles do not. This fix ensures that they now do, regardless of the facility being used (ISIS, DLS or a generic link).

## Testing instructions
Use the ISIS dev URLs for testing as the code makes use of alternative casing when referring to fields. Perform some test search queries in DataGateway search (e.g. "test", "dog", etc.) and experiment with filtering by investigation, dataset and datafile. _All_ results should have a link in their first column (corresponding to their name or title) that should take the user to the appropriate place. For example, if the link is for a DLS investigation, the user should be navigated to `/browse/proposal/...` and if the link is for an ISIS investigation, the user should be navigated to `/browse/investigation/...`. Datafiles also have a column for parent dataset (with a link) and datasets have a column for parent investigation (also with a link).

- [x] Review code
- [x] Check Actions build
- [ ] Check investigations link to investigation pages, datasets to dataset pages and datafiles to datafile pages
- [ ] Check that DLS results link to DLS pages and ISIS results link to ISIS pages
- [ ] Check the ID numbers in the constructed links tally with data received from ICAT

## Agile board tracking
Closes #566

### DIFF
--- a/packages/datagateway-search/src/table/datafileSearchTable.component.tsx
+++ b/packages/datagateway-search/src/table/datafileSearchTable.component.tsx
@@ -130,14 +130,19 @@ const DatafileSearchTable = (
         return datafileData.name;
       }
 
-      if (datafileData.dataset.startDate && facilityCycles.length) {
+      if (
+        facilityCycles.length &&
+        datafileData.dataset?.investigation?.startDate
+      ) {
         const filteredFacilityCycles: FacilityCycle[] = facilityCycles.filter(
           (facilityCycle: FacilityCycle) =>
-            datafileData.dataset?.startDate &&
+            datafileData.dataset?.investigation?.startDate &&
             facilityCycle.startDate &&
             facilityCycle.endDate &&
-            datafileData.dataset?.startDate >= facilityCycle.startDate &&
-            datafileData.dataset?.startDate <= facilityCycle.endDate
+            datafileData.dataset.investigation.startDate >=
+              facilityCycle.startDate &&
+            datafileData.dataset.investigation.startDate <=
+              facilityCycle.endDate
         );
         if (filteredFacilityCycles.length) {
           facilityCycleId = filteredFacilityCycles[0].id;
@@ -148,7 +153,7 @@ const DatafileSearchTable = (
         return linkType === 'dataset'
           ? tableLink(
               `/browse/instrument/${instrumentId}/facilityCycle/${facilityCycleId}/investigation/${datafileData.dataset.investigation.id}/dataset/${datafileData.dataset.id}`,
-              datafileData.dataset?.name
+              datafileData.dataset.name
             )
           : tableLink(
               `/browse/instrument/${instrumentId}/facilityCycle/${facilityCycleId}/investigation/${datafileData.dataset.investigation.id}/dataset/${datafileData.dataset.id}/datafile`,
@@ -175,7 +180,6 @@ const DatafileSearchTable = (
             datafileData.name
           );
     }
-
     if (linkType === 'dataset')
       return datafileData.dataset ? datafileData.dataset.name : '';
     return datafileData.name;
@@ -342,7 +346,11 @@ const mapDispatchToProps = (
           },
           {
             filterType: 'include',
-            filterValue: '"dataset"',
+            filterValue: JSON.stringify({
+              dataset: {
+                investigation: { investigationInstruments: 'instrument' },
+              },
+            }),
           },
         ],
       })
--- a/packages/datagateway-search/src/table/datafileSearchTable.component.tsx
+++ b/packages/datagateway-search/src/table/datafileSearchTable.component.tsx
@@ -343,7 +343,7 @@ const mapDispatchToProps = (
           },
           {
             filterType: 'include',
-            filterValue: 'dataset',
+            filterValue: '"dataset"',
           },
         ],
       })
--- a/packages/datagateway-search/src/table/datafileSearchTable.component.tsx
+++ b/packages/datagateway-search/src/table/datafileSearchTable.component.tsx
@@ -19,13 +19,18 @@ import {
   sortTable,
   filterTable,
   clearTable,
+  tableLink,
+  handleICATError,
+  readSciGatewayToken,
+  FacilityCycle,
 } from 'datagateway-common';
-import { IndexRange } from 'react-virtualized';
+import { TableCellProps, IndexRange } from 'react-virtualized';
 import { connect } from 'react-redux';
 import { StateType } from '../state/app.types';
 import { ThunkDispatch } from 'redux-thunk';
 import { Action, AnyAction } from 'redux';
 import { useTranslation } from 'react-i18next';
+import axios from 'axios';
 
 interface DatafileSearchTableStoreProps {
   sort: {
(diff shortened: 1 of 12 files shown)


sam-glendenning:
@louise-davies I think this is now ready for a more formal review, what makes this PR more complex is that it's based off of [this PR](https://github.com/ral-facilities/datagateway/pull/663) (so testing should be done using ISIS dev URLs) which has still not been merged but the base branch is currently master. So I'm not sure where this puts things for merging code together.

I believe the reason the e2e tests are failing is the same reason they are failing on Matt's branch but double check that to confirm.

sam-glendenning:
Small update on this: all tests are now passing except for one download e2e test. This test is currently failing across multiple branches for an unknown reason. Because e2e tests are currently being overhauled in `bugfix/fix-e2e-tests-#712` and because this one test is not particularly important, a small patch to temporarily `skip` it is in the process of being merged into master as part of [this commit](https://github.com/ral-facilities/datagateway/pull/716/commits/96c0533eaa7e44a037def865d10453e79ff3f334) in [this PR](https://github.com/ral-facilities/datagateway/pull/716). 

Once this PR has been merged, I will merge master into this branch and mark the PR as ready for review once again.

packages/datagateway-search/src/table/datasetSearchTable.component.tsx
```diff
@@ -123,6 +206,38 @@ const DatasetSearchTable = (
     />
   );
 
+  const fetchFacilityCycles = React.useCallback(() => {
```
louise-davies, line 209: This function is repeated in all three tables - but since this will eventually be refactored with the `react-query` changes it's fine for now

louise-davies:
It doesn't seem as though the links are rendering for me - I don't know what's going on with it?

image

and oddly a similar thing happens in the investigation view but not for each row:

image

Could you take a look a this please?

sam-glendenning:
Will do.

sam-glendenning:
@louise-davies I've figured out the datafile table problem. The links are not being generated because the dataset can't link to its parent investigation. This is because one is not returned from the datagateway-api/datafiles endpoint, meaning we cannot extract the investigation ID required as part of the URL. The investigation field itself is optional in the ICAT schema for Dataset objects so this means this will fail in its current form. Should I talk to Matt about this? What's your advice?

As for the investigations bug, I'm unable to replicate this with scigateway-preprod data. Could you confirm what icatUrl, apiUrl, etc. you were using for this test?

louise-davies:
@sam-glendenning the field should always be populated as long as we include it since Datasets always belong to an Investigation (it can't be null in the schema - see https://repo.icatproject.org/site/icat/server/4.11.1/schema.html#Dataset (the column saying "Card" means the minimum and maximum amount of each item. So `1,1` means it's required and there's only ever one) ).

I took a further look, and the issue with investigations is the same issue with TopCAT in that those investigations are missing facility cycles (e.g. if you try this search in ISIS TopCAT dev it also doesn't work right in that the links are rendered but if you click them there's an error).

I found some issues with the code, so I'll comment back in a review

packages/datagateway-search/src/table/datasetSearchTable.component.tsx
```diff
@@ -193,6 +321,10 @@ const mapDispatchToProps = (
               id: { in: luceneData },
             }),
           },
+          {
+            filterType: 'include',
+            filterValue: '"investigation"',
```
louise-davies, line 326: We also need to include investigationInstrument and Instrument otherwise there's not enough info to form the links. This should instead be `filterValue: JSON.stringify({investigation: { investigationInstruments: 'instrument' }})`

packages/datagateway-search/src/table/datasetSearchTable.component.tsx
```diff
@@ -76,10 +81,101 @@ const DatasetSearchTable = (
...
+      }
+
+      if (datasetData.startDate && facilityCycles.length) {
+        const filteredFacilityCycles: FacilityCycle[] = facilityCycles.filter(
+          (facilityCycle: FacilityCycle) =>
+            datasetData.startDate &&
+            facilityCycle.startDate &&
+            facilityCycle.endDate &&
+            datasetData.startDate >= facilityCycle.startDate &&
+            datasetData.startDate <= facilityCycle.endDate
```
louise-davies, line 128: We need to check the investigation startdate, not the dataset startdate

packages/datagateway-search/src/table/datafileSearchTable.component.tsx
```diff
@@ -80,10 +86,111 @@ const DatafileSearchTable = (
...
+
+      if (datafileData.dataset.startDate && facilityCycles.length) {
+        const filteredFacilityCycles: FacilityCycle[] = facilityCycles.filter(
+          (facilityCycle: FacilityCycle) =>
+            datafileData.dataset?.startDate &&
+            facilityCycle.startDate &&
+            facilityCycle.endDate &&
+            datafileData.dataset?.startDate >= facilityCycle.startDate &&
+            datafileData.dataset?.startDate <= facilityCycle.endDate
+        );
```
louise-davies, line 141: Again, need to use investigation startDate here

packages/datagateway-search/src/table/datafileSearchTable.component.tsx
```diff
@@ -201,6 +340,10 @@ const mapDispatchToProps = (
               id: { in: luceneData },
             }),
           },
+          {
+            filterType: 'include',
+            filterValue: '"dataset"',
```
louise-davies, line 345: Again, we need a more detailed include i.e.: `filterValue: JSON.stringify({dataset: {investigation: { investigationInstruments: 'instrument' }}})`

louise-davies (changes requested):
After investigating, these were what were causing my issues

packages/datagateway-search/src/table/datafileSearchTable.component.tsx
sam-glendenning: Very handy for improving my understanding of how these queries are structured!

sam-glendenning:
Having a small issue whereby hovering the mouse over a link which has been shortened because its length exceeds the current column width, thus revealing a popup displaying the full string, causes the page to crash. This only affects the dataset and datafile table. Looking into this, which is why this has been marked as draft in the meantime

louise-davies (approved):

PR #716, opened 2021-06-07
Fixing dark mode and width glitch in 'show more' text effect #691

### LINKED ISSUE #691: Fading in 'show more'
**Description**:
The fading in 'show more' is not spanning the width of the card. In addition, in dark mode is showing a white background. 

**Acceptance criteria**:
- [x] Fix width of the 'show more' fading in cards
- [x] Fix colour of 'show more' fading in dark mode

sam-glendenning:
## Description
This fixes a couple of small bugs with the 'Show more' button that appears in investigation descriptions when the string value is longer than the height of the card it is displayed in. This appears in card mode and gives a visual effect of the text fading to transparent above the 'Show more' button. There were two problems that this PR fixes:

1. In dark mode, the effect broke and a white bar was displayed instead of a transparent one matching the color scheme
2. The bar providing this effect was not long enough to cover the entire description

To see the bugs in action, visit https://datagateway-dev.isis.stfc.ac.uk/browse/instrument?view=card and toggle dark mode on. You can compare this fix to that page.

## Testing instructions
Load the Datagateway Dataview with SciGateway to properly test this. Navigate to a page containing an investigation card view such as `/browse/instrument?view=card`. I wasn't able to find an investigation description in the test data that was long enough to display the effect so either trust the below screenshot where I amended the descriptions manually:

Capture

or change the description value yourself. The way I do this is by navigating to `packages/datagateway-common/src/card/entityCard.component.tsx` and amend lines 279-281 to something like:

```
{description
? 'A SUITABLY LONG DESCRIPTION STRING HERE'
: t('entity_card.no_description')}
```

- [x] Review code
- [x] Check Actions build

## Agile board tracking
Closes #691


packages/datagateway-common/src/card/entityCard.component.tsx
```diff
@@ -76,9 +76,12 @@ const useCardStyles = makeStyles((theme: Theme) => {
     shadowVisible: {
       position: 'absolute',
       height: 30,
-      minWidth: '30vw',
+      minWidth: '66.5%',
```
sam-glendenning, line 79: This specific value is enough to cover all of the text but not cover the divider to the right of the text. However, this only fully works if the window is full size. Outside of this, part of the divider is covered but the effect isn't too noticeable. Open to suggestions on how to get this perfect.

sam-glendenning: Edit: now set this to `minWidth: '100%'`. Still open to suggestions though

packages/datagateway-common/src/card/entityCard.component.tsx
```diff
@@ -76,9 +79,11 @@ const useCardStyles = makeStyles((theme: Theme) => {
     shadowVisible: {
       position: 'absolute',
       height: 30,
-      minWidth: '30vw',
+      minWidth: '100%',
       top: 130,
-      background: 'linear-gradient(rgba(255, 255, 255, 0), #fff)',
+
+      //RGBA values seem to work best with linear gradients being dynamically changed by dark mode preference
+      background: 'linear-gradient(' + paperZero + ', ' + paperOne + ')',
```
louise-davies, line 86: This is probably easier to read if it used string templating i.e. 
```
`linear-gradient ('${paperZero}, ${paperOne}')`
```

packages/datagateway-common/src/card/entityCard.component.tsx
```diff
@@ -76,9 +79,11 @@ const useCardStyles = makeStyles((theme: Theme) => {
     shadowVisible: {
       position: 'absolute',
       height: 30,
-      minWidth: '30vw',
+      minWidth: '100%',
       top: 130,
-      background: 'linear-gradient(rgba(255, 255, 255, 0), #fff)',
+
+      //RGBA values seem to work best with linear gradients being dynamically changed by dark mode preference
```
louise-davies, line 85: Is there a reason we need to use RGBA values rather than HEX codes? Would seem simpler if we could but this comment implies that RGBA is better

sam-glendenning: I tried the HEX equivalent by adding alpha codes on the end of the values but it just didn't seem to work. I'll try one more time when I address your other comment with another commit

louise-davies: I think we should be able to use alpha values with hex codes directly: see https://caniuse.com/css-rrggbbaa

Whether we don't for browser incompatibility is another issue (looking at you IE 11...)

sam-glendenning: Yes I just got it working (was another issue causing the alpha values to fail). Fix inbound!

sam-glendenning:
@louise-davies I've reverted back to using that hexToRgba library. This was because the hex values received by the theme object can sometimes be 3 characters long and sometimes 6, which requires a bunch of extra code (and testing of this code) to convert the values to a consistent format so they can have their alpha properties modified, which was why the previous commit only came to 66% code coverage. I figured it would all be easier if I just used this library. If you'd rather I applied the hex code instead, I'll do it.

GoelBiju:
@sam-glendenning I have just merged in master so that the latest fixes from master would allow for the data to load on this branch.

packages/datagateway-common/src/card/entityCard.component.tsx
GoelBiju: @sam-glendenning what I am thinking of at the moment is to make use of the `descriptionRef` and getting the `clientWidth` which can then be used as the `width` of the `shadowVisible` class instead of using `minWidth`. The only issue with this is that we can only apply that to the style after we get the width of the paragraph which means it might need to be rendered first (?)

sam-glendenning: Interesting... I'd be open to a pair programming session on this so you can show me how this is done?

GoelBiju: @sam-glendenning sure more than happy to! :smiley:

GoelBiju: This [solution](https://stackoverflow.com/questions/19014250/rerender-view-on-browser-resize-with-react) online helped address this issue.

sam-glendenning:
@GoelBiju I'm not entirely sure what's happening with this failing e2e test but I'm investigating. I'm also looking into trying to bump coverage to 100%

GoelBiju:
> @GoelBiju I'm not entirely sure what's happening with this failing e2e test but I'm investigating. I'm also looking into trying to bump coverage to 100%

I noticed it as well, not sure why it is failing at the moment.

GoelBiju (approved):

PR #720, opened 2021-06-09
#717 - use date-fns-tz to parse date strings with tz info

### LINKED ISSUE #717: Improve date handling in datagateway-download
**Description**:
Currently, due to the changes to dates required by the Payara 5 upgrade, datagateway-download crashes on isis prod due to "Invalid Date Error". We should improve how we handle our date parsing so this error doesn't occur

**Acceptance criteria**:
- [ ] No more errors when pointing to a Payara 5 TopCAT

louise-davies:
## Description
TopCAT can return strings with timezones on the end when run in Payara 5 so we need to be able to parse these otherwise we get an error. We use a library to help with parsing a string with a potential timezone at the end.

## Testing instructions
- [ ] Review code
- [ ] Check Actions build
- [ ] You can compare ISIS prod and ISIS dev - if you go to `/download` and go into the download tab, ISIS dev should throw an error and ISIS prod should be fine
 
## Agile board tracking
Closes #717

### DIFF
--- a/packages/datagateway-download/src/downloadStatus/adminDownloadStatusTable.component.tsx
+++ b/packages/datagateway-download/src/downloadStatus/adminDownloadStatusTable.component.tsx
@@ -39,7 +39,8 @@ import {
 } from '@material-ui/icons';
 import RefreshIcon from '@material-ui/icons/Refresh';
 import BlackTooltip from '../tooltip.component';
-import { toDate, format } from 'date-fns-tz';
+import { toDate } from 'date-fns-tz';
+import { format } from 'date-fns';
 
 const paperStyles = (theme: Theme): StyleRules =>
   createStyles({
--- a/packages/datagateway-download/src/downloadStatus/adminDownloadStatusTable.component.tsx
+++ b/packages/datagateway-download/src/downloadStatus/adminDownloadStatusTable.component.tsx
@@ -39,8 +39,7 @@ import {
 } from '@material-ui/icons';
 import RefreshIcon from '@material-ui/icons/Refresh';
 import BlackTooltip from '../tooltip.component';
-import { toDate } from 'date-fns-tz';
-import { format } from 'date-fns';
+import { toDate, format } from 'date-fns-tz';
 
 const paperStyles = (theme: Theme): StyleRules =>
   createStyles({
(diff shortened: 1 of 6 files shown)


packages/datagateway-download/src/downloadStatus/adminDownloadStatusTable.component.tsx
```diff
@@ -39,6 +39,8 @@ import {
 } from '@material-ui/icons';
 import RefreshIcon from '@material-ui/icons/Refresh';
 import BlackTooltip from '../tooltip.component';
+import { toDate } from 'date-fns-tz';
+import { format } from 'date-fns';
```
sam-glendenning, line 43: What was the reason for not using the `format` function from `date-fns-tz`? I couldn't notice any difference when I tried it out. I'm going to add a commit which removes the `date-fns` library and approve the PR because I'm happy with it anyway but if I'm missing something underlying with the timezones, feel free to revert that commit.

packages/datagateway-download/src/downloadStatus/downloadStatusTable.component.tsx
```diff
@@ -15,6 +15,8 @@ import { RemoveCircle, GetApp } from '@material-ui/icons';
 import BlackTooltip from '../tooltip.component';
 import { DownloadSettingsContext } from '../ConfigProvider';
 import { useTranslation } from 'react-i18next';
+import { toDate } from 'date-fns-tz';
+import { format } from 'date-fns';
```
sam-glendenning, line 19: As above

sam-glendenning (dismissed):

packages/datagateway-download/src/downloadStatus/adminDownloadStatusTable.component.tsx
louise-davies: You can use the format function from `date-fns-tz`, but we don't actually need the functionality of that library for the `format` function (aka appending timezone information strings). If we wanted to display timezone info in the table I think scientists would probably just prefer UTC offsets 😅  we do need to add `date-fns` to `package.json` as `date-fns-tz` depends on it so we don't actually reduce any overhead since we are already implicitly importing `format` from `date-fns` (as that's what `date-fns-tz` uses) - in fact using the `date-fns-tz` one would actually increase our footprint I think since the function does "more stuff" and so is larger

sam-glendenning: That makes a lot of sense! In which case feel free to revert my commit (or let me know if you want me to do it)

sam-glendenning (approved):

PR #724, opened 2021-06-11
Small changes to investigation card view #723

### LINKED ISSUE #723: Changes to investigation card view
**Description**:

**Change 1**:
To avoid ambiguity in results, we would like to change the word "Name" to "Dataset" in the sub-header in an investigation card view. See the below screenshots highlighting this:

image

Capture

In the above examples, this would change the wording to "Dataset: raw" and "Dataset: DATASET 268" and "Dataset: DATASET 28" respectively.

This ensures that users know that "raw" and "DATASET 268" refer to datasets explicitly and are the same types of data.

**Acceptance criteria**:
- [ ] The word "Name" is changed to "Dataset"

**Change 2**:
The link in the header of the investigation card view is difficult to read as it is dark blue with a blue background. We should change its color to an opposite contrast such as orange. It may be worth looking into [color-contrast](https://developer.mozilla.org/en-US/docs/Web/CSS/color_value/color-contrast()) but this may not be implemented in browsers yet. See the below screenshot:

image

**Acceptance criteria**:
- [ ] The link is changed to an appropriate color that is easier to read

sam-glendenning:
## Description
As per the issue:

- Changing "Name" to "Dataset" to reduce ambiguity of page results
- Changing color of hyperlink to orange in ISIS branding header to improve readability

The hyperlink color is the same as `ukri.bright.orange` in the UKRI theme. I decided to make it stay this color even after being `visited` as the default visited color is purple, which clashes with the blue background.

## Testing instructions
Navigate to an investigation card view such as `/browse/instrument/1/facilityCycle/14/investigation/87?view=card` to see changes

- [x] Review code
- [x] Check Actions build
- [ ] Compare [SciGateway](https://scigateway-preprod.esc.rl.ac.uk/browse/instrument/1/facilityCycle/14/investigation/87?view=card) with this version. You should see a different color hyperlink in the header and "Name" now replaced with "Dataset" in the sub-header.

## Agile board tracking
Closes #723


packages/datagateway-dataview/public/res/default.json
```diff
@@ -100,6 +100,7 @@
     }
   },
   "datasets": {
+    "dataset": "Dataset",
```
sam-glendenning, line 103: Decided to add a new field here instead of simply replacing "Name". The name field is used across the dataview and is most useful when sorting or filtering data (e.g. sorting by dataset name). By changing "name" to "Dataset", the column is also called "Dataset" but then sorting and filtering by "Dataset" doesn't really make sense. So I defined a new field to just be used in the investigation landing page.

packages/datagateway-dataview/src/views/landing/isis/isisBranding.component.tsx
```diff
@@ -22,6 +22,17 @@ const useStyles = makeStyles((theme: Theme) =>
     },
     typography: {
       color: theme.palette.primary.contrastText,
+      '& a': {
+        '&:link': {
+          color: '#FF6900',
```
louise-davies, line 27: Can't we get the colour directly from the theme object that is passed to us via SciGateway? We would need to ignore typescript for accessing unknown properties perhaps but it would mean not having to change the colour here if we change the colour in SciGateway

sam-glendenning: I wasn't sure how to reference this in all honesty. It's easy in SciGateway:
`(theme as UKRITheme)`

but `UKRITheme` doesn't exist in DataGateway. What's the best method of doing this?

louise-davies: It's a choice between either duplicating the colour code or perhaps ts-ignoring this line and accessing the property via optional chaining. Thoughts? I think I prefer the latter method.

louise-davies (approved):

PR #732, opened 2021-06-30
Preventing 'null' from displaying in descriptions and summaries

### LINKED ISSUE #715: Change display of "null" fields
**Description**:
We are currently showing "null" for fields that are null (or empty). We need to change this - perhaps leave it empty?

See summary field in the screenshot below (from DLS pre-prod)

Screenshot 2021-06-03 at 14 26 30

**Acceptance criteria**:
- [x] Fix the display of null fields

sam-glendenning:
## Description
This fixes an issue where 'null' would display on the page in cases where an investigation, dataset or datafile description was being displayed on the page but this variable wasn't actually set. Usually this means it was never provided with the data to begin with so we now show a more helpful message: 'Description not provided'.

## Testing instructions
Browse various investigations and datasets to find instances of 'Description not provided'. Expand the 'More Information' segment as well to ensure this message is displayed there too. It may help to manually edit the passed-in data so that no description value is provided but the snapshots test this themselves.

- [x] Review code
- [x] Check Actions build

## Agile board tracking
Closes #715

### DIFF
--- a/packages/datagateway-common/src/card/entityCard.component.tsx
+++ b/packages/datagateway-common/src/card/entityCard.component.tsx
@@ -273,7 +273,7 @@ const EntityCard = (props: EntityCardProps): React.ReactElement => {
                     variant="body1"
                     paragraph
                   >
-                    {description
+                    {description && description !== 'null'
                       ? description
                       : t('entity_card.no_description')}
                   </Typography>
--- a/packages/datagateway-dataview/public/res/default.json
+++ b/packages/datagateway-dataview/public/res/default.json
@@ -187,7 +187,7 @@
     }
   },
   "entity_card": {
-    "no_description": "Description not provided",
+    "no_description": "No description available",
     "show_more": "Show more",
     "show_less": "Show less"
   },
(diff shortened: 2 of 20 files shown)


sam-glendenning:
Looking through the schema and from browsing DataGateway myself, I don't notice places where a null field is being displayed except for a description or summary of an investigation, dataset or datafile. As such, this is the only field I'm currently applying this fix to.

With other null fields that are being accessed, the header is displayed and nothing below it. I've seen this with 'End date', for example. I can apply this fix to these other string fields as well if necessary, such as 'Start date', 'End date', etc. but this might create problems with sorting.

agbeltran:
thanks @sam-glendenning - yes, for other field types such as dates we may need to choose an appropirate null/empty value

sam-glendenning:
Final note for this one: an e2e test that was showing as failing on GH Actions was passing locally. A checkbox was flagging as not checked. This happened across multiple test runs and for some reason it passed in the latest run, despite the associated commit not doing anything to address the actual "bug". Long and the short of it seems to be that the Actions runs were failing for a random reason and it's now righted itself. However, feel free to run the e2e tests locally yourself to confirm everything is okay.

GoelBiju (approved):

PR #745, opened 2021-07-09
Fixing crashing study landing page #711

### LINKED ISSUE #711: Study landing pages sometimes crashes with e.investigation is undefined
**Description**:
Testing out the study landing pages on preprod and sometimes the page crashes with the above error. see: https://datagateway.isis.stfc.ac.uk/browseStudyHierarchy/instrument/2/study/101224979

**Acceptance criteria**:
- [x] Study landing pages don't crash anymore

sam-glendenning:
## Description
This addresses a problem which caused the ISIS study landing page to completely crash DataGateway. It occurred when the associated study was missing all of its known investigations (the query to /studyInvestigations would return some information about the study back but not the attached investigation). According to the ICAT schema, the investigation is a required field, hence why the app was crashing. 

As to why an investigation doesn't return from the API, that is still a mystery, _however_ it doesn't cause the page to crash when navigating to the investigations tab on the landing page. This loads the ISIS Investigations table component, which runs a separate query to fetch all investigations matching this instrument and study ID. In this case, we wouldn't know about any missing investigations because they wouldn't return or be displayed. 

Currently, the issue is being looked into on the API side of things so this addresses the DataGateway side of things and prevents the crash. Instead, what happens now when we detect missing investigations is we show a simple message saying 'Missing investigation' in place of where the investigation would go on the right hand side of the page. I'm open to suggestions on changing this displayed text as it might not be the right wording. 

An example query that caused the page to crash can be seen when navigating to https://datagateway.isis.stfc.ac.uk/browseStudyHierarchy/instrument/2/study/101224979. With this query, **three** investigations are being found by the API but only **one** is being returned. The below screenshot shows what will now display instead:

Capture

## Testing instructions

- [x] Review code
- [x] Check Actions build
- [x] Try navigating to a range of study landing pages to replicate the issue. Alternatively, you can mock the data being returned to the component but there is a unit and e2e test to demonstrate this

## Agile board tracking
Closes #711

### DIFF
--- a/packages/datagateway-dataview/cypress/integration/landing/isis/study.spec.ts
+++ b/packages/datagateway-dataview/cypress/integration/landing/isis/study.spec.ts
@@ -28,7 +28,7 @@ describe('ISIS - Study Landing', () => {
     );
   });
 
-  it('should display a missing investigation label when investigation is null', () => {
+  it('should load correctly when investigation missing', () => {
     cy.intercept('/studyinvestigations', [
       {
         createId: 'uows/1050072',
@@ -43,8 +43,6 @@ describe('ISIS - Study Landing', () => {
       },
     ]);
     cy.visit('/browseStudyHierarchy/instrument/1/study/405');
-    cy.get('[aria-label="landing-study-part-label"').contains(
-      'Missing investigation'
-    );
+    cy.get('#datagateway-dataview').should('be.visible');
   });
 });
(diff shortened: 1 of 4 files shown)


GoelBiju (dismissed):

louise-davies:
Alejandra: remove the "missing investigation" thing - but further investigation needed into what's going on

sam-glendenning:
Ready for review again after making the requested change

GoelBiju (approved):

Issue #752, opened 2021-07-16
Use ICAT Ansible Master on CI


MRichards99:
**Description**:
On the local CI, the `temp-icat-server-config-test` of ICAT Ansible is used. This is because Lucene 1.1.1 needs to be released (and the version number updated on the Ansible) to be able to use master. This is just a reminder to change branches once that work has taken place.

**Acceptance criteria**:
- [ ] Use ICAT Ansible's master branch on our CI

Issue #751, opened 2021-07-16
Fix Column Resizing End-To-End Test


MRichards99:
**Description**:
While working on local CI, I've added a skip to one of the column resize tests. I got this to work locally but it wasn't working on CI. Column resizing does work on this page but I just couldn't get the numbers to be the same. I'll fix this at a later date.

Skip added on https://github.com/ral-facilities/datagateway/commit/7854f25ea5a7de9c14056c106069e9bf11e2dfe0

**Acceptance criteria**:
- [ ] Get column resize test to pass on `datagateway-dataview/cypress/integration/table/dls/datafiles.spec.ts`

MRichards99:
Work finished, issue not automatically closed as PR wasn't merged directly into master, but now is in master.

Issue #749, opened 2021-07-15
Update CI to use Ubuntu 18 or 20


MRichards99:
**Description**:
As outlined by GitHub, [Ubuntu 16.04 will be removed on 20th September 2021](https://github.com/actions/virtual-environments#available-environments). Both DataGateway and DataGateway API use this version of Ubuntu and therefore need to be moved to 18 or 20 (ideally 20?). 

I've just quickly tested it and it seems ICAT Ansible has some issues getting packages (e.g. mariadb). Further investigation is needed on this to fully understand exactly what needs doing.

**Acceptance criteria**:
- [ ] GitHub Actions workflows need to run on a newer version of Ubuntu

PR #753, opened 2021-07-16
Implement Local ICAT Stack for GitHub Actions

### LINKED ISSUE #641: Change CI to use local ICAT & datagateway-api
**Description**:
Due to the firewall audit, we can't justify having those ports open just for testing when we have an alternative. We instead need to operate similar to `datagateway-api` for it's tests & set up an ICAT stack (although now we need a full stack all the way up to topcat/datagateway-download-api, plus actually creating fake files so the download tests will work) and `datagateway-api` server.

This will likely make it so that it is no longer beneficial to separate the plugin e2e tests into their own jobs, as that would require setting up the ICAT stack every time. Plus, GitHub Actions doesn't allow you to restart specific jobs anyway (unlike Travis), so the only sacrifice is being able to run the tests in parallel to save time, but I predict the time it takes to set up the servers 3 times vs 1 trumps this time save.

**Acceptance criteria**:
- [ ] CI no longer relies on access to `scigateway-preprod` to run tests successfully

MRichards99:
## Description
This closes #641. This branch means the CI is no longer dependent on preprod. The CI creates its own ICAT stack, populates data into databases and runs the repo's tests. 

I've made a couple of additions to the CI. I've added the `workflow_dispatch` keyword at the top of the workflow which allows anyone to start a CI run on any branch. This means developers won't need to open a PR just to run the CI, they can just start one by clicking the CI Build button workflow and click the Run workflow drop down. This feature will become available once this is merged into master. The feature has already been implemented on DataGateway API:
image

I've also used the [artifact feature](https://docs.github.com/en/actions/guides/storing-workflow-data-as-artifacts) within GitHub Actions to upload the Cypress screenshots when an e2e test fails. I've spent so many hours working out why a CI test fails but it passes locally so being able to see the screenshots has helped so much. Here's an [example](https://github.com/ral-facilities/datagateway/actions/runs/1007797831), it's at the bottom of the page, just click on it to download the screenshots in a zip.

The CI used to do the unit tests before moving onto the e2e tests to avoid conflicts between tests. Now each job has its own ICAT stack, this is no longer a problem so all 4 jobs can run in parallel, taking [between 22-25 minutes](https://github.com/ral-facilities/datagateway/actions/runs/1037213161). I did try running linting and all tests sequentially (all on 1 job so they share the same ICAT stack) but I found this to be [slower](https://github.com/ral-facilities/datagateway/actions/runs/1036937506), and the Search e2e tests didn't execute on this run as download had failed.

I've added a skip to one of the resize column tests on DLS datafiles on table view as I couldn't get that to work and didn't want to hold this up any longer. I've created a separate issue for this (#751) which I'll fix in the future. The functionality works, just the column width wasn't exactly right.

The CI still uses the `temp-icat-server-config-test` I was using to get things working. The changes on that branch have mostly been implemented onto that repo's master but ICAT Lucene 1.1.1 needs to be released before we can use master. There's no reason not to use that branch for the time being, it's causing no harm. I've created #752 as a reminder of this.

I've cleaned up the commit history a bit (it was a right mess...) but if you spot any other commits which can be rebased/squashed/removed, let me know. 

## Testing instructions
Hopefully the CI on this PR will pass to demonstrate that it works, then I guess it's a case of looking over the code changes

- [ ] Review code
- [ ] Check Actions build
- [ ] Check commit history for anything that could be made cleaner

## Agile board tracking
connect to #641


.github/workflows/ci-build.yml
```diff
@@ -1,22 +1,112 @@
 name: CI Build
 on: 
+  workflow_dispatch:
   pull_request:
   # By default, the pull_request event type is not triggered when a PR is merged into master
   push:
     branches:
       - master
+      - use-local-ICAT-in-CI-runs-#641
```
sam-glendenning, line 9: Is this still needed? Was it being used for testing just on this branch?

.github/workflows/ci-build.yml
```diff
@@ -1,22 +1,112 @@
...
+      - name: Install dependencies
+        run: cd datagateway-api; poetry install; cd ..
+      - name: Add dummy data to icatdb
+        run: |
+          cd datagateway-api; poetry run python -m util.icat_db_generator; cd ..
+
+      - name: Start API
+        run: cd datagateway-api; nohup poetry run python -m datagateway_api.src.main > api-output.txt &
+      - name: Back a dir
+        run: cd ..
```
sam-glendenning, line 100: This step might not be needed. I remember I added this because I was experimenting with moving between directories and wanted to make sure I was definitely in the home dir before executing any further commands. I'll try and double check this locally

.github/workflows/ci-build.yml
```diff
@@ -68,13 +162,109 @@ jobs:
...
```
sam-glendenning, line 265: As above

.github/workflows/ci-build.yml
```diff
@@ -99,15 +289,134 @@ jobs:
...
```
sam-glendenning, line 417: As above

.github/workflows/ci-build.yml
```diff
@@ -132,15 +441,134 @@ jobs:
...
+          curl -k --request POST 'https://localhost:8181/icat/lucene/db/Investigation/1' --header 'Content-Type: application/x-www-form-urlencoded' --data-urlencode 'sessionId='`jq -r '.sessionId' login_output`
+      - name: Let reindexes complete
+        run: sleep 30
+      - name: Remove session ID data
+        run: rm -f login_output
+
+      - name: Start API
+        run: cd datagateway-api; nohup poetry run python -m datagateway_api.src.main > api-output.txt &
+      - name: Back a dir
+        run: cd ..
```
sam-glendenning, line 569: As above

packages/datagateway-download/server/e2e-settings.json
```diff
@@ -1,18 +1,20 @@
...
+  "apiUrl": "http://localhost:5000",
   "downloadApiUrl": "https://localhost:8181/topcat",
   "idsUrl": "https://localhost:8181/ids",
-  "icatUrl": "https://scigateway-preprod.esc.rl.ac.uk:8181/icat",
+  "icatUrl": "https://localhost:8181/icat",
   "fileCountMax": 5000,
   "totalSizeMax": 1000000000000,
   "accessMethods": {
     "https": {
+      "_comment": "go back to this",
```
sam-glendenning, line 11: What's this used for?

sam-glendenning:
So what's the advice for running tests locally? I typically use Windows so not sure if the setup for running tests would translate well to this. I can make use of a Linux VM but it's more fidgety. Do I work through the steps provided by the yaml file to run the tests manually or is there any specific command I can use to just run it all automatically?

louise-davies:
@sam-glendenning I will update scigateway-preprod to use the new db generator script so that the e2e tests can continue to be run against it for developers. It might then be useful to make use of environment variables to set up the relevant URLs for e2e settings development and for CI

MRichards99:
@sam-glendenning for reference I've been running the tests locally doing very similar things to the yaml file (the things I fixed out on `bugfix/fix-e2e-tests-#712` formed the additions to the yaml). Unless you feel particularly keen to make a local ICAT stack, it's probably just better to use preprod once that's updated.

packages/datagateway-download/server/e2e-settings.json
MRichards99: If only JSON supported proper comments, removed in a6eae53717000e841f357e6d77c1165c99784575

.github/workflows/ci-build.yml
MRichards99: Ah yes, removed in https://github.com/ral-facilities/datagateway/pull/753/commits/fd1ce9bed3130d8b505614cbd37a5be20ad4e318

.github/workflows/ci-build.yml
```diff
@@ -9,14 +10,100 @@ on:
...
       - name: Checkout repo
         uses: actions/checkout@v2
       - name: Setup Node.js
         uses: actions/setup-node@v1
         with:
           node-version: '14.x'
+
       # Cache yarn dependencies/ restore the cached dependencies during future workflows
       - name: Get yarn cache directory path
         id: yarn-cache-dir-path
```
sam-glendenning, line 109: I could only comment on a limited number of lines but come to think of it, I can't think of any reason why this is here. We don't need any ICAT stack setup for the regular old unit tests, right? There's no overlap with the e2e tests, is there?

MRichards99: Umm I'm actually not sure, I think I just pasted in the setup just cause those red crosses are sad to look at. One would assume ICAT might not be needed though? I'll remove it and see what happens :)

MRichards99: I've removed the ICAT and DataGateway API setup and it seemed all good as per https://github.com/ral-facilities/datagateway/actions/runs/1049320004

.github/workflows/ci-build.yml
```diff
@@ -9,14 +10,27 @@ on:
...
+      - name: Setup Java
+        uses: actions/setup-java@v1
+        with:
+          java-version: 8
+          java-package: jdk
+      - name: Setup Python
+        uses: actions/setup-python@v2
+        with:
+          python-version: 3.x
+          architecture: x64
```
sam-glendenning, line 28: With your last comment in mind, @MRichards99, we can probably also remove these lines too. This takes us back to what is currently on master, which works fine. There's no need for the universe repo, java or python in these steps.

sam-glendenning (changes requested):

sam-glendenning (approved):
Looks good to me! Thanks for handling this Matt and cleaning up the mess I left this in when you took over! All we have to do is remember to upgrade the Ubuntu version and icat-ansible branch used when the time comes that this won't break things but looks like there's issues for these in #749 and #752

MRichards99:
Not a problem! I think the Ubuntu 18/20 stuff will be sorting out versions of packages for the ICAT stack as defined in ICAT Ansible - I'll probably talk to Stuart before tackling this as he might have some suggestions that might prevent me from going round the houses.

PR #756, opened 2021-07-19
ISIS recommended changes

### LINKED ISSUE #741: Remove description column from facility cycle table
**Description**:
This is always the same value in the ISIS database, so this is kind of redundant.

**Acceptance criteria**:
- [x] This column is no longer present

### LINKED ISSUE #742: In Study view, remove study.description and add investigation.title & add investigation dates
**Description**:

In the study view, we are showing empty study.description column - we need to change it to show the investigation.title instead

Similarly the dates are empty and we should show the investigation dates (start date is the first experiment part start date).

Screenshot 2021-07-08 at 09 54 36

**Acceptance criteria**:
- [x] Showing columns with data in study (coming from investigation)

### LINKED ISSUE #743: In the Investigation list, remove visitID and only show it as part of the expanded metadata view
**Description**:

The column Visit ID, which corresponds to the experiment part, is not so useful for the users so we should remove the column from the view.

Screenshot 2021-07-08 at 10 03 28

https://datagateway.isis.stfc.ac.uk/browse/instrument/1/facilityCycle/1/investigation

We still need to show this info in the expanded metadata, but ```Investigation.details.visitId``` should be ```Experiment part```

Screenshot 2021-07-08 at 10 05 39

**Acceptance criteria**:
- [x] Remove visitID column
- [x] change label for visitId to experiment part

GoelBiju:
## Description

Addresses the changes suggested by ISIS to DataGateway. These are issues #741, #742, and #743.

## Testing instructions

- [x] Review code
- [x] Check Actions build

## Agile board tracking

Closes #741, closes #742, closes #743


sam-glendenning (changes requested):

sam-glendenning:
Shame about those failing tests :/ I'm happy to have a pair programming sesh with you on this!

packages/datagateway-dataview/cypress/integration/table/isis/facilityCycles.spec.ts
```diff
@@ -24,13 +24,14 @@ describe('ISIS - FacilityCycles Table', () => {
...
-  it('should be able to resize a column', () => {
+  // TODO: Having three columns does not work since the width not calulated correctly.
+  it.skip('should be able to resize a column', () => {
     let columnWidth = 0;
 
     cy.window()
       .then((window) => {
         const windowWidth = window.innerWidth;
-        columnWidth = windowWidth / 4;
+        columnWidth = windowWidth / 3;
```
sam-glendenning, line 34: It's interesting because the columnWidth variable comes to 1000 / 3, which is exactly what the width of the columns should be and yet it determines them to be 327.666...

packages/datagateway-dataview/cypress/integration/table/isis/investigations.spec.ts
```diff
@@ -24,7 +24,8 @@ describe('ISIS - Investigations Table', () => {
     cy.get('[aria-rowcount="75"]').should('exist');
   });
 
-  it('should be able to resize a column', () => {
+  // TODO: Removal of the "visitID" column is breaking this resizing test.
+  it.skip('should be able to resize a column', () => {
```
sam-glendenning, line 28: Another interesting part - dividing (windowWidth - 40 - 40) by 7 instead of 8 (which is correct, as there are now 7 columns instead of 8) gives 131.4285 while it detects the width of the columns to be 131.4166. Ridiculously close to correct but it makes you wonder why. But I'm also unsure where the values of 40 are coming from in that calculation.

packages/datagateway-dataview/cypress/integration/table/isis/facilityCycles.spec.ts
GoelBiju: @sam-glendenning the issue with this is the scrollbar being present when the data is loaded in the table. The CSS width for the table however is `1000px`. One possible way to get this to work might be to stop the data from being loaded for the specific e2e test and then check the width is what we expect it to be.

packages/datagateway-dataview/cypress/integration/table/isis/investigations.spec.ts
GoelBiju: I was discussing this with @louise-davies and have yet to see why this isn't correct

GoelBiju:
> Shame about those failing tests :/ I'm happy to have a pair programming sesh with you on this!

Sure @sam-glendenning we can organise a session to discuss what is going on.

packages/datagateway-dataview/cypress/integration/table/isis/facilityCycles.spec.ts
GoelBiju: As @sam-glendenning has done in the latest commit, we now use the filtering box to enter a filter term that does not match the data to ensure no data is loaded. As a result, a scrollbar does not appear. To get around the issue of the width values not matching exactly, we round it to 1 decimal place and then compare them with the column width.

packages/datagateway-dataview/cypress/integration/table/isis/investigations.spec.ts
GoelBiju: This has been solved by rounding the widths.

packages/datagateway-dataview/server/e2e-settings.json
```diff
@@ -6,9 +6,9 @@
     "datasetGetSize": false,
     "datasetGetCount": true
   },
-  "idsUrl": "https://localhost:8181/ids",
-  "apiUrl": "http://localhost:5000",
-  "downloadApiUrl": "https://localhost:8181/topcat",
+  "idsUrl": "https://scigateway-preprod.esc.rl.ac.uk:8181/ids",
+  "apiUrl": "https://scigateway-preprod.esc.rl.ac.uk/datagateway-api",
+  "downloadApiUrl": "https://scigateway-preprod.esc.rl.ac.uk:8181/topcat",
```
sam-glendenning, line 11: Hi Goel, these changes should be reverted as Matt has now merged in work that makes use of a local ICAT stack for e2e tests. The dependency on preprod has therefore been removed and should only be used when testing locally.

GoelBiju:
At present this PR is failing due to intermittent end-to-end test errors on DataView. @MRichards99 is also looking into timezone issues with the test data to see if there are any issues.

MRichards99:
Status update before my leave:

I haven't really been able to find too much in my limited time investigating this. I did notice that not all parts of DataGateway use the time (and timezone). Generic investigation table and investigation search table only display the date - why is this? I did some git blames and traced the addition of this code to https://github.com/ral-facilities/datagateway/commit/24a4bc5065f39ddf3460aced1b39b956dc5b1d0c and https://github.com/ral-facilities/datagateway/commit/9618f9b5c4c5012f2339824ebbaa981f0dec4a3f with no specific explanation. Now I'm not saying we remove the time from everywhere else on the frontend, but I thought it was interesting (potentially thought provoking), particularly in absence of the actual cause of the issue shown in CI/preprod.

When I return from leave, I'll try to investigate further if a solution hasn't been found. I think my access to the preprod machine needs to be resolved as that's the only place with the issue of mixed timezones. I might try to regenerate the data at first point, then maybe if I can force the timezone on the generator script, as suggested by Louise.

sam-glendenning:
I hadn't noticed that inconsistency across tables before. Louise may have an answer for this so worth bringing up to her

sam-glendenning:
Failing tests are now being skipped due to data mismatches as described in #782. The intention is to get this changes merged and sort out the deeper problems in the CI in another PR.

sam-glendenning (approved):

PR #757, opened 2021-07-19
Remove "rbNumber" references from code (#699)

### LINKED ISSUE #699: Fix references to "rbNumber"
**Description**:
The RB Number column used by ISIS is not a value named "rbNumber" in the database, instead it is the investigation name just with a different label. Any references to "rbNumber" will not display any actual data

**Acceptance criteria**:
- [x] Remove references to RB number in the code - it is purely a translation/label issue. Instead ISIS investigation views should have a name column instead

GoelBiju:
## Description

Removes the use of "rbNumber" in code for table/card views and instead makes use of the investigation name as specified in #699. ISIS investigation view has a "name" column instead.

## Testing instructions

- [x] Review code
- [x] Check Actions build

## Agile board tracking

Closes #699


packages/datagateway-search/src/table/investigationSearchTable.component.tsx
```diff
@@ -227,8 +227,7 @@ const InvestigationSearchTable = (
         return (
           <div>
             <Typography>
-              <b>{t('investigations.rb_number')}:</b>{' '}
-              {investigationData.rbNumber}
+              <b>{t('investigations.name')}:</b> {investigationData.name}
```
sam-glendenning, line 230: Careful here. I don't think this reference actually exists in datagateway-search's `res/default.json`. `rb_number` is there though, so it may be worth replacing `rb_number` with `name`

packages/datagateway-search/src/table/investigationSearchTable.component.tsx
```diff
@@ -259,8 +258,8 @@ const InvestigationSearchTable = (
           filterComponent: textFilter,
         },
         {
-          label: t('investigations.rb_number'),
-          dataKey: 'rbNumber',
+          label: t('investigations.name'),
```
sam-glendenning, line 261: As above

sam-glendenning (changes requested):
So there are still multiple references to rbNumber in the code and I'd like to know if this is intentional or not. Perhaps it has a purpose I'm unaware of.
- The `Investigation` type declaration in `common/src/app.types.tsx` contains `rbNumber`
- Test data in `datasetSearchTable` and `datafileSearchTable` contains `rbNumber` references (lines 65 and 75 respectively) which is appearing in their relevant snapshots
- Multiple e2e tests that are being skipped for testing ISIS study tables (RB Number is the header for the first column containing the study name). One of these tests is _not_ being skipped (see line 142 of `table/isis/studies.spec.ts`)
- RB Number still appears in `res/default.json` of dataview and search

packages/datagateway-search/src/table/investigationSearchTable.component.tsx
GoelBiju: Thanks, @sam-glendenning, I had overlooked the `res/default.json` file. Made the changes now.

GoelBiju:
> So there are still multiple references to rbNumber in the code and I'd like to know if this is intentional or not. Perhaps it has a purpose I'm unaware of.
> 
> * The `Investigation` type declaration in `common/src/app.types.tsx` contains `rbNumber`
> * Test data in `datasetSearchTable` and `datafileSearchTable` contains `rbNumber` references (lines 65 and 75 respectively) which is appearing in their relevant snapshots
> * Multiple e2e tests that are being skipped for testing ISIS study tables (RB Number is the header for the first column containing the study name). One of these tests is _not_ being skipped (see line 142 of `table/isis/studies.spec.ts`)
> * RB Number still appears in `res/default.json` of dataview and search. Is this best being removed/replaced with Name instead?

Thanks @sam-glendenning, these are all really helpful. So just to address them individually:

* The `rbNumber` can be kept in the `Investigation` type declaration since we still receive the data from the API (as mentioned by @louise-davies),

* I have added the `rbNumber` which I removed from `investigationSearchTable` unit test since it was supposed to mock the data returned from the API (where `rbNumber` is present),

* It seems those were there a long time, I'll have a look to see if those tests are working or not,

* I think the only question I had was whether we should be renaming the existing "RB Number" columns for ISIS study views to "Name"?

louise-davies:
@GoelBiju there is no `rbNumber` field that is returned by the API - this is just a label given to the `investigation.name` field in ISIS. This can be seen in the ICAT schema in that there are no `rbNumber` fields. Any existing references should be changed to reflect the fact that the field is actually `investigation.name`. There should thus therefore be absolutely no references to RB number at all, except maybe in `default.json` (although the default should just be "Name" in my opinion and have "RB Number" configured specifically in the ISIS language files over on Gitlab)

sam-glendenning:
I agree with Louise, happy to approve once these changes have been made :)

GoelBiju:
> @GoelBiju there is no `rbNumber` field that is returned by the API - this is just a label given to the `investigation.name` field in ISIS. This can be seen in the ICAT schema in that there are no `rbNumber` fields. Any existing references should be changed to reflect the fact that the field is actually `investigation.name`. There should thus therefore be absolutely no references to RB number at all, except maybe in `default.json` (although the default should just be "Name" in my opinion and have "RB Number" configured specifically in the ISIS language files over on Gitlab)

I have removed all of the references to "RB Number" in code, tests, and in the `default.json` file as well (replacing it with "Name").

sam-glendenning (approved):

PR #768, opened 2021-07-23
Fix sorting on instrument column for generic investigation table

### LINKED ISSUE #767: Sorting on Instrument column on generic investigation table view not working
**Description**:
When trying to sort by instrument on generic investigation table, a 400 comes back from the API, thanks @sam-glendenning for the screenshots ;)
image

This is because of an incorrect order filter input, it should be `investigationInstruments.instrument.name`:
image

Having done a quick search of the codebase, it looks as if this is the only instance of it happening so doesn't need fixing elsewhere.

**Acceptance criteria**:
- [ ] On the generic investigation table, sorting by investigation should work, resulting in a 200 from the API request

MRichards99:
## Description
Tiny PR to fix #767. This means sorting on that column will occur on a one-to-many relationship, **Python ICAT 0.19.0 is required**, ideally with the changes from DataGateway API issue #238 (the corresponding PR for that brings both. If you don't have that, you'll get a 400 from the API request saying you cannot sort on a one to many relationship.

Screenshot below shows this works - both ascending and descending requests now return 200:
image

The second commit in this PR actually makes data present in that column, as proven when pointing at ISIS dev:
image

## Testing instructions
Discussed between Sam and I over Slack

- [ ] Review code
- [ ] Check Actions build
- [ ] {more steps here}

## Agile board tracking
connect to #767

### DIFF
--- a/packages/datagateway-dataview/src/views/table/investigationTable.component.tsx
+++ b/packages/datagateway-dataview/src/views/table/investigationTable.component.tsx
@@ -247,6 +247,15 @@ const InvestigationTable = (
           icon: <AssessmentIcon />,
           label: t('investigations.instrument'),
           dataKey: 'investigationInstruments.instrument.name',
+          cellContentRenderer: (cellProps: TableCellProps) => {
+            const investigationData = cellProps.rowData as Investigation;
+            if (investigationData?.investigationInstruments?.[0]?.instrument) {
+              return investigationData.investigationInstruments[0].instrument
+                .name;
+            } else {
+              return '';
+            }
+          },
           filterComponent: textFilter,
         },
         {
@@ -281,7 +290,19 @@ const mapDispatchToProps = (
   dispatch: ThunkDispatch<StateType, null, AnyAction>
 ): InvestigationTableDispatchProps => ({
   fetchData: (offsetParams: IndexRange) =>
-    dispatch(fetchInvestigations({ offsetParams })),
+    dispatch(
+      fetchInvestigations({
+        offsetParams,
+        additionalFilters: [
+          {
+            filterType: 'include',
+            filterValue: JSON.stringify({
+              investigationInstruments: 'instrument',
+            }),
+          },
+        ],
+      })
+    ),
   fetchCount: () => dispatch(fetchInvestigationCount()),
 
   addToCart: (entityIds: number[]) =>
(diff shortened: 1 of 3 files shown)


packages/datagateway-dataview/src/views/table/investigationTable.component.tsx
```diff
@@ -246,7 +246,16 @@ const InvestigationTable = (
         {
           icon: <AssessmentIcon />,
           label: t('investigations.instrument'),
-          dataKey: 'instrument.name',
+          dataKey: 'investigationInstruments.instrument.name',
+          cellContentRenderer: (cellProps: TableCellProps) => {
+            const investigationData = cellProps.rowData as Investigation;
+            if (investigationData?.investigationInstruments?.[0]?.instrument) {
+              return investigationData.investigationInstruments[0].instrument
```
sam-glendenning, line 253: I think if you wanna get this line covered, you'll have to change the test data from this:
```
state.dgcommon.data = [
    {
        ....
        instrument: {
            name: 'LARMOR',
        },
    }
]
```
to something like this (I borrowed it from `investigations.text.tsx`):
```
state.dgcommon.data = [
    {
        ....
        investigationInstruments: [
            {
                id: 3,
                instrument: {
                    id: 4,
                    name: 'LARMOR',
                },
            }
        ],
    }
]
```

sam-glendenning (changes requested):

MRichards99:
CI failures seem to be due to a timeout on download and a cypress issue (the deconstruct packet thing) on search. Will re-run them and see if it fixes itself

packages/datagateway-dataview/src/views/table/investigationTable.component.tsx
MRichards99: I definitely would've forgotten to sort out the snapshot if it wasn't for the reminder :)

MRichards99:
deconstruct packet errors on download and search this time, hmm

MRichards99:
Think I'll leave this for the weekend and rerun on Monday morning to see if the errors have cleared

MRichards99:
Seems as if the Cypress errors from Friday afternoon have fixed themselves this morning - ready for review :)

sam-glendenning (approved):

PR #774, opened 2021-07-28
Center and maximise width of search table (#693)

### LINKED ISSUE #693: Centre the search-box and search-table containers
**Description**:
The search-box and search-table (results) container are left justified. I think they will look better if centred. 

**Acceptance criteria**:
- [x] show search-box and search-table containers in the centre of the page

Screenshot 2021-05-18 at 12 24 36

### LINKED ISSUE #778: Search filters/table exceeds width of plugin container
**Description**:

Due to centering the search filters and search table in `datagateway-search` they exceed the width of the actual container in which the plugin loads when the sidebar is opened. This might be due to us setting the width of the filters/table by specifying the viewport width. We need to ensure the filters to not just move to the left to keep the same width but instead resize when the sidebar is opened on datagateway-search.

image

**Acceptance criteria**:

- [x] Search filters and search table do not exceed the width of the plugin container.

GoelBiju:
## Description

Addresses #693 by centering the search table and search filters on the page.

## Testing instructions

- [x] Review code
- [x] Check Actions build

## Agile board tracking

Closes #693, closes #778


sam-glendenning:
It may be useful to merge work from https://github.com/ral-facilities/datagateway/pull/761 to fully test new functionality

GoelBiju:
> It may be useful to merge work from #761 to fully test new functionality

That's a good idea @sam-glendenning, I'll test it out to make sure everything works correctly.

MRichards99:
Just as a note, those CI failures on DataView look like intermittent ones regarding `getSize` - Sam and I quickly looked at them the other day, the `/getSize` request seems to 400 for some reason. We couldn't reliably recreate the failures and so investigating the cause was tricky. Rerunning them should clear the failures for you

GoelBiju:
> Just as a note, those CI failures on DataView look like intermittent ones regarding `getSize` - Sam and I quickly looked at them the other day, the `/getSize` request seems to 400 for some reason. We couldn't reliably recreate the failures and so investigating the cause was tricky. Rerunning them should clear the failures for you

Thanks @MRichards99 for telling me about it 😃.

sam-glendenning (changes requested):
This centres everything fine but it's not the best solution at the moment and I think you've highlighted some of this in #778. For example, resizing the window can cause both horizontal and vertical scroll bars to appear:



which gets even worse when you expand the navigation drawer:



I don't believe this is caused by the tables, as I commented in #778, because search uses the same Table component as dataview which handles this fine. I believe it's caused by the parent element (the below screenshot shows this exact thing happens on preprod, demonstrated by zooming the page right in)



If we could find a way to address the problem of #778 in this PR and have it cover both issues, I think we'd have a fine solution that fixes the problem.

GoelBiju:
> This centres everything fine but it's not the best solution at the moment and I think you've highlighted some of this in #778. For example, resizing the window can cause both horizontal and vertical scroll bars to appear:
> 
> 
> 
> which gets even worse when you expand the navigation drawer:
> 
> 
> 
> I don't believe this is caused by the tables, as I commented in #778, because search uses the same Table component as dataview which handles this fine. I believe it's caused by the parent element (the below screenshot shows this exact thing happens on preprod, demonstrated by zooming the page right in)
> 
> 
> 
> If we could find a way to address the problem of #778 in this PR and have it cover both issues, I think we'd have a fine solution that fixes the problem.

@sam-glendenning like you mentioned it doesn't seem to be the tables but I think it might be due to the `vw` which we are setting on the `Grid` components to hold them in. I'll investigate this and see what can be done to fix it in this PR.

sam-glendenning:
The issue with the horizontal scroll bar has now been fixed. Previously I had mentioned a problem with a vertical scroll bar appearing as well but this was in error as it was actually caused by the height of the search box being greater than normal. This was caused by the labels not being translated in my local build, causing the text that appears to be longer than it would in deployment. This had the effect of overflowing them onto another row and increasing the height of the search box larger than expected. 

TL;DR - the vertical scroll bar issue was never an issue in the first place

sam-glendenning (dismissed):

sam-glendenning (approved):

PR #795, opened 2021-08-09
Fix and improve accessibility for sorting/filtering options (#763)

### LINKED ISSUE #763: Reduce Tab Presses on Card View Sorting Options
**Description**:
While tabbing through card view, I noticed that navigating through the sort options required two presses of the tab key to get to the next one:
image

This isn't a _major_ thing because a user can still sort the data - I wondered if there's a technical reason (ascending and descending maybe?) why this happens?

Opened this issue as a result of #738 

**Acceptance criteria**:
- [x] One tab press should get the user to the next sort field option

GoelBiju:
## Description

Fixes #763 for the sorting options on the `CardView` component and adds in aria-labels to sorting and filtering options that have been tested with the NVDA screen-reader.

## Testing instructions

- [ ] Review code
- [ ] Check Actions build

## Agile board tracking

Closes #763

### DIFF
--- a/packages/datagateway-common/src/card/__snapshots__/cardView.component.test.tsx.snap
+++ b/packages/datagateway-common/src/card/__snapshots__/cardView.component.test.tsx.snap
@@ -86,7 +86,7 @@ exports[`Card View renders correctly 1`] = `
             </Styled(MuiBox)>
             <Styled(MuiBox)>
               <WithStyles(ForwardRef(List))
-                aria-label="sorting-by-list"
+                aria-label="sort-by-list"
                 component="nav"
               />
             </Styled(MuiBox)>
--- a/packages/datagateway-common/src/card/__snapshots__/cardView.component.test.tsx.snap
+++ b/packages/datagateway-common/src/card/__snapshots__/cardView.component.test.tsx.snap
@@ -86,6 +86,7 @@ exports[`Card View renders correctly 1`] = `
             </Styled(MuiBox)>
             <Styled(MuiBox)>
               <WithStyles(ForwardRef(List))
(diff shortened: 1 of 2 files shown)


MRichards99 (commented):
This fixes the original issue I submitted so thanks for sorting that, although I've just noticed something else. When you sort on a field, the direction is displayed (asc/desc with a corresponding arrow). When you focus on that button, the current value/direction isn't picked up by a screen reader. Do you think it would be possible for it to be like 'sort by summary button, ascending' or something to that effect?

image

GoelBiju:
> This fixes the original issue I submitted so thanks for sorting that, although I've just noticed something else. When you sort on a field, the direction is displayed (asc/desc with a corresponding arrow). When you focus on that button, the current value/direction isn't picked up by a screen reader. Do you think it would be possible for it to be like 'sort by summary button, ascending' or something to that effect?
> 
> image

Sure @MRichards99 this can be implemented, one question I had would be whether it would show the current value/direction or the next direction after the button click?

MRichards99:
@GoelBiju If you could clearly mark out that it's the current direction, then I think that might be better. 'sort by summary button, current direction ascending' or something to that effect

GoelBiju:
> @GoelBiju If you could clearly mark out that it's the current direction, then I think that might be better. 'sort by summary button, current direction ascending' or something to that effect

I have now added this in @MRichards99 in https://github.com/ral-facilities/datagateway/pull/795/commits/827e2ab0f65988d53402785b826c69b7708892f8.

MRichards99 (approved):
The direction is picked up by the screen reader, works well!

PR #813, opened 2021-08-26
Bugfix/aria labels #772

### LINKED ISSUE #772: Improve Aria Labels on DataView
**Description**:
There's quite a few aria-label related improvements that could be made on DataView, so I'll collate them all into this issue. I'm guessing I'll spot things on Download and on Search so I'll create separate issues for those components when I get to them.

I won't include anything related to the view toggle switch, as that's being redesigned in #718 and I've spoken to @GoelBiju about having an aria label which is sensible for the final design of that.

**Acceptance criteria**:
- [x] Does whether the use of container in "container table search button" make sense to users? Decide and fix if not
- [ ] ~Add gaps to table view information~ (check if we want to fix this or not)
- [x] Remove duplication on text and date filter components
- [x] Resolve unneeded gap caused by "..." on include example text
- [x] Make cog aria label more sensible
- [x] Fix aria labels outputting on screen readers on the first row of data (row 2 in code)
- [x] Remove duplication on pagination navigation element
- [x] Make screen reader output on max results drop down less confusing
- [x] Remove "button button" when tabbing through card view sort by columns. This could be automatically be fixed through the solution of #763 
- [x] On date filters, add an appropriate aria label to make the calendar button functionality clear
- [x] The page view button has an aria label of 'page view button', which, when read by a screen reader, will say 'page view button button'. Remove the word 'button' in the aria label.
- [x] The dates are read as 'start date date filter from edit from... (yyyy-MM-dd)', which sounds jarring. The 'start date date filter from' section is the aria label followed by the placeholder text 'from... (yyyy-MM-dd)'. We can't find a way of not making the placeholder text be read aloud so this is hard to fix right now
- [x] doi is pronounced as doy 
- [x] investigate inconsistencies with tab selecting from dropdown lists. Currently, it seems like tabbing to a dropdown requires you to press space to open the dropdown and then either space or enter to start interacting with it before requiring enter to be pressed to select an option
- [x] Focusing on data in the table using tabs sometimes allows for the screen reader to read out "grid grouping" or read the whole data in the table

GoelBiju:
## Description

PR for aria-label issues and improvements mentioned in #772.

## Testing instructions

- [x] Review code
- [x] Check Actions build

## Agile board tracking

Closes #772


packages/datagateway-common/src/card/cardView.component.tsx
```diff
@@ -451,15 +451,21 @@ const CardView = (props: CardViewProps): React.ReactElement => {
...
+                  <NativeSelect
                     id="select-max-results"
                     value={results}
+                    inputProps={{
+                      name: 'Max Results',
+                      id: 'max-results-native-helper',
+                    }}
                     onChange={(e) => {
-                      const newResults = e.target.value as number;
+                      const newResults = (e.target.value as unknown) as number;
```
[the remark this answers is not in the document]
AkhilMag13: Yes, I implemented the solution using the prop in this 3bf20e23292a3ebbf45f010d2cf50235d9e8ae8b and it is passing the current implementation of the unit tests

MRichards99 (commented):
I'm happy with the majority of this, with many of the improvements made actually much better than what I thought would be possible. I've found a couple of things which seem a little quirky:

I've noticed that when I focus on particular filter components, the screen reader/aria label describes the wrong column? On generic investigation table view, if I tab to the visit ID column header, it will tell me about dataset count. Likewise, if I tab all the way to the instrument column, it will tell me about both start date and end date. I'm not sure why this is, have any of you noticed/can reproduce this? Happy to reproduce for you on Zoom this afternoon if that helps.

When tabbing through filter components (the column headers), the screen reader will tell me about table (i.e. 'clickable table with 50 rows and 10 columns'). Sometimes, this will just be on the select all rows column, which I'd expect as this is the first time I've focused onto the table and is behaviour I'm happy with. Other times, it does this on each of the text filter components. This seems unnecessary as the user will have heard about the table the first time they focused on it, so doesn't need to be repeated each time. Is this something you've noticed while trying to split up the information?

MRichards99:
As discussed in sprint planning, the cause of the two issues I described above might be because not every column can be tabbed to. Perhaps adding tabindexes to ones which have no functionality might help, so at least they can be tabbed to

packages/datagateway-common/src/table/columnFilters/dateColumnFilter.component.tsx
```diff
@@ -73,16 +73,14 @@ const DateColumnFilter = (props: {
       <MuiPickersUtilsProvider utils={DateFnsUtils}>
         <KeyboardDatePicker
           style={{ whiteSpace: 'nowrap' }}
-          clearable
-          inputProps={{ 'aria-label': `${props.label} filter from` }}
+          inputProps={{ 'aria-label': `${props.label} filter` }}
```
GoelBiju, line 76: @sam-glendenning @AkhilMag13 I am adding the "from" that was removed in this line back in, otherwise the same aria-label is shared for both the "from" and "to" inputs.

packages/datagateway-common/src/table/columnFilters/dateColumnFilter.component.tsx
```diff
@@ -100,16 +98,14 @@ const DateColumnFilter = (props: {
         />
         <KeyboardDatePicker
           style={{ whiteSpace: 'nowrap' }}
-          clearable
-          inputProps={{ 'aria-label': `${props.label} filter to` }}
-          aria-hidden="true"
-          label="To...   [yyyy-MM-dd]"
+          inputProps={{ 'aria-label': `${props.label} filter` }}
```
GoelBiju, line 101: @sam-glendenning @AkhilMag13 as mentioned above, I am adding the "to" back here as well.

packages/datagateway-common/src/table/columnFilters/dateColumnFilter.component.tsx
AkhilMag13: There is a problem with this - the from and to are repeated twice when the screen reader is on the datepickers. It reads it out as: "Start Date filter from edit  From...  blank"

GoelBiju:
@sam-glendenning @MRichards99 @AkhilMag13  the CI build seems to be passing now. It was failing at random before due to data not being found but several re-runs and everything is passing. I am not sure if there is anyone work that needs to be added to this PR?

sam-glendenning:
Thanks for fixing it! Marking as ready now.

GoelBiju:
> I'm happy with the majority of this, with many of the improvements made actually much better than what I thought would be possible. I've found a couple of things which seem a little quirky:
> 
> I've noticed that when I focus on particular filter components, the screen reader/aria label describes the wrong column? On generic investigation table view, if I tab to the visit ID column header, it will tell me about dataset count. Likewise, if I tab all the way to the instrument column, it will tell me about both start date and end date. I'm not sure why this is, have any of you noticed/can reproduce this? Happy to reproduce for you on Zoom this afternoon if that helps.
> 
> When tabbing through filter components (the column headers), the screen reader will tell me about table (i.e. 'clickable table with 50 rows and 10 columns'). Sometimes, this will just be on the select all rows column, which I'd expect as this is the first time I've focused onto the table and is behaviour I'm happy with. Other times, it does this on each of the text filter components. This seems unnecessary as the user will have heard about the table the first time they focused on it, so doesn't need to be repeated each time. Is this something you've noticed while trying to split up the information?

I'll just reference these issues in #823 since these are still issues but at the moment there doesn't seem to be a way around it.

MRichards99 (approved):
I'm happy with this - hopefully we can solve #823 at some point, but I think this PR makes screen readers easier/more helpful and takes a big step from when I created the issue. Thanks for spotting (and then fixing) some things I missed, it's good to have multiple perspectives at this kind of thing!

PR #818, opened 2021-09-01
Unskipping MyData e2e tests #659

### LINKED ISSUE #659: Fix My Data e2e tests

sam-glendenning:
## Description
This changes the Cypress config to make use of an existing user instead of a 'test' user, in order to fetch existing data that is mapped to a user. The user Thomas409 happes to have the most investigations mapped to them so they were picked for this. Now that we have data returning from the API, we can now properly run e2e tests.

## Testing instructions
Check all relevant functionality is tested.

- [ ] Review code
- [x] Check Actions build

## Agile board tracking
Closes #659

### DIFF
--- a/packages/datagateway-dataview/cypress/integration/table/dls/myData.spec.ts
+++ b/packages/datagateway-dataview/cypress/integration/table/dls/myData.spec.ts
@@ -9,21 +9,21 @@ describe('DLS - MyData Table', () => {
     cy.visit('/my-data/DLS');
   });
 
-  it.skip('should load correctly', () => {
+  it('should load correctly', () => {
     cy.title().should('equal', 'DataGateway DataView');
     cy.get('#datagateway-dataview').should('be.visible');
   });
 
-  it.skip('should be able to click an investigation to see its datasets', () => {
+  it('should be able to click an investigation to see its datasets', () => {
     cy.get('[role="gridcell"] a').first().click({ force: true });
 
     cy.location('pathname').should(
       'eq',
-      '/browse/proposal/INVESTIGATION%201/investigation/1/dataset'
+      '/browse/proposal/INVESTIGATION%20131/investigation/131/dataset'
     );
   });
 
-  it.skip('should be able to resize a column', () => {
+  it('should be able to resize a column', () => {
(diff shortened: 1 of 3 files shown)


MRichards99:
I'm guessing to select a user that has a decent number of investigations assigned to them, you would've looked at a number of other users to work out which was most suitable; a selection process for fake users :) Just from your knowledge working on this issue, do users roughly have a similar number of investigations assigned to them or does Thomas409 have a bunch of investigations but the other users only have 1 or 2?

Just thinking about this in terms of the generator script and whether generated data could to be distributed differently/more evenly or not.

sam-glendenning:
@MRichards99 you're right, that's exactly what I did! Or at least what Louise did, I did nothing...

From various table lookups and the like, Thomas409 has 4 investigations mapped to them and all other users generally only have 1 or 2. This made them the ideal user to make use of. So they were definitely an outlier. Would it be good to map more investigations to more users? Possibly but I don't see a use case for that right now. The data is already distributed quite well and technically I only needed a user with at least 2 investigations mapped to them for testing sorting and filtering.

louise-davies:
@MRichards99 yeah, for reference, I ran this command in the database to find a user with at least two investigations:

`select USER_ID, COUNT(*) from INVESTIGATIONUSER GROUP BY USER_ID HAVING COUNT(*) > 1;`

Then I saw the specific user with 4 and chose them to use as the test user. As Sam said, I was just looking for a user with at least two for testing. Since the DB generator scripts produces this naturally for this seed I'd say not to worry about it (if the script hadn't made users with at least 2 investigations I would have suggested a change).

louise-davies (approved):

PR #825, opened 2021-09-07
Bugfix/minor fixes #815

### LINKED ISSUE #815: Minor fixes to EntityCard/CardView
**Description**:

There are some small issues that I wanted to address with the `EntityCard` and `CardView` components.

**Acceptance criteria**:

- [x]  the paper background which is apparent when dark mode is turned on,

- [x] the dividers for summary information which appears when the information array/prop is empty (could also be an issue with buttons array/prop),

- [x] ~hide pagination/filters/sort options when no results are loaded~ (this might not be useful anymore since we only really want to update the cards themselves on page/sort/filter changes, otherwise we do not want to show it when there are no results),

- [x] ~do not show advanced filters when data isn't loaded yet~ (it might be best to not let props like `totalDataCount` or `loadedData` control when the `AdvancedFilters` component is shown just for usability as it continues staying open for the user),

- [x] ~advanced filters jumping when no data is present and when data is loaded~ (this doesn't appear to be an issue)

- [x]  when a user changes pages, only change the cards themselves and not the rest of the UI (as mentioned by @louise-davies),

- [x] hide the no results message when loading data, only show it once loading has been performed and the `totalDataCount` is 0,

- [x] check filtering/no results message,

- [x] selecting filtering options hides all of the UI components until the `totalDataCount` is greater than zero; this is since all of the UI elements will hide based on `totalDataCount` (addressed by adding a `filterUpdate` prop which prevents the UI from hiding on a filter change),

- [x] ~"sort by" box does not reflect the summary information displayed on the card itself, shows all objects regardless if data is shown or not?~ (maybe we cannot control this since data for one card may not be shown but others will have it, sorting should work fine regardless if one card does not show the information for it unless we show a "N/A" if the specific information content to show for that card is empty): https://github.com/ral-facilities/datagateway/blob/fcae11c92d5440061eb41bf7550d60f9ff69fa37/packages/datagateway-common/src/card/entityCard.component.tsx#L217

- [x] ~investigate if advanced filters need icons defined, can we just pass them through as props from card view~ (crossing this off for now but it might be worth thinking about having the information needed to render the correct icon passed through to the advanced filter rather stating a list of icons to choose from): https://github.com/ral-facilities/datagateway/blob/fcae11c92d5440061eb41bf7550d60f9ff69fa37/packages/datagateway-common/src/card/advancedFilter.component.tsx#L67

- [x] fix "undefined" prop warnings when running unit tests on `datagateway-dataview`,

- [x] card chips do not load any information at the moment (or disable them for now); **fixed by adding an "include" filter to the data request which fetches the "type" and "facility"**,

- [x] better aesthetically to have the chips show above the more information or below it? (it looks better to keep it as it is since the "more information" section being above the tags associates the section with the rest of the information in the card),

- [x] it might be ideal to show instruments with chips/tags since then you can filter by the type e.g, "crystallography" (this can be an issue or feature request at a later point if needed),

- [x] add support for ~icons/avatar in chips, colour and~ prefixing the label name before the data (at present having an icon and changing the colour is not a requirement, changing the color is not supported through the `color` prop either but can be done through by changing `backgroundColor` in inline CSS although this is not as good as if it is supported),

- [x] check status of #265 to see if we can implement this (this has been now implemented in https://github.com/ral-facilities/datagateway/commit/fdc68cee98ee4e49c4b3a2c0c4ed9303e84808fa).

GoelBiju:
## Description

Addresses some small items with regards to the layout of the `CardView` component.

## Testing instructions

- [x] Review code
- [x] Check Actions build

## Agile board tracking

Closes #815


MRichards99:
I've pushed a fix for the investigation card view test that had been skipped. The CI is failing because it can't get file sizes from TopCAT. You can see from the screenshots in the Actions run that those requests 400 but I'm not sure why. I think my suggested action would be to merge master in (that will contain the latest Ubuntu 20/ICAT Ansible changes) but I don't want to interfere too much as this isn't my PR

AkhilMag13 (dismissed):

GoelBiju:
@AkhilMag13 like you mentioned, the `index.tsx` needs better code coverage. I'll look into improving the coverage of the new react-query hooks.

GoelBiju:
@AkhilMag13 this branch is failing due to the new unit tests for the `useCustomFilterCount` react-query hook not working. If you add a `.only` to this [line](https://github.com/ral-facilities/datagateway/pull/825/commits/8d87724489d50665d449e420a9ffd51b580e4101#diff-2588b4e297fec03d0d16928beebd13bd044040d340f4eb70a57a9c32cd596ab6R415) it will work on its own but not when run as a file. I haven't looked into why this is the case yet.

joelvdavies:
I believe the 'useCustomFilterCount' test is failing due to interference. Using `(axios.get as jest.Mock).mockClear()` in `afterEach` fixes the 'useCustomFilterCount' test but then breaks 'useCustomFilter'one instead.

joelvdavies:
I think this issue is now fixed. I am not entirely sure on the expected result for the 'useCustomFilter' test, but looking at the code I don't believe it will currently use the 'order' parameter so I removed it. Is this right?

PR #836, opened 2021-09-17
Rename instances of "master" to "main"

GoelBiju:
## Description

This PR is to try and rename instances of "master" to "main" in the repository when we change the default branch name from master to main.

Once it is changed, in your local development environments you may have to pull and checkout the new branch.


GoelBiju:
@MRichards99 I am not sure if there need to be other changes in order to change the branch name?

MRichards99 (approved):
I'm happy with these changes, but we should coordinate when the change will happen via the web interface, ideally on all three DataGateway repos at the same time. We should then be prepared to distribute the commands required for each developer to change their default branch locally

GoelBiju:
> I'm happy with these changes, but we should coordinate when the change will happen via the web interface, ideally on all three DataGateway repos at the same time. We should then be prepared to distribute the commands required for each developer to change their default branch locally

I agree @MRichards99, that would be the best way to handle this.

agbeltran (approved):

louise-davies (dismissed):

louise-davies (approved):

Issue #834, opened 2021-09-09
Add label to browse view when a user is logged in anonymously


louise-davies:
**Description**:
This label should emphasise to the user that they are only viewing the open data

**Acceptance criteria**:
- [ ] A user can read a banner of some sort telling them they are viewing open data when logged in anonymously

louise-davies:
maybe display this in the space before the results button? or near the card view toggle button? maybe some mockups are needed

GoelBiju:
> maybe display this in the space before the results button? or near the card view toggle button? maybe some mockups are needed

I think that near the results option sounds pretty good, I'll create wireframes with both and see what looks better.

agbeltran:
@GoelBiju please remember to include the wireframe here

GoelBiju:
@agbeltran @joelvdavies @louise-davies I've just added a label to the wireframes on both locations that were suggested:

1. Near the results box (more of an informative box):

![image](https://user-images.githubusercontent.com/16869075/134168453-f04ff69e-a070-4b5d-94fb-a7c6fc7a5119.png)

2. Near the "Display as cards" button (informative box):

![image](https://user-images.githubusercontent.com/16869075/134169904-db45e461-358f-4c02-bfd0-302a64b2ec06.png)

The following is using a warning box that @joelvdavies had already created:

3. Near the results (warning box):

![image](https://user-images.githubusercontent.com/16869075/134168976-a9cb8ecd-97d9-4a05-a2b4-81dc11bea038.png)

4. Near the "Display as cards" button (warning box):

![image](https://user-images.githubusercontent.com/16869075/134169789-16609348-e30a-4b9c-9d75-a09d52d84dcf.png)

Which do you think looks more appropriate?

agbeltran:
I prefer the information icon rather than the warning icon and the text "You are viewing open data", so a combination of the two. 

About the colour, we haven't used purple in the theme before, have we? Could we try perhaps green, as it is more indicative of information than warning?

If we have a new wireframe, let's wait for confirmation before implementing.

agbeltran:
I like it! and seems that it is already implemented? great

GoelBiju:
I think that looks really nice @joelvdavies !

joelvdavies:
@agbeltran @GoelBiju Thanks. What should I do? Commit/try out some tests? Or wait until we have confirmation?

GoelBiju:
[...]

PR #838, opened 2021-09-22
Add browsing anonymously label #834

joelvdavies:
## Description
This adds a label to the left of the results label in datagateway-dataview when the user is not logged in stating 'You are viewing open data'.

## Testing instructions
Still need to add tests.

- [ ] Review code
- [ ] Check Actions build

## Agile board tracking
Closes #{834}

### DIFF
--- a/packages/datagateway-dataview/src/page/pageContainer.component.tsx
+++ b/packages/datagateway-dataview/src/page/pageContainer.component.tsx
@@ -10,7 +10,6 @@ import {
   Badge,
   makeStyles,
   Button,
-  Tooltip,
 } from '@material-ui/core';
 import ShoppingCartIcon from '@material-ui/icons/ShoppingCart';
 import SearchIcon from '@material-ui/icons/Search';
@@ -190,15 +189,13 @@ const NavBar = React.memo(
                 justify="center"
               >
                 <Grid item>
-                  <Tooltip title="Only open data is displayed when browsing anonymously">
-                    <IconButton>
-                      <InfoIcon color="primary" />
-                    </IconButton>
-                  </Tooltip>
+                  <IconButton disabled>
+                    <InfoIcon color="primary" />
+                  </IconButton>
                 </Grid>
                 <Grid item>
                   <Typography color="inherit" variant="h6" component="h3">
-                    <b>You are viewing open data</b>
+                    <b>{t('app.view_open_data_warning')}</b>
                   </Typography>
                 </Grid>
               </Grid>
--- a/packages/datagateway-dataview/src/page/pageContainer.component.tsx
+++ b/packages/datagateway-dataview/src/page/pageContainer.component.tsx
@@ -10,9 +10,11 @@ import {
   Badge,
   makeStyles,
   Button,
+  Tooltip,
 } from '@material-ui/core';
 import ShoppingCartIcon from '@material-ui/icons/ShoppingCart';
 import SearchIcon from '@material-ui/icons/Search';
+import InfoIcon from '@material-ui/icons/Info';
 import { StyleRules } from '@material-ui/core/styles';
 import {
   DownloadCartItem,
@@ -149,6 +151,9 @@ const NavBar = React.memo(
   }): React.ReactElement => {
     const [t] = useTranslation();
 
+    //Determine whether logged in anonymously
+    const loggedInAnonymously = localStorage.getItem('autoLogin') === 'true';
+
     return (
       <Sticky>
         <StyledGrid container>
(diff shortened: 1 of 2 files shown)


packages/datagateway-dataview/src/page/pageContainer.component.tsx
```diff
@@ -149,6 +150,9 @@ const NavBar = React.memo(
   }): React.ReactElement => {
     const [t] = useTranslation();
 
+    //Determine whether logged in anonymously
+    const loggedInAnonymously = localStorage.getItem('autoLogin') === 'true';
```
louise-davies, line 154: A user can technically login as anonymous and not be `autoLogin === 'true'` - perhaps it's better to check that the username contains "anon" instead? (as the current anon username by default is `anon/anon`)

joelvdavies: Is there any reason not to require the username is exactly 'anon/anon', or maybe 'anon/' to avoid it falsely recognising someone with anon in their username?

louise-davies: Theoretically `authn.anon` could be setup with no mnemonic or a different mnemonic other than `anon`, meaning the anon username could be just `anon`, or `anonymous/anon`, or `unauthenticated/anon` etc. However this is a very specific problem and I don't think anyone ever configures `authn.anon` that way and so the default is pretty much always `anon/anon`.

However, I think in SciGateway we assume the anonymous authenticator is setup with `anon` as the mnemonic, so maybe it's fine and consistent to expect `anon/anon`

joelvdavies: @louise-davies Okay, I have changed it to that for now in the latest commit.

louise-davies:
From the meeting with ISIS on Friday, they liked the mock up we showed them, but they expressed that it would be useful for there to be some hover text perhaps explaining what open data is (link to ISIS page about open data policy?) and prompting a user to login to see their embargoed data.

Also, we need to ensure that this doesn't overlap visually with #840

louise-davies:
@joelvdavies This is the ISIS Data Policy page: https://www.isis.stfc.ac.uk/Pages/Data-Policy.aspx

joelvdavies:
@louise-davies Here is a screenshot of what I have at the moment. 
image

joelvdavies:
In dark mode the green looks like this colour looks like this:
image

louise-davies:
@joelvdavies I think it would be better to use the "standard" tooltip component - see `arrowtooltip` in `datagateway-common`. Additionally, in dark mode I think the light green with white text does not have sufficient contrast (needed in order to be compliant with accessibility standards) - a contrast checker site like https://contrastchecker.com/ may be helpful to experiment with

joelvdavies:
@louise-davies Here are updated screenshots:
image

image

packages/datagateway-dataview/src/page/pageContainer.component.test.tsx
```diff
@@ -240,4 +245,26 @@ describe('PageContainer - Tests', () => {
 
     expect(history.location.search).toBe('?view=card');
   });
+
+  it('displays warning label when browsing data anonymously', () => {
+    const response = { username: 'anon/anon' };
+    readSciGatewayToken.mockReturnValueOnce(response);
```
louise-davies, line 251: Is there a reason why this line doesn't need the `as jest.Mock` type assertion whereas the next test does? Just wondering because it looks a little inconsistent

louise-davies (changes requested):
I think the warning is a little too big, especially on smaller screens as it forces the breadcrumbs to wrap.

Additionally, I don't think the "i" should be a button as it doesn't do anything, but just display the tooltip on hover. This is both confusing (as clicking on it does nothing) and bad for accessibility (things should act like users expected them to act)

packages/datagateway-dataview/src/page/pageContainer.component.test.tsx
joelvdavies: This was when I was trying to fix some of the error highlighting VSCode was giving. I will change this one as well.

joelvdavies:
@louise-davies I have just pushed a commit that disables the button and makes the test more consistent. Did you want me to fully remove the button or is this sufficient as removing it breaks the alignment when using the icons. Also do you mean you want the entire warning paper smaller than the results paper?

joelvdavies:
@louise-davies That does seem strange, it works fine for me in Chrome and Firefox.

PR #839, opened 2021-09-22
 redirect to login page rather than showing warning (#833)

### LINKED ISSUE #833: "My data view" - redirect to login page rather than showing warning
**Description**:
Currently, when loading the "My data view", you get a warning if you are not logged in. It seems to me that it would be better to redirect to the login page directly. 

Screenshot 2021-09-09 at 15 44 43

**Acceptance criteria**:
- [ ] When loading the "My data view", redirect to login page if not logged in already

joshuadkitenge:
## Description
Changed the "My Data ISIS" and "MY Data DSL" so that it redirects you to the login page when the user isn't logged in, instead of getting a warning if you are not logged in.

## Testing instructions

- [ ] Review code
- [ ] Check Actions build
- [ ] {more steps here}

## Agile board tracking
closes #833

### DIFF
--- a/packages/datagateway-dataview/src/page/pageRouting.component.tsx
+++ b/packages/datagateway-dataview/src/page/pageRouting.component.tsx
@@ -296,6 +296,7 @@ SafeDLSDatasetsCardView.displayName = 'SafeDLSDatasetsCardView';
 interface PageRoutingProps {
   view: ViewsType;
   location: LocationType;
+  loggedInAnonymously: boolean;
 }
 
 class PageRouting extends React.PureComponent<PageRoutingProps> {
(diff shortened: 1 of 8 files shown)


packages/datagateway-dataview/src/page/pageRouting.component.tsx
```diff
@@ -317,8 +317,21 @@ class PageRouting extends React.PureComponent<PageRoutingProps> {
         />
 
         {/* My Data routes */}
-        <Route path={paths.myData.dls} component={DLSMyDataTable} />
-        <Route path={paths.myData.isis} component={ISISMyDataTable} />
+        <Route exact path={paths.myData.dls}>
+          {localStorage.getItem('autoLogin') === 'true' ? (
```
louise-davies, line 321: I made this similar comment to Joel so it might be worth looking at his "anonymous" PR, but I would prefer that this checks for anonymous as theoretically you can login as anonymous without being autoLogged in.

louise-davies (changes requested):

packages/datagateway-dataview/src/page/pageRouting.component.tsx
joshuadkitenge: I have implemented this yesterday, now I'm working on changing the testing, so it works for changes (anonymous login)

PR #842, opened 2021-10-01
My data role selector #797

### LINKED ISSUE #797: Improve the myData view for instrument scientists
**Description**:
We've had feedback that the myData view is currently not that useful for instrument scientists as they get back ALL investigations ran on their instrument, which makes finding investigations that they're actually the PI of a pain. Therefore, there should be two "views" for instrument scientists in myData - one that shows all their data as an experimenter and one that shows all their data as an instrument scientist.

Additionally, someone has replied that they can only see open data for their instrument and not all of the instrument data - but I suspect this is an ICAT issue with the user not having the correct role/the rules being slightly wrong.

**Acceptance criteria**:
- [ ] An additional myData "view" is created which only shows instrument scientists data that they're an experimenter on. This should just be part of the existing views and a toggle/button perhaps changes which requests are sent.
- [ ] Investigate the latter issue

louise-davies:
## Description
I've added a RoleSelector component that is rendered on my data views (currently it renders in the same place as card view button does as there's no "my data card view"). The roles are fetched by a query asking for all the distinct roles a user has. We leverage the fact that my data already filters based on "filters", so if we add our roles to the query params as filters then we don't have to modify the my data views at all.

This is a draft for now as I've written no tests or done much extensive testing - I just wanted to test the queries worked then it was relatively little effort to add the dropdown after sorting out the queries.

## Testing instructions

- [ ] Review code
- [ ] Check Actions build
- [ ] Review that dropdown looks okay and works

## Agile board tracking
Closes #797

### DIFF
--- a/packages/datagateway-dataview/src/page/pageContainer.component.tsx
+++ b/packages/datagateway-dataview/src/page/pageContainer.component.tsx
@@ -36,6 +36,7 @@ import { Location as LocationType } from 'history';
 import ViewListIcon from '@material-ui/icons/ViewList';
 import ViewAgendaIcon from '@material-ui/icons/ViewAgenda';
 import TranslatedHomePage from './translatedHomePage.component';
+import RoleSelector from '../views/roleSelector.component';
 import { useIsFetching, useQueryClient } from 'react-query';
 
 const usePaperStyles = makeStyles(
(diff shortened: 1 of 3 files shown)


louise-davies:
Role selector version 1

Here's my first version to show that the querying works & it filters the table properly - plus to gather thoughts on looks and  positioning of the dropdown.

MRichards99:
I think that looks good. When the gif goes from CI back to all, the drop down becomes blank - I wonder if this might be slightly unclear to users whether they've actually selected all or something strange has happened.

PR #846, opened 2021-10-04
Feature/add remove banner #840

### LINKED ISSUE #840: Add banner to indicate to users items have been added & removed from cart
**Description**:
See solution 3 here: https://stfc365.sharepoint.com/:p:/r/sites/TopcatRedevelopment/_layouts/15/Doc.aspx?sourcedoc=%7B610DBA01-EEDA-4272-866C-144EA13A2DF8%7D&file=DataGateway%20-%20Feedback%20and%20Changes.pptx&action=edit&mobileredirect=true

This banner should be very visible to users, and perhaps a little animation displayed to indicate that the value selected has changed (as ISIS wanted removals to be very visible)

**Acceptance criteria**:
- [ ] A banner added like in solution 3 of the mock ups that shows the current items added to the cart from the current view
- [ ] The banner updating is visible and clear when a user selects or deselects an item

joelvdavies:
## Description
This adds a banner at the stop of dataview and search pages when the user adds or removes items from their selections. This banner displays a colour change animation to make changes more obvious.

## Testing instructions
Still a work in progress.

- [ ] Review code
- [ ] Check Actions build
- [ ] {more steps here}

## Agile board tracking
Closes #840

### DIFF
--- a/packages/datagateway-common/src/views/selectionAlert.component.tsx
+++ b/packages/datagateway-common/src/views/selectionAlert.component.tsx
@@ -1,29 +1,49 @@
-import { Theme, createStyles, IconButton, makeStyles } from '@material-ui/core';
-import Alert from '@material-ui/lab/Alert';
+import {
+  Theme,
+  createStyles,
+  IconButton,
+  makeStyles,
+  Grid,
+  Paper,
+} from '@material-ui/core';
 import CloseIcon from '@material-ui/icons/Close';
 import { DownloadCartItem } from '../app.types';
 import React from 'react';
 import { useTranslation } from 'react-i18next';
 import { Link } from 'react-router-dom';
 
-//Note: Use widthPercentValue = -1 to auto fill to the screen
+//Note: By default auto fill to the avaiable space
 type SelectionAlertProps = {
-  widthPercentValue: number;
+  width?: string;
+  marginSide?: string;
 };
 
 const selectionAlertStyles = makeStyles<Theme, SelectionAlertProps>(
   (theme: Theme) =>
     createStyles({
       root: {
-        width: ({ widthPercentValue }) =>
-          widthPercentValue < 0 ? 'auto' : `${widthPercentValue}%`,
-        marginTop: '2px',
-        flex: '1',
-        marginRight: 'auto',
+        backgroundColor: 'orange',
+        color: 'white',
+        width: ({ width }) => (width === undefined ? 'auto' : width),
+        marginTop: '8px',
+        marginLeft: ({ marginSide }) =>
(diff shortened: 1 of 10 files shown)


packages/datagateway-search/cypress/integration/search/datafileSearch.spec.ts
```diff
@@ -72,6 +72,8 @@ describe('Datafile search tab', () => {
     cy.get('[aria-label="select all rows"]')
       .should('have.attr', 'data-indeterminate')
       .and('eq', 'false');
+
+    //
```
louise-davies, line 76: Random empty comment

packages/datagateway-common/src/views/selectionAlert.component.tsx
```diff
@@ -0,0 +1,168 @@
...
+
+    //Check for a change and assign text based on increase or decrease
+    if (newNumSelecItems !== numSelectedItems) {
+      const difference = newNumSelecItems - numSelectedItems;
+
+      //Obtain plural if needed
+      const itemText =
+        Math.abs(difference) > 1
+          ? t('selec_alert.items')
+          : t('selec_alert.item');
```
louise-davies, line 104: This is why we have an internationalisation library ;) see https://www.i18next.com/translation-function/plurals

packages/datagateway-common/src/views/selectionAlert.component.tsx
```diff
@@ -0,0 +1,168 @@
...
+            ${difference * -1} ${itemText} ${t('selec_alert.removed')}`);
+
+      setNumSelectedItems(newNumSelecItems);
+      //Change has occurred so need to ensure displayed
+      setAlertOpen(true);
+      setAnimating(true);
+    }
+
+    return (
+      <React.Fragment>
```
louise-davies, line 120: Why do we have a Fragment here? We only normally need these if we're returning multiple React components but here we're only returning a singular one.

louise-davies (changes requested):
Some small changes requested. Also, here's a GIF that may be useful to show to confirm this fits the description of the mockup:

cart-indicator

packages/datagateway-common/src/views/selectionAlert.component.tsx
joelvdavies: @louise-davies This is because the return cannot be null and the contents is conditional. I am not sure if there is a better way other than returning some other empty element.

joelvdavies: @louise-davies I managed to sort this out, it was because I had explicitly stated the function should return ReactElement.

PR #851, opened 2021-10-08
Use givenName and familyName fields for Users #817

### LINKED ISSUE #817: Use givenName and familyName fields for Users instead of fullName if they are not null
**Description**:
ISIS is starting to move to use givenName and familyName rather than the fullName field for Users. We need to support this.

Note: ensure this doesn't affect DLS (I presume they're just not filling those fields - otherwise we may need a settings toggle to say which field to use)

Note: is there any concerns about internationalisation & localisation regarding this? Or do we just format everything in the "English" format

**Acceptance criteria**:
- [ ] Use `givenName` and `familyName` fields if they exist
- [ ] Ensure this change only affects ISIS - if DLS fills those fields in then a setting will be needed to ensure both ways (`fullName` preferred vs `givenName` `familyName` preferred are possible)

joelvdavies:
## Description
Use 'givenName' and 'familyName' for a User on ISIS landing pages instead of 'fullName' when they are defined.

## Testing instructions
Still need to check with DLS before merging.

- [ ] Review code
- [ ] Check Actions build

## Agile board tracking
Closes #817


joelvdavies:
@louise-davies At the moment I have preserved the order, so the switch would be `Mr Joel Davies` to `Joel Davies`.

Issue #835, opened 2021-09-09
Embargoed study landing pages can't read information from investigations


louise-davies:
**Description**:
This is most likely an issue to do ICAT rules more than anything else, but this is a blocker on decommissioning the old DOI landing pages.

**Acceptance criteria**:
- [ ] Embargoed/unreleased studies display all relevant info on their landing pages

louise-davies:
@stuartpullinger highlighting this issue - is this something that can be changed in such a way to not affect the integrity of keeping data closed? i.e. a public step between study and investigation perhaps? As this functionality already exists on the DOI landing pages (i.e. embargoed investigations have their metadata displayed on study landing pages)

agbeltran:
We discussed with @louise-davies that this will be sorted once we have ```DataPublication``` added to the ICAT schema, but it would be good to have an interim solution.

MRichards99:
@louise-davies I've spent much of this afternoon trying to find a solution for this. I've been looking at the current query but trying to find a way to solve the missing investigations. I've been experimenting with some of the JPQL specific operators (`SIZE`, `IS EMPTY` etc.) but ultimately, because the investigation entity isn't a collection (not like `studyInvestigations` where they're presented as arrays since it's a one-to-many relationship) none of them work. 

Upon realising the above doesn't work, I've been trying to think about how to take your `/investigations` query further to remove the duplicates. Sadly ICAT Server doesn't support SELECT on multiple attributes which removes the simple solution of putting using a distinct filter on each of the fields we need. I've tried hacking where filters looking for `null` or `""` but that doesn't seem have materialised.

I think I've spent enough time on this to understand the problem so it might be time for a meeting to discuss further. The only other thought I had was to create another ISIS specific endpoint for this purpose but I'm not sure how it would work as I can't even find a solution in JPQL. Perhaps a seed for you to think about.

PR #888, opened 2021-10-27
Filter out embargoed studies from Study views

louise-davies:
## Description
We add a filter to the endDate to filter out investigations within the embargo period. We still need to double check with ISIS that this is the correct filter - although looking at the data this seems to be correct and works well.

## Testing instructions
- [ ] Review code
- [ ] Check Actions build
- [ ] Verify no blank study records visible anymore

## Agile board tracking
connect to #835 - doesn't close as it doesn't fix the problem with DOI landing pages for embargoed data


louise-davies:
Kevin has passed along some confirmation on what the specific query ISIS uses is and it is based on `releaseDate` and not `endDate`, so this PR will have to be changed to reflect that

louise-davies:
add open data label to this view

joelvdavies (approved):
Seems to work as I could not find any more blank results. Reviewed code and changes to coverage, CI passes.

PR #902, opened 2021-11-01
Fix datafile search with Lucene 4.10.0 (#890)

### LINKED ISSUE #890: Fix datafile search so it works on lucene 4.10.0
**Description**:
Currently, datagateway-search is only compatible with icat-lucene 4.11.1 where it fixes the timestamp datafile search issue. We can work around this issue and do as TopCAT does and not send `upper` and `lower` query parameters if the user hasn't specified a date between search

**Acceptance criteria**:
- [ ] Datafile search works on datafiles when run against a icat-lucene 4.10.0 server

joelvdavies:
## Description
Fix datafile search so it works with Lucene 4.10.0, by removing `upper` and `lower` from the query when no start/end date is defined.

## Testing instructions

- [ ] Review code
- [ ] Check Actions build

## Agile board tracking
Closes #890

### DIFF
--- a/packages/datagateway-common/src/api/lucene.tsx
+++ b/packages/datagateway-common/src/api/lucene.tsx
@@ -1,6 +1,6 @@
 import { MaterialUiPickersDate } from '@material-ui/pickers/typings/date';
 import axios, { AxiosError } from 'axios';
-import { format } from 'date-fns';
+import { format, set } from 'date-fns';
 import { useQuery, UseQueryResult } from 'react-query';
 import { useSelector } from 'react-redux';
 import { StateType } from '..';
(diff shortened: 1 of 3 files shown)


packages/datagateway-common/src/api/lucene.tsx
```diff
@@ -34,27 +34,35 @@ const urlParamsBuilder = (
...
-      : '00000-01-01';
-  const stringStartDateArray = stringStartDate.split('-');
-  query.lower =
-    stringStartDateArray[0] +
-    stringStartDateArray[1] +
-    stringStartDateArray[2] +
-    '0000';
+  //Do not assign query.lower for datafiles when there is no start date
+  //as this fixes issue when using lucene 4.10.0 instead of 4.11.1
+  if (datasearchtype !== 'Datafile' || params.startDate !== null) {
```
louise-davies, line 39: Is it not possible to just have some logic that if both `startDate` and `endDate` are null then to not include them in the query? That feels more appropriate and would be something we wouldn't need to remove once we upgrade to 4.11.1

i.e. I feel just wrapping the original code in a `if (params.startDate !== null || params.endDate !== null)` would work and be future proof

louise-davies (changes requested):
This solution would mean we can't use datafile date searches on fixed versions of lucene

louise-davies (approved):
Tested against both lucene 4.10 and 4.11 and works as expected on both

PR #917, opened 2021-11-10
#913 - Remove uneccessary includes and type filters

### LINKED ISSUE #913: Remove uneccessary "includes" from views
**Description**:
Some views, e.g. the Dataset views, have include filters added to their requests even though they do not require the object they're including. This is a holdover from the DB backend days where you needed to include an item that you were filtering on. Removing these includes will increase performance.

**Acceptance criteria**:
- [ ] Only `include` filters that are necessary (i.e. add data that the user is going to see in that view) remain in the code

louise-davies:
## Description
The dataset views all had an include for investigation which I believe was a holdover from the DB backend days. This makes the queries less performant to the point I couldn't fetch 50 datasets on DLS preprod.

Also, the type filters on DLS card views was meaningless and so I removed it as well.

## Testing instructions
- [ ] Review code
- [ ] Check Actions build

## Agile board tracking
Closes #913

### DIFF
--- a/packages/datagateway-dataview/cypress/integration/card/dls/visits.spec.ts
+++ b/packages/datagateway-dataview/cypress/integration/card/dls/visits.spec.ts
@@ -65,17 +65,6 @@ describe('DLS - Visits Cards', () => {
   });
 
   it('should be able to filter by multiple fields', () => {
-    cy.contains('[role="button"]', 'Type ID').click();
-    cy.contains('[role="button"]', 'Type ID')
-      .parent()
-      .contains('[role="button"]', '2')
-      .click()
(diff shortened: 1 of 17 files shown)


joshuadkitenge (changes requested):
All the functionality still works as expected. CI test have passed

- Does the "include" also need to be removed in the isiStudiesCardView ??? 

<br><br>
Screenshot 2021-11-11 092437

<br><br>
Screenshot 2021-11-11 092349
.

louise-davies:
@joshuadkitenge No, the `include` there is needed as you can see in the study view we are displaying properties from the `investigation`, so that means we need the server to "include" the info so we can display it.

joshuadkitenge (approved):

PR #919, opened 2021-11-10
#914 - use selectAllSetting to stop sending allID requests

### LINKED ISSUE #914: If selectAllSetting is false - then we shouldn't fetch all IDs in table views
**Description**:
In table views where select all can be enabled an array of all IDs on the server is fetched. Because this can be costly, the intent was to disable this with the `selectAllSetting`. However, this just doesn't display the select all checkbox, and the fetch IDs request is still sent. I can imagine this being extremely harmful to DLS data if there are millions of datafiles in a dataset for example.

**Acceptance criteria**:
- [ ] `selectAllSetting` disables both the all ID request being sent and the visual of the select all checkbox

louise-davies:
## Description

We now no longer send `allId` requests if `selectAll` has been disabled via `selectAllSetting`. This improves performance. Additionally, I had to add `selectAllSetting` to `datagateway-search`, and I also did some further modifications to search to ensure DLS views don't have investigation download avaialble nor direct download.

## Testing instructions
- [x] Review code
- [ ] Check Actions build

## Agile board tracking
Closes #914

### DIFF
--- a/packages/datagateway-common/src/api/index.test.tsx
+++ b/packages/datagateway-common/src/api/index.test.tsx
@@ -509,6 +509,18 @@ describe('generic api functions', () => {
       expect(result.current.data).toEqual([1, 2, 3]);
     });
 
+    it('does not send axios request to fetch ids when set to disabled', async () => {
+      const { result } = renderHook(
+        () => useIds('investigation', undefined, false),
+        {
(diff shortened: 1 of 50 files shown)


joelvdavies (approved):
Reviewed changes, checked setting works as described in dataview & search. Tests pass.

PR #935, opened 2021-11-18
Fix incorrect size & counts after sorting and filtering #927

### LINKED ISSUE #927: Sorting & filtering can cause the size/fileCount values to be inaccurate
**Description**:
When sorting or filtering, the values in the size/fileCount column can get mixed up, causing the wrong values to be displayed to the user.

**Acceptance criteria**:
- [ ] size/file counts are always consistent

louise-davies:
## Description
The problem was with the batching mechanism, and it's fixed by resetting the "amount of successful requests" back to zero when the data changes using a useState hook.

I also changed the investigationTable component to use sizes rather than counts, as this means I'm able to e2e test this fix.

## Testing instructions
- [ ] Review code
- [ ] Check Actions build

## Agile board tracking
Closes #927


joshuadkitenge (changes requested):
All the CI Test have passed. All the functionality works as expected.

- I think we should have some testing (e2e tests using 2004 cycle 1, We might need more data on prep- prod as only 2004 cycle 1 has more than 1 row) in one of the table views to these use cases, just encase these issues pop up in the future.

louise-davies:
@joshuadkitenge I did add a line testing this to the generic investigations table e2e test as that view has sufficient data and uses sizes. Did you mean that we should add more of these types of tests?

joshuadkitenge (approved):
Functionality works as expected. The CI tests have all passed.

joshuadkitenge (approved):
Functionality works as expected. CI test have all passed.

PR #938, opened 2021-11-19
Default sorting in table and card views #931

### LINKED ISSUE #931: Certain views should have sorts applied by default
**Description**:
We received some feedback that it would be nice for certain views to have a default sort applied E.g. instrument view sorted alphabetically. Another example not given in the feedback is facility cycles sorted by date. TopCAT I believe has some of these defined so we should probably copy what is already configured.

**Acceptance criteria**:
- [ ] Replicate TopCAT default sorting params

joelvdavies:
## Description
Adds sorting by default on table and card views by adding a new parameter `defaultSort` to the parameters taken by table and card components. I have also renamed the method `usePushSort` to `useSort` to resemble a previous change to `usePushView` which allows the sort parameters to be replaced in the URL to ensure back functionality works as expected. The default sorting behaviour matches that of Topcat in ISIS and DLS data views.

## Testing instructions
This is a draft as I still need to go through the e2e tests.

- [x] Review code
- [x] Check Actions build

## Agile board tracking
Closes #931

### DIFF
--- a/packages/datagateway-common/src/card/cardView.component.tsx
+++ b/packages/datagateway-common/src/card/cardView.component.tsx
@@ -21,7 +21,14 @@ import { createStyles, makeStyles, Theme } from '@material-ui/core/styles';
 import ExpandMoreIcon from '@material-ui/icons/ExpandMore';
 import { Pagination } from '@material-ui/lab';
 import ArrowTooltip from '../arrowtooltip.component';
-import { Entity, Filter, Order, SortType, FiltersType } from '../app.types';
+import {
+  Entity,
+  Filter,
+  Order,
+  SortType,
+  FiltersType,
+  UpdateMethod,
+} from '../app.types';
 import React from 'react';
 import { useTranslation } from 'react-i18next';
 import AdvancedFilter from './advancedFilter.component';
@@ -95,7 +102,11 @@ export interface CardViewProps {
   onPageChange: (page: number) => void;
   onFilter: (filter: string, data: Filter | null) => void;
   onResultsChange: (page: number) => void;
-  onSort: (sort: string, order: Order | null) => void;
+  onSort: (
+    sort: string,
+    order: Order | null,
+    defaultSort: UpdateMethod
+  ) => void;
 
   // Props to get title, description of the card
   // represented by data.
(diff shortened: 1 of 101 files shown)


packages/datagateway-common/src/card/cardView.component.tsx
```diff
@@ -95,7 +103,11 @@ export interface CardViewProps {
   onPageChange: (page: number) => void;
   onFilter: (filter: string, data: Filter | null) => void;
   onResultsChange: (page: number) => void;
-  onSort: (sort: string, order: Order | null) => void;
+  onSort: (
+    sort: string,
+    order: Order | null,
+    defaultSort: UpdateMethod
```
louise-davies, line 109: this variable should be renamed to "updateMethod" or something because it currently doesn't make sense.

packages/datagateway-common/src/table/table.component.tsx
```diff
@@ -96,7 +97,11 @@ interface VirtualizedTableProps {
   loadMoreRows?: (offsetParams: IndexRange) => Promise<unknown>;
   totalRowCount?: number;
   sort: { [column: string]: Order };
-  onSort: (column: string, order: Order | null) => void;
+  onSort: (
+    column: string,
+    order: Order | null,
+    defaultSort: UpdateMethod
```
louise-davies, line 103: Again, defaultSort sounds like the name of a bool - and not the strings "push" or "replace"

joelvdavies: Oops, yes I think I must have copy and pasted, it should have matched the `onSort` in index.tsx. Thanks for spotting that 😄

louise-davies (changes requested):
The facility cycle view has the wrong default sort - in topcat it's ordered by startDate desc, name asc. I think getting the ordering on default sorts is difficult, so I think the easiest solution is to remove the default sort by name in the facility cycle view so that the startDate desc can take precedence.

Additionally, the DLS proposal view in topcat has no default sort, but I think the default specified here of title asc is fine I think we just leave that one.

louise-davies (approved):
Default sort works well and the facility cycle issue I noticed has been fixed

PR #946, opened 2021-11-24
Add description and type to ISIS instruments table (#872)

### LINKED ISSUE #872: Browse main page
**Description**:
The current main page when browsing (for ISIS) shows the list of instrument names without any other information/column. It seems we could use better use of the space showing extra information about the instrument in separate columns, allowing also for other ways of filtering the instruments. I would consider description (even though the text could be too lengthy) and certainly type (allowing to filter by type). 

**Acceptance criteria**:
dg-instruments

- [ ] Extend the first table in Browse with more information about the instruments

joelvdavies:
## Description
Adds type to ISIS instruments table as requested in #872.

image

## Testing instructions

- [x] Review code
- [x] Check Actions build

## Agile board tracking
Closes #872


louise-davies:
I would run this by ISIS first before merging this in as this might be information that isn't needed by users

louise-davies (changes requested):
Can we just add the type column and not the description column?

louise-davies (approved):

PR #976, opened 2021-12-02
Citation formatter on ISIS study landing pages #868

### LINKED ISSUE #868: Citation formatter
**Description**:
Currently on the landing pages, we have a recommended citation format and have a 'copy to clipboard' option.

It would be great to add a citation formatter relying on the API by datacite:
https://support.datacite.org/v1.1/docs/datacite-citation-formatter

that should work as in 
https://citation.crosscite.org/

so, I suggest that - in addition to the current recommended citation format - we add an option to choose the formatting style (a dropdown list of styles) and show the citation with a copy to clipboard option too

**Acceptance criteria**:
- [ ] DOI landing pages give the option to copy the citation in different formats

joelvdavies:
## Description
Adds a citation formatter to the ISIS study landing page when a DOI is defined. A drop down menu can then be used to select a format to use and then displays it with a 'Copy citation' button to copy the citation. The allowed formats can be configured using the language file. The possible formats can be found here: https://github.com/citation-style-language/styles.

This also slightly modifies isisStudyLanding to use better aria-labels and the language files and adds .component to the advancedHelpDialogue component file in search.

image
image
image

## Testing instructions

- [x] Review code
- [x] Check Actions build

## Agile board tracking
Closes #868


joelvdavies:
After speaking to Louise, I have updated this to replace the old citation using the `chicago-author-date` format as a default which is fetched on page load. This can also be configured through `default.json` as the default is taken as the first defined in the list of formats. Here are some updated screenshots:
image
image

louise-davies (commented):
This looks good! The only concern I have is we're basically ignoring ISIS's recommendation here: https://www.isis.stfc.ac.uk/Pages/Digital-Object-Identifiers-(DOIs)-for-ISIS-Data.aspx. I can't seem to find any format that matches this - but perhaps we can have a "default" dropdown option which doesn't fetch via the citation formatter API and instead does similar to what the old one does: https://github.com/ral-facilities/common-doi/blob/0d603ca1311df1f287fbd6af742ecf9e78ee6ffc/src/main/java/uk/ac/stfc/doi/model/DOIModel.java#L242

Or maybe I'm overthinking this - it just seems wrong to not offer ISIS's preferred format.

louise-davies (changes requested):
I believe we also need the formatter on the investigation landing pages to match with what the current landing pages offer: https://data.isis.stfc.ac.uk/doi/INVESTIGATION/85007962/

joelvdavies:
I have also just modified it to display the same as before in cases where there is no DOI e.g.
image
as this matches TopCAT.

packages/datagateway-dataview/src/views/landing/isis/isisInvestigationLanding.component.tsx
```diff
@@ -394,67 +391,16 @@ const LandingPage = (props: LandingPageProps): React.ReactElement => {
...
-                Copied citation
-              </Button>
-            )}
+            <CitationFormatter
+              doi={doi}
+              formattedUsers={formattedUsers}
+              title={title}
+              startDate={
+                studyInvestigation
+                  ? studyInvestigation[0]?.study?.startDate
```
louise-davies, line 400: The study very often does not have a start date itself and instead it gets the start date from the underlying investigation. So here this should just be the investigation startDate

louise-davies (changes requested):

louise-davies (approved):

PR #1014, opened 2021-12-15
Bugfix/switch formatbytes function to calculate in decimal #993

### LINKED ISSUE #993: Switch formatBytes function to calculate in decimal
**Description**:
The `formatBytes` method technically calculates Kibi, Mibi etc. bytes and not Kilo, Mega etc. bytes. This was a config option in TopCAT but both ISIS and DLS use the Kilo/Mega calculation (i.e. 1000 bytes in a kB, not 1024 bytes in a kB). This means we're displaying different sizes to the user than TopCAT did. I don't think it's worth creating a config option for this like TopCAT had if both facilities are using decimal bytes.

**Acceptance criteria**:
- [x] Switch `formatBytes` to calculate using 1000 not 1024

AkhilMag13:
## Description
Changed the formatBytes function to use 1000 instead of 1024 when calculating the size of data.

## Testing instructions

- [ ] Review code
- [ ] Check Actions build

## Agile board tracking
Closes #993

### DIFF
--- a/packages/datagateway-common/src/table/cellRenderers/cellContentRenderers.test.tsx
+++ b/packages/datagateway-common/src/table/cellRenderers/cellContentRenderers.test.tsx
@@ -8,7 +8,6 @@ import {
   investigationLink,
   tableLink,
   formatCountOrSize,
-  getStudyInfoInvestigation,
 } from './cellContentRenderers';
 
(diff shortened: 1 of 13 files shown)


joshuadkitenge (approved):
Functionality works as expected. CI tests have passed

PR #1033, opened 2021-12-20
Change exclude filters to use `nilike` operator to API

MRichards99:
## Description
This is a very small PR to change the not like (`nlike`) to an insensitive not like (`nilike`). The difference in these operators is only noticeable in production where OracleDB is used, but that's been tested in the past with DataGateway API.

## Testing instructions
When using an exclude filter, check that the request sent to DataGateway API contains the `nilike` operator, and not the `nlike` operator.

- [ ] Review code
- [ ] Check Actions build

## Agile board tracking
connect to #{issue number}


AkhilMag13 (approved):
Works as expected - uses the `nilike` operator when sending the request to DataGateway API.
image

Issue #1008, opened 2021-12-14
Filters applied when switching to another search tab


agbeltran:
**Description**:
After a search is performend, and a filter applied to one of the results tabs, the same filter is applied when switching to another tab.

Steps to reproduce:
1. go to search and include 'battery' as a query string, hit search
2. in the investigation tab, in the Title filter, select Exclude and filter excluding 'structure'
3. then go to the Dataset tab
4. 400 error - bad request - is raised

**Acceptance criteria**:
- [ ] Fix filters between search results tab, consider what happens if user gets back to original search tab

louise-davies:
separate the filter params into "investigation"/"dataset"/"datafile" filter params perhaps?

PR #1034, opened 2021-12-20
Bugfix/update count when filtering in search tabs #1006

### LINKED ISSUE #1006: Update count when filtering in search tabs
**Description**:
After a search, the results table is not updating the count of results when filtering columns.

Steps to test:
1. go to search and search for 'battery'
2. in the investigation tab (with 239 results), filter by title to include 'structure'
3. it is clear that the results chage, but the count (239) is not updated.

This should be fix for filtering in each of the tabs.

Screenshot 2021-12-14 at 11 29 50

**Acceptance criteria**:
- [x] Update count of results in each search tab after filtering.

AkhilMag13:
## Description
Added a progress bar and additional code to update the count of the results when using filters in the search tabs.

## Testing instructions
Once opened, press the search button to get all the results in the search and use the filters available in the table column for advanced searching. Once the filter is used, a progress bar should appear while the results are being filtered and the results count on the tab labels should change to reflect the number of filtered results.

- [ ] Review code
- [ ] Check Actions build

## Agile board tracking
Closes #1006


MRichards99 (commented):
I'm happy with the changes, they work as expected. I haven't experienced the slow loading on investigations, but that might be because I only have a small amount of data on my ICAT instance. 

These changes aren't present on card view of DataGateway Search. Are they supposed to be?

AkhilMag13:
> I'm happy with the changes, they work as expected. I haven't experienced the slow loading on investigations, but that might be because I only have a small amount of data on my ICAT instance.
> 
> These changes aren't present on card view of DataGateway Search. Are they supposed to be?

Yes, it should be in cardview too, I'll add that in 👍

MRichards99 (approved):
Looks good on card view. When #1008 is fixed/reviewed, we should check that the other tab counts don't go to 0 (your comment [here](https://github.com/ral-facilities/datagateway/issues/1006#issuecomment-997961147)) but the feature looks good :)

PR #1041, opened 2022-01-04
Change Python to use 3.6 as workaround for import issue

MRichards99:
## Description
A change to use Python 3.6 as this _seemed_ to work on the API's CI. 
## Testing instructions
See if this resolves the import error as reported on Slack.

- [ ] Review code
- [ ] Check Actions build


joelvdavies (approved):
CI passes.

PR #1042, opened 2022-01-05
Improved table padding #1022

### LINKED ISSUE #1022: Padding for facilities cycle table
**Description**:
The facilities cycle column has no padding to the left, and it looks at the border of the window/screen. We should add some padding for that column. This doesn't happen in other tables as there are selection options/etc.

Screenshot 2021-12-16 at 15 03 31

**Acceptance criteria**:
- [ ] Fix padding on facilites table

joelvdavies:
## Description
Improves table padding so that the first column in a table also obtains padding before any text when there is no checkbox/detail panel to display.

image
image
image
image

## Testing instructions
This change will effect all tables in dataview/search/download.

- [ ] Review code
- [ ] Check Actions build

## Agile board tracking
Closes #1022


AkhilMag13 (approved):
Works as described on all the plugins 👍

PR #1043, opened 2022-01-05
Filters applied when switching to another search tab #1008

### LINKED ISSUE #1008: Filters applied when switching to another search tab

joshuadkitenge:
## Description
- When switching from one tab to another, the advanced filters carried over to the next tab, causing a 400 error (bad request). <br><br> 

search 400 error

- When switching from one tab to another, the page number is carried over to the next page, causing the tab to be on the page the previous tab was on (card view). 
<br><br>

search page error

- DOI for the experiment doesn't have a hover tool tip on the Investigation Landing page 
<br><br>
hover tool tip study

- [x] use specific "investigationFilters", "datasetFilters" and "datafileFilters" for query for the respective tabs.
- [x] use specific "investigationPage" and "dataPage" for query for the respective tabs in card view. 
- [x] add ArrowTooltip to the investigation landing page (Dataset) and study landing page (investigation)

## Testing instructions

- [ ] Review code
- [ ] Check Actions build
- [ ] {more steps here}

## Agile board tracking
closes #1008


joshuadkitenge:
- Search filters<br><br>
search filters
 <br><br>
- Search Page <br><br>
search pages
<br><br>
- Arrowtooltip <br><br>
tooltip on study landing pages

louise-davies (changes requested):
Switching tabs and then applying a new filter gets rid of the old filters for other tabs. This means going back to a tab loses it's filter. If that's the functionality - we might as well just clear filters on tab change and not split the three entity types into three separate filters.

Also - sorting doesn't work properly either. If you have a filter set and you sort it deletes the filter, and if you switch tabs with a sort active it can cause the same issue as the original issue.

joshuadkitenge:
Created a fresh PR to work on the comments.  https://github.com/ral-facilities/datagateway/pull/1069

PR #1058, opened 2022-01-12
Disable sorting in DLS proposal views #1049

### LINKED ISSUE #1049: Disable sorting in DLS proposal views
**Description**:
Matt and I realised that sorting on these columns is highly likely to return incorrect results.

Quotes from slack:

Louise: 
> I'm not sure limit is being applied correctly to distinct filters, as I can get the behaviour where card view sees less data than table and the only difference in the query is the number of items requested i.e the limit (see gif)

Matt: 
> I've been playing around with the Diamond issue in Postman and basically I think it's a quirk with combining distinct and orders in a single query.

> With the request that DataGateway sends, if you set limit to 11, that fixes the problem. If you remove the distinct and limit filters, the first NR23571 appears as the 11th result. When the limit is set to 10, the database appears to fetch the first 10 results, then do the distinct on those 10 rather than the entire result set.

> However, if you remove the order filters from the API query, the problem resolves itself. You can set the limit to as low as 2 and you'll still see both results.

> I think this is explained by the second bullet point of documentation I found online: ```If you combine LIMIT row_count with ORDER BY, MySQL stops sorting as soon as it has found the first row_count rows of the sorted result, rather than sorting the entire result.
One manifestation of this behavior is that an ORDER BY query with and without LIMIT may return rows in different order, as described later in this section.```

> So when the query has a LIMIT with just a DISTINCT (no orders), "MySQL stops as soon as it finds row_count unique rows". I'm guessing that with limit, order and distinct all in one go, it stops sorting after the first x rows then applies the distinct behaviour on that.

> While the problem is caused by having order filters, I guess the solution would be to increase the limit and accept that this bug could appear again where there's more data? Because the alternative is to remove order filters but then users won't be able to sort on data

Louise: 
> maybe we just have to disable sort then

I also believe this odd behaviour also applies to TopCAT based on my testing with DLS preprod, but I don't think we should worry. We might just have to explain the problem to users at some point as to why we've removed the sorting for this view.

**Acceptance criteria**:
- [ ] Sorting on columns in the DLS proposals table and card views is disabled

joelvdavies:
## Description
Disables sorting in DLS proposal table and card views. Also modifies `cardView` to display cards in the centre when there are no sorting or filtering options to display on the side.

image

## Testing instructions
Since 'cardView' is used in dataview and search both should be checked.

- [ ] Review code
- [ ] Check Actions build

## Agile board tracking
Closes #1049


packages/datagateway-dataview/cypress/integration/card/dls/proposals.spec.ts
```diff
@@ -57,68 +53,7 @@ describe('DLS - Proposals Cards', () => {
       .should('exist');
   });
 
-  describe('should be able to sort by', () => {
```
MRichards99, line 60: @louise-davies Do we intend to re-enable sorting one day? If so, would it be useful to keep these tests and skip over them for the time being?

MRichards99 (commented):
I'm happy with the changes, they work but I thought I should query (as per review comment) if there's any point in keeping the tests around or not. Happy to approve if we decide that we don't want the tests any more :)

packages/datagateway-dataview/cypress/integration/card/dls/proposals.spec.ts
MRichards99: @joelvdavies has just told me @louise-davies agreed that the tests should be removed. Like Joel I just thought I'd check :) will resolve this and approve the PR

MRichards99 (approved):

packages/datagateway-dataview/cypress/integration/card/dls/proposals.spec.ts
louise-davies: @MRichards99 I thought that sorting is pretty much impossible due to the incompatibility of limit and order on distinct queries? Hence we would never be able to re-enable them

MRichards99: @louise-davies I agree, I don't think it would be a realistic objective to get them re-enabled but I didn't know if there would be anything changing in the medium-long term to allow sorting to be fixed.

PR #1067, opened 2022-01-14
Fix bug in ResizeObserver code

louise-davies:
## Description
Joel noticed that search could error when run standalone from scigateway. This was due to the fact that the ResizeObserver code assumed the search box is always mounted (which is the case in scigateway) - but when ran locally we load the blank page which throws an error. I fixed it by using the technique described here: ``ttps://reactjs.org/docs/hooks-faq.html#how-can-i-measure-a-dom-node` and adapting it to work for a ResizeObserver

## Testing instructions
- [x] Review code
- [ ] Check Actions build
- [x] No errors when running search locally, and the table gets resized properly still on different window widths


joelvdavies (approved):
Reviewed changes. Tests pass. Issue fixed.

louise-davies:
Ignoring patch coverage as we can't test the ResizeObserver code itself in Jest

PR #1069, opened 2022-01-18
Filters applied when switching to another search tab (take 2) #1008

### LINKED ISSUE #1008: Filters applied when switching to another search tab

joshuadkitenge:
## Description

- When switching from one tab to another, the advanced filters carried over to the next tab, causing a 400 error (bad request). <br><br> 

search 400 error

- When switching from one tab to another, the page number is carried over to the next page, causing the tab to be on the page the previous tab was on (card view). 
<br><br>

search page error

- DOI for the experiment doesn't have a hover tool tip on the Investigation and study landing pages 
<br><br>
hover tool tip study

- The non-current tab numbers are being updated when filtering on the current tab.    
## Testing instructions

- [x] Review code
- [x] Check Actions build
- [ ] {more steps here}

## Agile board tracking
closes #1008


joshuadkitenge:
E2E test for cardview are failing as the test appears to get a different number of results locally compared to the automated tests on github meaning the test fails

joshuadkitenge:
Screenshot 2022-01-25 130428

The functionality is working, but it is still giving an error. I'm currently investigating why this is happening

louise-davies:
@joshuadkitenge give me a ping once you've finished investigating and it's ready to review or if you need some help

packages/datagateway-search/src/searchPageCardView.component.tsx
```diff
@@ -385,14 +404,8 @@ const mapStateToProps = (state: StateType): SearchCardViewStoreProps => {
     datasetTab: state.dgsearch.tabs.datasetTab,
     datafileTab: state.dgsearch.tabs.datafileTab,
     investigationTab: state.dgsearch.tabs.investigationTab,
-    currentTab: state.dgsearch.tabs.currentTab,
+    searchableEntities: state.dgsearch.searchableEntities,
```
louise-davies, line 407: This prop isn't being used right? It should probably be removed then

packages/datagateway-search/src/searchPageContainer.component.tsx
```diff
@@ -444,11 +548,15 @@ const SearchPageContainer: React.FC<SearchPageContainerCombinedProps> = (
                       <SearchPageCardView
                         containerHeight={containerHeight}
                         hierarchy={match.params.hierarchy}
+                        onCurrentTab={pushCurrentTab}
```
louise-davies, line 551: `onCurrentTab` is a little confusing a name for me - could this be `onCurrentTabChange` or `onTabChange` to indicate that this prop is the function that should be called when tabs change?

packages/datagateway-search/src/searchPageTable.component.tsx
```diff
@@ -408,16 +414,10 @@ const mapStateToProps = (state: StateType): SearchTableStoreProps => {
   return {
     maxNumResults: state.dgsearch.maxNumResults,
     datasetTab: state.dgsearch.tabs.datasetTab,
+    searchableEntities: state.dgsearch.searchableEntities,
```
louise-davies, line 417: Again, I don't think this is being used

packages/datagateway-search/src/searchPageContainer.component.tsx
```diff
@@ -43,6 +46,79 @@ import ViewListIcon from '@material-ui/icons/ViewList';
...
+};
+
+export const storeResults = (
+  results: number,
+  searchableEntities: string
+): void => {
+  const resultsNumber = (searchableEntities as string) + 'Results';
+
+  localStorage.setItem(resultsNumber, JSON.stringify(results));
+};
```
louise-davies, line 80: Minor optimisation - a lot of the time these will end up storing the default value in `localStorage` - and thus cause us to update the query params later on as the result from localStorage is not null. Perhaps some guards to prevent this will optimise updates a little?

i.e.
```
if (Object.keys(filters).length !== 0)
   localStorage.setItem(filter, JSON.stringify(filters));
```
and
```
if (page !== 1)
   localStorage.setItem(filter, JSON.stringify(page));
```

joshuadkitenge:
image

bad request 400 error of correct links

joshuadkitenge:
search 400 error when filtering on a tabs that is not investigationsfhdg

louise-davies (approved):
Looks good and I can't recreate any bugs. As for the requests occasionally failing - it's only occurred once on my machine and it seems like it's a problem on `datagateway-api` end (it adds order by ID twice to the request accidentally) - if we see this problem out in the wild in production we should raise the issue on the `datagateway-api` repo

joshuadkitenge:
search tabs fix

PR #1079, opened 2022-01-24
Use ResizeObserver to calculate whether a tooltip is visible (#1077)

### LINKED ISSUE #1077: Maximising table views allows tooltips to be hoverable
**Description**:
This is similar to https://github.com/ral-facilities/datagateway/issues/878. This is present when double clicking to maximise on firefox and chrome or using the maximise button on chrome.
tooltipIssue

Louise discovered that this is likely caused by the window resize event being fired before the window has finished resizing meaning the width values used by the tooltip are old.

A delay of 200ms after resizing is enough to prevent this although the better option is to use the ResizeObserver as in https://github.com/ral-facilities/datagateway/pull/1063.

**Acceptance criteria**:
- [ ] Issue fixed

joelvdavies:
## Description
- Modifies tooltips to use ResizeObserver to calculate whether a tooltip is visible which then fixes #1077.
- Makes small change to homepage margins to ensure a scroll bar isn't needed at 1080p.
image
- Changes threshold for which cards on the homepage go to separate lines as before the text would overrun before doing so
- Also fixes issue with tooltips not displaying properly for long card titles

## Testing instructions

- [x] Review code
- [x] Check Actions build

## Agile board tracking
Closes #1077


packages/datagateway-common/src/arrowtooltip.component.tsx
```diff
@@ -28,96 +29,85 @@ export const getTooltipText = (node: React.ReactNode): string => {
 const ArrowTooltip = (
   props: TooltipProps & {
     percentageWidth?: number;
```
louise-davies, line 31: We don't use this prop any more do we? Can this be removed if so?

joelvdavies: @louise-davies It is currently used in breadcrumbs and it didn't work without it, not sure about after the refactor though.

louise-davies: @joelvdavies Ah, I didn't notice that. My refactor doesn't affect the breadcrumbs themselves (only the code which renders the individual breadcrumbs) so I'd anticipate the same result but I'll double check

louise-davies (approved):
Works well and fixes the bug occurring when maximising windows

I'll merge this to get around the patch coverage failing (due to difficulties in testing resizeObserver in unit tests)

PR #1085, opened 2022-01-25
Feature/breadcrumb landing page links #1056

### LINKED ISSUE #1056: Add link in breadcrumb for investigation landing page
**Description**:
When looking at a dataset, the breadcrumb for the investigation is not clickable and the only way to go back to the investigation information is to get all the data from the cycle, and search through

https://datagateway.isis.stfc.ac.uk/browse/instrument/9/facilityCycle/84793031/investigation/87863101/dataset?sort=%7B%22createTime%22%3A%22desc%22%7D

We should be linking the breadcrumb to the investigation landing page.

**Acceptance criteria**:
- [ ] add link to breadbrumb for the investigation

louise-davies:
## Description
This PR refactors the breadcrumb component to be a functional component, which allows for us to use `react-query` to fetch the data and thus eliminate the need for the `breadcrumbHierarchy` which acted as a cache. This simplifies the code quite a lot as a result (in my opinion - it's at least fewer LOC!)

Then, I added some code to add a link to the landing page when looking at the view below it (i.e. when looking at an investigation's datasets, clicking on the investigation name takes you to the landing page). I didn't add any extra breadcrumbs though - so if a user is further down the hierarchy they'll need to click the e.g. investigation name twice to get back to it's landing page. Perhaps we want to make separate breadcrumbs for the landing pages themselves? Or is the current solution fine?

## Testing instructions
- [ ] Review code
- [ ] Check Actions build

## Agile board tracking
Closes #1056


joelvdavies (commented):
@louise-davies I personally don't mind the current solution having to click twice, at least you can navigate backwards everywhere now. The only thing I noticed is that previously clicking an investigation/study title always took you back to the landing page except in the case Alejandra mentioned so the only other solution I could think of without adding more links is to keep this behaviour and expect the user to click back onto view investigations/datasets to go back to the list.

louise-davies:
@joelvdavies it seems the behaviour you are describing was a bug - the breadcrumbs were coded before landing pages were a thing and I don't think their code was updated to account for them, as I get different behaviour depending on how I've navigated to a page:

breadcrumbs-bug-1

breadcrumbs-bug-2

I believe my code has fixed this "bug" i.e. the behaviour is no longer inconsistent - whether we wanted the landing page first vs dataview first link is a separate issue

joelvdavies (dismissed):
Works well and is consistent. Reviewed code and coverage, CI passes.

joelvdavies (approved):
Works well and is consistent. Reviewed code and coverage, CI passes.

PR #1086, opened 2022-01-25
Sort DLS proposal views by title rather than ID #1084

### LINKED ISSUE #1084: Sort DLS proposal views by title or name rather than by ID
**Description**:
Normally, we sort by ID on every request to ensure the database returns results in a deterministic order so the data can be paginated, but sorting on a field not returned by distinct causes missing results. @MRichards99 found that we can fix both problems by instead sorting on one of the two `distinct`ed fields (i.e. `title` and `name`). I'd say `title` looks nicer to sort on in DLS data but either is fine.

This also means theoretically we could reintroduce filters on these views - but they would work quite weirdly as you *need* to sort by at least one of the two fields, so you couldn't have both be null. This may be more effort to implement than it's worth, and perhaps should be a different issue.

**Acceptance criteria**:
- [ ] DLS proposal views are sorted by `title` or `name` rather than `id`

joelvdavies:
## Description
This removes the default sorting by ID and replaces it with a default sort on titles for DLS proposal views as in https://github.com/ral-facilities/datagateway/issues/1084.

## Testing instructions

- [ ] Review code
- [ ] Check Actions build

## Agile board tracking
Closes #1084

### DIFF
--- a/packages/datagateway-dataview/src/views/card/dls/dlsProposalsCardView.component.tsx
+++ b/packages/datagateway-dataview/src/views/card/dls/dlsProposalsCardView.component.tsx
@@ -40,12 +40,17 @@ const DLSProposalsCardView = (): React.ReactElement => {
       filterValue: JSON.stringify(['name', 'title']),
     },
   ]);
-  const { isLoading: dataLoading, data } = useInvestigationsPaginated([
-    {
-      filterType: 'distinct',
-      filterValue: JSON.stringify(['name', 'title']),
-    },
-  ]);
+  const { isLoading: dataLoading, data } = useInvestigationsPaginated(
+    [
+      {
+        filterType: 'distinct',
+        filterValue: JSON.stringify(['name', 'title']),
+      },
+    ],
+    //Do not add order by id as id is not a distinct field above and will otherwise
+    //cause missing results
+    true
+  );
 
   const title: CardViewDetails = React.useMemo(
     () => ({
@@ -59,6 +64,9 @@ const DLSProposalsCardView = (): React.ReactElement => {
           'dls-proposal-card-title'
         ),
       filterComponent: textFilter,
+      //Sort to ensure a deterministic order for pagination (ignoreIDSort: true is
+      //used above in useInvestigationsPaginated)
+      defaultSort: 'asc',
       disableSort: true,
     }),
     [t, textFilter, view]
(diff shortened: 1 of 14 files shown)


packages/datagateway-dataview/src/views/card/dls/dlsProposalsCardView.component.tsx
```diff
@@ -40,12 +40,17 @@ const DLSProposalsCardView = (): React.ReactElement => {
...
-  ]);
+  const { isLoading: dataLoading, data } = useInvestigationsPaginated(
+    [
+      {
+        filterType: 'distinct',
+        filterValue: JSON.stringify(['name', 'title']),
+      },
+    ],
+    //Do not add order by id as id is not a distinct field above and will otherwise
+    //cause missing results
```
MRichards99, line 51: Could you add spaces between `//` and the comment please?

packages/datagateway-dataview/src/views/card/dls/dlsProposalsCardView.component.tsx
```diff
@@ -59,6 +64,9 @@ const DLSProposalsCardView = (): React.ReactElement => {
           'dls-proposal-card-title'
         ),
       filterComponent: textFilter,
+      //Sort to ensure a deterministic order for pagination (ignoreIDSort: true is
+      //used above in useInvestigationsPaginated)
```
MRichards99, line 68: Same spacing comment here too

packages/datagateway-dataview/src/views/table/dls/dlsProposalsTable.component.tsx
```diff
@@ -30,12 +30,17 @@ const DLSProposalsTable = (): React.ReactElement => {
...
-  ]);
+  const { fetchNextPage, data } = useInvestigationsInfinite(
+    [
+      {
+        filterType: 'distinct',
+        filterValue: JSON.stringify(['name', 'title']),
+      },
+    ],
+    //Do not add order by id as id is not a distinct field above and will otherwise
+    //cause missing results
```
MRichards99, line 41: Same spacing comment here too

packages/datagateway-dataview/src/views/table/dls/dlsProposalsTable.component.tsx
```diff
@@ -66,6 +71,9 @@ const DLSProposalsTable = (): React.ReactElement => {
           );
         },
         filterComponent: textFilter,
+        //Sort to ensure a deterministic order for pagination (ignoreIDSort: true is
+        //used above in useInvestigationsInfinite)
```
MRichards99, line 75: Same spacing comment here too

MRichards99 (changes requested):
I've tested this locally by using DLS preprod URLs and logging in as a test user and this fixes the issues seen on DLS preprod. Could you address the spacing comments please, just for consistency?

MRichards99 (approved):

PR #1109, opened 2022-02-10
Fixing failing unit and E2E tests for clear filters pr #1019

### LINKED ISSUE #1019: Option to clear all filters
**Description**:
Both in browse and search, it is not possible to easily clear all the filters applied to the table of results. We need to think of a design for this.

**Acceptance criteria**:
- [x] Ability to clear up all the filters.

joshuadkitenge:
## Description
Fixing the page container unit test for the my-data DLS page.  
Making the E2E test less prone to failure 

## Testing instructions

- [ ] Review code
- [ ] Check Actions build
- [ ] {more steps here}

## Agile board tracking
closes #1019

### DIFF
--- a/packages/datagateway-dataview/cypress/integration/card/pageContainer.spec.ts
+++ b/packages/datagateway-dataview/cypress/integration/card/pageContainer.spec.ts
@@ -62,13 +62,6 @@ describe('PageContainer Component', () => {
     cy.url().then((url) => {
       cy.get('[data-testid="advanced-filters-link"]').click();
       cy.get('input[id="Title-filter"]').type('South');
-
-      cy.get('[data-testid="card"]')
-        .first()
-        .contains(
-          'Season identify professor happen third. Beat professional blue clear style have. Light final summer.'
-        );
-
       cy.get('[data-testid="clear-filters-button"]').click();
       cy.url().should('eq', url);
     });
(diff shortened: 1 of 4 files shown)


louise-davies (changes requested):
I dislike hardcoding the url - I think the old solution worked well. I had a look at the test and it seemed to me like the problem was that the test wasn't waiting for the relevant requests to finish so that's why they were unstable.

e.g. I refactored the table pageContainer test like so:

Replace `cy.visit()` in `beforeEach` with
```
cy.intercept('/investigations/').as('getInvestigations');
cy.visit('/browse/investigation/').wait('@getInvestigations');
```

```
it('should be able to click clear filters button to clear filters', () => {
    cy.url().then((url) => {
      cy.get('input[id="Title-filter"]').type('South');
      cy.wait('@getInvestigations');

      cy.get('[aria-rowindex="1"] [aria-colindex="4"]').contains('36');

      cy.get('[data-testid="clear-filters-button"]').click();
      cy.url().should('eq', url);
      cy.get('[aria-rowindex="1"] [aria-colindex="4"]').contains('42');
    });
  });
```

This to me seems to make sense. Could you investigate refactoring the tests to be more like that and see if the tests are stable?

joshuadkitenge:
Ok, It is looking like I revert to the initial testing method (I'll add relevant changes (waiting ) as it was working fine locally and the only issue occurred when being run on the CI

louise-davies:
I was able to recreate the errors locally I think, hence why I realised it was due to there being no waits

joshuadkitenge:
I have re-run the CI 6 times and the E2E tests I have added for the clear filter button always passes. However, some other test of which I haven't changed fails in some runs.

PR #1123, opened 2022-02-15
#927 - Fix incorrect counts (take 2)

### LINKED ISSUE #927: Sorting & filtering can cause the size/fileCount values to be inaccurate

louise-davies:
## Description
There was still a bug in the size/counts batching code, which meant if the number of results decreased from a lot to less than the batch size (i.e. 5) then they would never update, causing the UI to display incorrect sizes/counts. This was fixed by ensuring we clear the sizes/counts array when we detect the data has changed.

I also added unit tests for both the overall batching behaviour & this specific bug (and I tested that removing my fix causes the unit tests to fail)

## Testing instructions
- [ ] Review code
- [ ] Check Actions build
- [ ] Ensure that filtering to a small number of results still shows the correct sizes/counts

## Agile board tracking
Closes #927

### DIFF
--- a/packages/datagateway-common/src/api/datasets.test.tsx
+++ b/packages/datagateway-common/src/api/datasets.test.tsx
@@ -1,5 +1,5 @@
 import { Dataset } from '../app.types';
-import { renderHook } from '@testing-library/react-hooks';
+import { act, renderHook } from '@testing-library/react-hooks';
 import { createMemoryHistory, History } from 'history';
 import axios from 'axios';
 import handleICATError from '../handleICATError';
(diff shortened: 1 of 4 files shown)


joshuadkitenge (changes requested):
Some unit tests in datagateway-search (searchPageContainer , searchPageTable and searchPageCardView) are failing.

PR #1127, opened 2022-02-17
fixing the calculate button when value is 0 #1122

### LINKED ISSUE #1122: Calculate size button doesn't recognise 0 as a valid size
**Description**:
Reported by Chris, it seems as if the Calculate Size button does nothing if the return result is 0. We should instead display this as 0 B instead

**Acceptance criteria**:
- [ ] Calculate Size button correctly handles a response of 0

joshuadkitenge:
## Description
The Calculate button on the DLS proposals on the investigation and dataset details panel
 doesn't recognise the result 0
## Testing instructions

- [ ] Review code
- [ ] Check Actions build
- [ ] {more steps here}

## Agile board tracking
closes #1122

### DIFF
--- a/packages/datagateway-dataview/cypress/integration/table/dls/datasets.spec.ts
+++ b/packages/datagateway-dataview/cypress/integration/table/dls/datasets.spec.ts
@@ -265,9 +265,11 @@ describe('DLS - Datasets Table', () => {
     });
 
     it('and then calculate file size when the value is 0 ', () => {
+      cy.intercept('/getSize', '0');
+
       // need to wait for counts to finish, otherwise cypress might interact with the details panel
       // too quickly and it rerenders during the test
-      cy.intercept('/getSize', '0');
+
       cy.contains('[aria-rowindex="1"] [aria-colindex="4"]', '55').should(
         'exist'
       );
--- a/packages/datagateway-dataview/cypress/integration/table/dls/datasets.spec.ts
+++ b/packages/datagateway-dataview/cypress/integration/table/dls/datasets.spec.ts
@@ -264,6 +264,25 @@ describe('DLS - Datasets Table', () => {
       cy.contains('4.55 GB', { timeout: 10000 }).should('be.visible');
     });
 
+    it('and then calculate file size when the value is 0 ', () => {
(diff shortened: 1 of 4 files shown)


joshuadkitenge:
0 btyes issue

packages/datagateway-dataview/cypress/integration/table/dls/datasets.spec.ts
```diff
@@ -264,6 +264,25 @@ describe('DLS - Datasets Table', () => {
       cy.contains('4.55 GB', { timeout: 10000 }).should('be.visible');
     });
 
+    it('and then calculate file size when the value is 0 ', () => {
+      // need to wait for counts to finish, otherwise cypress might interact with the details panel
+      // too quickly and it rerenders during the test
```
joelvdavies, line 269: I think this comment should be moved one line below as it refers to the .should('exist').

packages/datagateway-dataview/cypress/integration/table/dls/visits.spec.ts
```diff
@@ -215,6 +215,23 @@ describe('DLS - Visits Table', () => {
       cy.contains('11.01 GB', { timeout: 10000 }).should('be.visible');
     });
 
+    it('and then calculate file size when the value is 0', () => {
+      // We need to wait for counts to finish, otherwise cypress
+      // might interact with the details panel too quickly and
+      // it re-renders during the test.
```
joelvdavies, line 221: Same here.

joelvdavies (changes requested):
Looks good, just think the comments could be confusing.

packages/datagateway-dataview/cypress/integration/table/dls/visits.spec.ts
joelvdavies: No I think those are correct, its just the new intercept should be above the comment since I think it refers to ` cy.contains('[aria-rowindex="1"] [aria-colindex="4"]', '55').should(
        'exist'
      );`

joelvdavies (approved):
Comments addressed. Reviewed changes, CI passes.

PR #1154, opened 2022-03-09
Use react-router links for example searches

louise-davies:
## Description
Based off of Joel's work in another PR that got reverted, I fixed the bug I found in Joel's code so now this seems to work fairly well - meaning you can click on example search links and the whole website doesn't reload. Additionally, I applied the fix to the advanced search modal - currently clicking on a link doesn't close the modal which I kinda like as you can be reading the help and click a link and see the result of the search in the background whilst still reading the modal, but it's fairly trivial to attach the onClose method to onClick props on the Links as well.

## Testing instructions
- [ ] Review code
- [ ] Check Actions build


louise-davies:
@joelvdavies this is ready for review now

joelvdavies (approved):
Works well, CI passes. Looks like I needed to use `searchTextURL` instead of `location.search` in the useEffect dependencies 😄.

louise-davies:
@joelvdavies yeah, `location.search` changed when the dates and the checkboxes changed as well which was causing the bug I reported last time. Switching it to only consider the searchText fixed it

PR #1184, opened 2022-03-28
DG download "remove all" button disables once pressed #1177

### LINKED ISSUE #1177: Disable Remove All button in download table after it's clicked when the request is processing
**Description**:
Currently, users can submit multiple Remove All requests as the button is always active, but this can cause errors if the Remove All operation takes some time and the user clicks it multiple times - we would get deadlock errors. We should disable the button and probably also show some kind of loading indicator to show the user their request is processing.

**Acceptance criteria**:
- [ ] User cannot click remove all button multiple times for the same set of data

sam-glendenning:
## Description
Once pressed, the remove all button is disabled and displays an indeterminate loading spinner until the request completes. This prevents excessive requests as the user was previously able to click this button several times, which could cause an error.

I found it difficult to get Jest's fake timers to work properly, which resulted in a unit test that isn't the best but works for now. I think this is due to multiple levels of promises trying to resolve and becoming conflicted. So anytime I try to use `jest.useFakeTimers()` I get an exceeded timeout error. See my comment below which points out the test in question. Any feedback on how to properly test this would be appreciated!

## Testing instructions
This behaviour usually only presents itself when there are a large number of items in the cart. Try adding a lot of individual datafiles to the cart. This will hopefully allow for the new behaviour to display for long enough to see in action.

- [ ] Review code
- [ ] Check Actions build

## Agile board tracking
Closes #1177

### DIFF
--- a/packages/datagateway-download/src/downloadCart/downloadCartTable.component.test.tsx
+++ b/packages/datagateway-download/src/downloadCart/downloadCartTable.component.test.tsx
@@ -245,6 +245,37 @@ describe('Download cart table component', () => {
     expect(wrapper.exists('[data-testid="no-selections-message"]')).toBe(true);
   });
 
+  it('disables remove all button while request is processing', async () => {
+    (removeAllDownloadCartItems as jest.Mock).mockImplementation(() => {
+      return new Promise((resolve) => setTimeout(resolve, 2000));
+    });
+
+    const wrapper = createWrapper();
+
+    await act(async () => {
+      await flushPromises();
+      wrapper.update();
+      await flushPromises();
+      wrapper.update();
+    });
+
+    await act(async () => {
+      wrapper.find('button#removeAllButton').simulate('click');
+      await flushPromises();
+      wrapper.update();
+    });
+    expect(
+      wrapper.find('button#removeAllButton').prop('disabled')
+    ).toBeTruthy();
+
+    setTimeout(() => {
+      wrapper.update();
+      expect(wrapper.exists('[data-testid="no-selections-message"]')).toBe(
+        true
+      );
+    }, 2001);
+  });
+
   it("removes an item when said item's remove button is clicked", async () => {
     const wrapper = createWrapper();
 
--- a/packages/datagateway-download/src/downloadCart/downloadCartTable.component.tsx
+++ b/packages/datagateway-download/src/downloadCart/downloadCartTable.component.tsx
@@ -20,6 +20,7 @@ import {
   makeStyles,
   Theme,
   Link,
+  CircularProgress,
 } from '@material-ui/core';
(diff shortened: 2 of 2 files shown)


packages/datagateway-download/src/downloadCart/downloadCartTable.component.test.tsx
```diff
@@ -245,6 +245,37 @@ describe('Download cart table component', () => {
...
+    expect(
+      wrapper.find('button#removeAllButton').prop('disabled')
+    ).toBeTruthy();
+
+    setTimeout(() => {
+      wrapper.update();
+      expect(wrapper.exists('[data-testid="no-selections-message"]')).toBe(
+        true
+      );
+    }, 2001);
```
sam-glendenning, line 276: This is what I'm not a fan of. It manually adds a non-mocked timeout to the test to force the wrapper to update and display the correct thing on the DOM. We want to use fake timers to adhere to good testing practice. Any help on this one would be appreciated!

louise-davies: Fake timers and promises inherently don't work well with one another - see https://github.com/ral-facilities/scigateway/blob/169b596d761fc7426136faca946db692849f3f19/src/App.test.tsx#L32 for an example of me working around it. See this issue for info: https://github.com/facebook/jest/issues/10221. One option is to try using the legacy Jest fake timers, as for example there are a few tests in SG which work with old timers and not with the new ones (like the one I linked). However, the only way I could get things working was to manually wait for real timers. Perhaps using a dedicated `sleep` function or even just a `sleep` line of code like `await new Promise((r) => setTimeout(r, 2000));` is more semantic, so you use that line then you wrapper.update/assert to your heart's content rather than within the timeout itself?

sam-glendenning: Thanks for that, I'll use that example as a reference for another attempt. I had a go with the legacy timers, to no avail. Perhaps this will help. I'll ping you again when I have something ready for review :)

packages/datagateway-download/src/downloadCart/downloadCartTable.component.test.tsx
```diff
@@ -245,6 +245,37 @@ describe('Download cart table component', () => {
     expect(wrapper.exists('[data-testid="no-selections-message"]')).toBe(true);
   });
 
+  it.only('disables remove all button while request is processing', async () => {
```
sam-glendenning, line 248: Note to remove .only

louise-davies (approved):

louise-davies:
As discussed - I'll likely be refactoring this again in #1185 so I'll take a look at the unit test again there

PR #1185, opened 2022-03-28
Refactor download cart to improve performance

### LINKED ISSUE #1215: Improve download cart performance
**Description**:
> I selected 3k files from mx21625-8 (selecting first file then paging down for ~1 min and shift selecting the last visible file).
> I then went to the cart, it takes quite a while to display any selected files so you are left with a mostly blank screen for ~10 s (TC seems to be quicker for the same set of files but also the flicking progress bar suggests there is something happening, there is nothing obvious in DGW)
> The cart file count and size calculation in DGW takes about the same time to complete fully as TC, but there is no ‘count up’ of the number of files like there is in TC – this makes you wonder if anything is happening.

**Acceptance criteria**:
- [ ] Cart performance improved
- [ ] Show more loading indicators

louise-davies:
## Description
I refactored the download cart to use react-query as I figured that woud be best as we already have quite optimised code for the things the download is doing already. Then I apply these changes as well as address other comments (i.e. display the progress of the counting of files and sizes, improve time to first data display, display loading indicator properly)

## Testing instructions

- [ ] Review code
- [ ] Check Actions build
- [ ] Add lots of items to the cart and test the performance (Chris mentioned 3000 files) - might be worth deploying this on DLS preprod to more effectively test performance (as I at least get degraded performance due to many OPTIONS requests when running locally)

## Agile board tracking
Closes #1215


packages/datagateway-download/src/downloadCart/downloadCartTable.component.test.tsx
```diff
@@ -180,8 +203,13 @@ describe('Download cart table component', () => {
...
     expect(getDatafileCount).toHaveBeenCalled();
+    expect(wrapper.find('[aria-colindex=4]').find('p').first().text()).toEqual(
+      '7'
+    );
+
+    // TODO fix this. It seems getDatafileCount is not being called if entityType === 'datafile'
     expect(wrapper.find('p#fileCountDisplay').text()).toEqual(
-      expect.stringContaining('downloadCart.number_of_files: 28')
+      expect.stringContaining('downloadCart.number_of_files: 28 / 5000')
     );
```
sam-glendenning, line 213: There's something weird going on here and I'm struggling to figure out what and why. The test is correct but failing - there are 4 cart items, `getDatafileCount` should return 7 for every item, so we should get 28 as the number of files. However, we are instead getting a result of 22 and it seems to be because whenever `entityType` === 'datafile', we return 1. But because `getDatafileCount` is being mocked, this 1 is not coming from this function. My guess is it's coming somewhere from `useDatafileCounts()`

```
// useDatafileCounts
return data
      ? data.map((cartItem) => {
          const { entityId, entityType } = cartItem;
          return {
            queryKey: ['datafileCount', entityId],
            queryFn: () =>
              datafileCountslimit(getDatafileCount, entityId, entityType, {
                apiUrl,
              }),
            onError: (error) => {
              handleICATError(error, false);
            },
            staleTime: Infinity,
            enabled: entityType !== 'datafile',
            initialData: entityType === 'datafile' ? 1 : undefined,        // Maybe to do with this?
          };
        })
      : [];
```
So for whatever reason, `getDatafileCount` is not being called if the `entityType` is a datafile. I'm having an ill day and am still to fully refamiliarise myself with React Query to figure out how to fix this at the moment but thought this would be a good note to raise.

louise-davies: @sam-glendenning ah, this is my bad that I added this optimisation but didn't adjust the test for it. We don't need to send a request to count datafiles for a single datafile (as we know it's 1), and I figured the most efficient way to stop this is to disable the react-query request if `entityType === 'datafile'`, and we essentially "mock" the data by using `initialData` when the `entityType` is `datafile`. This ensures that the query function is not called for datafiles (thus better optimisation) but that the `data` prop of the query result is correctly `1`.

I'll fix the test to account for this behaviour

sam-glendenning: I thought so! Nice bit of code nonetheless

packages/datagateway-download/src/downloadApiHooks.ts
```diff
@@ -0,0 +1,231 @@
...
+//     {
+//       onSuccess: (data) => {
+//         queryClient.setQueryData('cart', data);
+//       },
+//       onError: (error) => {
+//         handleICATError(error);
+//       },
+//     }
+//   );
+// };
```
sam-glendenning, line 141: Is there a plan to tackle this at some point?

louise-davies: I'll likely make it a separate issue once this is merged in - I didn't want to do all that work in this one PR which is already changing a lot

sam-glendenning (approved):

PR #1171, opened 2022-03-24
Organise dev and prod dependencies

VKTB:
## Description
This PR moves the build dependencies from `devDependencies` to `dependencies` in the `package.json` files so that production dependencies can be installed and optimised production build can be created.

The `--production` option is deprecated on `install` in `yarn` 2, so `yarn workspaces focus --all --production` should be used instead of `yarn install --production`. To run this command, the `yarn` `workspace-tools` plugin needs to be installed. It can be installed by running `yarn plugin import workspace-tools`.

## Testing instructions
- [ ] Review code
- [ ] Check Actions build
- [ ] Review changes to test coverage
- [ ] `yarn workspaces focus --all --production` followed by `yarn build` at the root of the project should install and create production optimised builds for the plugins

## Agile board tracking
closes #1155

PR #1129, opened 2022-02-18
React 17 and MUIv5 upgrade #830 #1096

joelvdavies:
## Description
Upgrades React 16 to 17, MUIv4 to v5 and yarn to yarn 2.

## Notes
- Simplified tooltip as close on escape now handled by MUI
- Refactored styling to use the new sx prop and `styled` instead of `style` (for MUI components only) and `makeStyles` (see https://mui.com/guides/migration-v4/#migrate-from-jss)
- Removed `generateClassNames` in App.tsx files as should no longer be necessary since emotions class names are hash based (I have not seen any evidence of collisions so far)
- Investigated https://github.com/ral-facilities/datagateway/issues/821 and it is still broken, it seems the position will not update without the resize event, only the width
- Emotion is now used for creating animations for `mark/sticky/selectionAlertBanner' in common
- Have added `searchTabs` component to replace `seachPageTable` & `searchCardView` as previously the tabs and their styling were defined twice making the code harder to maintain
- Have installed and used `@testing-library/react` in a few places, particularly where there were giant snapshots that got even bigger after the upgrade (e.g. `downloadConfirmDialog`) as testing-library resolves to standard html elements in snapshots instead of MUI ones with tonnes of styling
- Date pickers now use @mui/lab - by default in testing these pickers end up using their mobile counterparts so have added an 'applyDatePickerWorkaround' function in setupTests.tsx which applies a workaround so that they render as desktop date pickers otherwise their contents can't be changed in tests
- Have upgraded to yarn 2 as this fixed an issue with installing the latest @mui/lab (following https://yarnpkg.com/getting-started/migration)

## Dev Environment Changes
- Due to the yarn 2 upgrade you may want follow step 1 in https://yarnpkg.com/getting-started/migration to ensure you are also using the latest yarn - I had to use https://stackoverflow.com/questions/57673913/vsc-powershell-after-npm-updating-packages-ps1-cannot-be-loaded-because-runnin as well

## Testing instructions
Test with https://github.com/ral-facilities/scigateway/pull/930.

- [ ] Review code
- [ ] Check Actions build
- [ ] Review changes to test coverage

## Agile board tracking
Closes #830, Closes #1096, Closes #1155

Issue #1096, opened 2022-02-07
Upgrade to MUI v5


joelvdavies:
**Description**:
Upgrade to MUI v5 following https://mui.com/guides/migration-v4/.

**Acceptance criteria**:
- [ ] Upgraded to MUI v5

Issue #830, opened 2021-09-08
Upgrade to React 17


GoelBiju:
**Description**:

At present in DataGateway we are using React 16.13.1. With the release of React 17, it would be ideal to upgrade to the newer version. The official [blog post](https://reactjs.org/blog/2020/10/20/react-v17.html) highlights the benefits of upgrading.

The only issue preventing an upgrade at the moment is there not being an official Enzyme adapter for React 17. There is an issue on the Enzyme repository (https://github.com/enzymejs/enzyme/issues/2429) which is currently in progress. This upgrade would have to wait until that is officially created.

I have created a matching issue in SciGateway for upgrading to React 17 (https://github.com/ral-facilities/scigateway/issues/768). Both will need to be upgraded at the same time.

**Acceptance criteria**:

- [ ] Upgrade DataGateway to React 17 (upgrade react and react-dom, check other packages too),
- [ ] Ensure types for React is up to date,
- [ ] Enzyme Adapter is updated for React 17.

GoelBiju:
@louise-davies There is an [unofficial](https://github.com/wojtekmaj/enzyme-adapter-react-17) Enzyme adapter but we would need to look at whether it works fine with TypeScript (?).

louise-davies:
I think what we need to do is to:
- [ ] Do upgrade to React 17 (react & react-dom), including type declarations
  - [ ] Fix anything broken due to React 17 upgrade
- [ ] Use unofficial Enzyme 17 adapter for now
  - [ ] Check that tests pass at this point
- [ ]  Future: evaluate whether it's worth switching over to testing-library/react. Many libraries are switching/have switched (including Material UI and that means to upgrade to Material UI 5 would be harder as we use their enzyme utils) and it seems to be "reacting" (pun intended) to React version updates an a more graceful way - which will help with future React updates. We can slowly move tests over component at a time as a "background" task (or e.g. next time someone works on some code, try to migrate the tests for that code)

joelvdavies:
Here are some problems that need fixing after the upgrade:
- [ ]  Search page does not initially display results until tab is switched (seems set_current_tab action is never called initially)

PR #1187, opened 2022-03-29
Create Dockerfile

VKTB:
## Description
Creates a multipart build `Dockerfile` to build and deploy the DataGateway plugins on ports 5001, 5002 and 5003. Also creates a GitHub Actions workflow for building and pushing the `Dockerfile` to the SCD harbor Docker registry. It uses a harbor robot account who's credentials are stored as secrets in this repo.

### DIFF
--- a/.github/workflows/docker-build.yml
+++ b/.github/workflows/docker-build.yml
@@ -1,3 +1,8 @@
+# This workflow file builds and pushes the Dockerfile
+# located in this repo, to the SCD harbor Docker registry.
+# It uses a harbor robot account, who's credentials are 
+# stored as secrets in this repo.
+
 name: Build & Push Docker Image
 
 on:
--- a/.github/workflows/docker-build.yml
+++ b/.github/workflows/docker-build.yml
@@ -25,7 +25,7 @@ jobs:
         id: meta
         uses: docker/metadata-action@98669ae865ea3cffbcbaa878cf57c20bbf1c6c38
         with:
-          images: harbor.stfc.ac.uk/datagateway
+          images: harbor.stfc.ac.uk/datagateway/dgw
       
       - name: Build and push Docker image
         uses: docker/build-push-action@ad44023a93711e3deb337508980b4b5e9bcdc5dc
--- a/.github/workflows/docker-build.yml
+++ b/.github/workflows/docker-build.yml
@@ -21,7 +21,7 @@ jobs:
         uses: actions/checkout@v2
       
       - name: Log in to Harbor
-        uses: docker/login-action@f054a8b539a109f9f41c372932f1ae047eff08c9
+        uses: docker/login-action@v1
         with:
           registry: harbor.stfc.ac.uk/datagateway
           username: ${{ secrets.HARBOR_USERNAME }}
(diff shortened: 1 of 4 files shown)


.github/workflows/docker-build.yml
```diff
@@ -0,0 +1,42 @@
...
+# It uses a harbor robot account, who's credentials are 
+# stored as secrets in this repo.
+
+name: Build & Push Docker Image
+
+on:
+  workflow_dispatch:
+  push:
+    branches:
+      - feature/adding-docker-file
```
VKTB, line 12: Once this PR is merged into `feature/k8s-deployment`, this workflow will no longer trigger. If we want it to trigger then we need to change this.

Dockerfile
```diff
@@ -0,0 +1,41 @@
...
+# TODO: use yarn install --production:
+# https://github.com/ral-facilities/datagateway/issues/1155
+
+# Install dependancies
+COPY . .
+RUN yarn install
+RUN yarn build
+
+# Put the output of the build into an apache server
+FROM httpd:alpine
```
VKTB, line 16: Should we specify a version here? I am not sure if it just grabs the latest version if a version isn't specified but if it does then that may potentially cause issues when new versions are released in future.

Dockerfile
```diff
@@ -0,0 +1,41 @@
...
+WORKDIR /usr/local/apache2/htdocs
+COPY --from=build /datagateway/packages/ .
```
VKTB, line 18: Can't we just copy the `build` directories of the plugins rather than copying everything? I don't think the apache server needs to serve source code files etc.

Dockerfile
```diff
@@ -0,0 +1,41 @@
+# Multipart build dockerfile to build and serve datagateway
+
+FROM node:16-alpine as build
+WORKDIR /datagateway
+ENV PATH /datagateway/node_modules/.bin:$PATH
+
+# TODO: use yarn install --production:
```
moonraker595, line 7: Where did we get to with the `use yarn install --production` issue?

moonraker595: Was it dependency issues related to the latest version of yarn?

VKTB: I organised the dev and prod dependencies as part of https://github.com/ral-facilities/datagateway/pull/1171 which is needed to successfully do a production install. Once this PR is merged, the install should be succeeding. However, that PR will merge the changes into the `feature/react-17-upgrade-new-#830` branch (PR for it [here](https://github.com/ral-facilities/datagateway/pull/1129)) which deals with a few upgrades, one of which is yarn 2, so we cannot really address the TODO until both of them are merged. I will update the table in the wiki that you created to track this too.

Dockerfile
VKTB: I tested Node 17 but the build failed due to some of the build tools not yet supporting it. Sticking to Node 16 for now, in fact the [release build workflow](https://github.com/ral-facilities/datagateway/blob/master/.github/workflows/release-build.yml) uses Node 14 so it doesn't seem to matter too much.

I have configured for Yarn 1 (version 1.22.18) to be used for now and added a TODO to upgrade to Yarn 2 once the project is upgraded to Yarn 2 which will happen after [React 17 and MUIv5 upgrade #830 #1096](https://github.com/ral-facilities/datagateway/pull/1129) is merged. I will add a note to the wiki.

I also set a version for `httpd`.

moonraker595 (approved):

PR #1192, opened 2022-03-31
Prevent users from downloading carts with empty items

louise-davies:
## Description

Disable download button if there are empty items in the cart
and also display a message to the user telling them this. Plus add a file count column so that users can see the problematic data

## Testing instructions

- [ ] Review code
- [ ] Check Actions build
- [x] I will be deploying this change to ISIS prod shortly so can be tested there

## Agile board tracking
Hotfix so not connected to an issue

### DIFF
--- a/packages/datagateway-download/public/res/default.json
+++ b/packages/datagateway-download/public/res/default.json
@@ -75,6 +75,7 @@
     "name": "Name",
     "type": "Type",
     "size": "Size",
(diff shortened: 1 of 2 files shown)


sam-glendenning (dismissed):

sam-glendenning (approved):
Approving further changes relating to page height adjustment

PR #1203, opened 2022-04-01
#1199 Fix instrument view default sorts when switching between study hierarchy and not

### LINKED ISSUE #1199: Facility Cycles and Experiments page sorts are lost
**Description**:
When navigating to ISIS's _Facility Cycles_ and _Experiments_ pages, the instruments are sorted by name in ascending order but the sort is lost if the same page that is displayed is accessed from the menu again. The sort is also lost on the second accessed page if one page is accessed first and the other after.

When the sort is lost, this is reflected in the URL - it no longer contains the `?sort` part
DG

**Acceptance criteria**:
- [ ] Sorting is always applied on the _Facility Cycles_ and _Experiments_ pages

louise-davies:
## Description
Fix Instrument views not remounting when switching between studyHierarchy true & false by adding a key - this forces React to remount and thus applies the initial mount logic. We need to force remount as the default sort logic only applies on mount.

instrument-switch-fix

## Testing instructions
- [ ] Review code
- [ ] Check Actions build
- [ ] See gif where I added links to switch between /browse/instrument and /browseStudyHierarchy/instrument without reloaded page. Or mount in SG

## Agile board tracking
Fixes #1199

### DIFF
--- a/packages/datagateway-dataview/public/datagateway-dataview-settings.example.json
+++ b/packages/datagateway-dataview/public/datagateway-dataview-settings.example.json
@@ -117,31 +117,31 @@
     },
     {
       "section": "Browse",
-      "link": "/browse/proposal",
+      "link": "/browse/proposal?sort={\"title\":\"asc\"}",
       "displayName": "DLS View",
       "order": 2
     },
     {
       "section": "Browse",
-      "link": "/browse/instrument",
+      "link": "/browse/instrument?sort={\"fullName\":\"asc\"}",
       "displayName": "ISIS View",
       "order": 3
     },
     {
       "section": "Browse",
-      "link": "/browseStudyHierarchy/instrument",
+      "link": "/browseStudyHierarchy/instrument?sort={\"fullName\":\"asc\"}",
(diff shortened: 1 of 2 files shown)


sam-glendenning (changes requested):
Decent but doesn't seem to fix the issue. I mounted with SG to test and I can switch between the two views with the correct sort maintained but clicking the same view again removes the correct sort

ezgif com-gif-maker

sam-glendenning:
That's what I was thinking. If we really only ever want that sort as the default, starting sort of the page, why not add those params in the links too?

louise-davies:
@sam-glendenning this needed some code changing in SG, so there's a PR upcoming there 😅

louise-davies:
instrument-switch-fix-2

@sam-glendenning here's an updated GIF showing the behaviour using ral-facilities/scigateway#1017 (ignore the two testing links I left in DG dataview and couldn't be bothered rebuilding 😅)

sam-glendenning (approved):

PR #1216, opened 2022-04-06
Adding ability to use DatePicker or DateTimePicker in date filter column #1175

### LINKED ISSUE #1175: DatePickers in AdminDownloadTable needs to be a DateTimePicker
**Description**:
Most of the pickers in TopCAT were just date pickers, but there's one view - the Admin Download table - where they should instead be DateTimePickers. We should probably add a prop to our existing DateFilter component to let it be both a date picker and a datetime picker.

**Acceptance criteria**:
- [ ] DateFilter component can render date pickers or datetime pickers
- [ ] Datetime picker is used in Admin download table

sam-glendenning:
## Description
Depending on if a `filterByTime` prop is passed, the date filter component will return either a `KeyboardDatePicker` or a `KeyboardDateTimePicker`. Thus, the admin download table now contains a `KeyboardDateTimePicker` for more accurate filtering of results, as per user request.

This required some amendments to how time is passed to the backend date query. Previously, if a date was passed, it was assumed the time value would be 00:00:00 or 23:59:59, depending on `startDate` or `endDate` being changed. Now we must provide a time value if the user specifies one.

Fortunately, the formatting for this is done in the `dateColumnFilter` component. However, the GUI for `KeyboardDateTimePicker` can only be accurate to hours and minutes - we can't enter a value for seconds. 

This causes a slight amendment to queries - if we enter an end date of `2022-04-06 23:59`, we won't get any results for `2022-04-06 23:59:00` < x < `2022-04-06 23:59:59`. We'd need to update the query to an end date of `2022-04-07 00:00`. If anyone has any thoughts on this, how this might impact users, other unintended side-effects, solutions for including a value for seconds, etc. please comment!

## Testing instructions
Because this is a change to the globally-used date filter, it is important to test date filtering in a variety of pages, including dataview, search, download and admin download. Currently, only the admin download table wants date and time filtering so make sure that other filters are date filter only and continue to work as expected. Finally, check to make sure that date and time values can be filtered on the admin download table correctly.

- [ ] Review code
- [ ] Check Actions build

## Agile board tracking
Closes #1175

### DIFF
--- a/packages/datagateway-common/src/table/columnFilters/dateColumnFilter.component.tsx
+++ b/packages/datagateway-common/src/table/columnFilters/dateColumnFilter.component.tsx
@@ -1,4 +1,4 @@
-import React, { useState } from 'react';
+import React, { ReactElement, useState } from 'react';
 import DateFnsUtils from '@date-io/date-fns';
 import { format, isValid, isEqual } from 'date-fns';
 import {
@@ -74,6 +74,19 @@ interface DateColumnFilterProps {
   filterByTime?: boolean;
 }
 
+interface DatePickerCommonProps {
+  clearable: boolean;
+  allowKeyboardControl: boolean;
+  invalidDateMessage: string;
+  ariaHidden: boolean;
+  format: string;
+  color: 'secondary' | 'primary' | undefined;
+  strictCompareDates?: boolean;
+  okLabel: ReactElement;
+  cancelLabel: ReactElement;
+  clearLabel: ReactElement;
+}
+
 const DateColumnFilter = (props: DateColumnFilterProps): React.ReactElement => {
   //Need state to change otherwise wont update error messages for an invalid date
   const [startDate, setStartDate] = useState(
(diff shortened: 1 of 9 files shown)


louise-davies (changes requested):
The regular download table also needs a time picker (I checked TopCAT and it has this)

The seconds bit is fine for now - I checked and MUIv5 has support for seconds in its datetimepicker so we can add support for that once that PR is merged in

packages/datagateway-common/src/table/columnFilters/dateColumnFilter.component.tsx
```diff
@@ -79,74 +90,149 @@ const DateColumnFilter = (props: {
...
-          cancelLabel={<span style={{ color: buttonColour }}>Cancel</span>}
-          clearLabel={<span style={{ color: buttonColour }}>Clear</span>}
-        />
-      </MuiPickersUtilsProvider>
+      {props.filterByTime ? (
+        <MuiPickersUtilsProvider utils={DateFnsUtils}>
+          <KeyboardDateTimePicker
+            clearable
+            allowKeyboardControl
+            style={{ whiteSpace: 'nowrap' }}
```
louise-davies, line 98: It's probably worth extracting some of these props into a `commonProps` variable which we spread via `{...commonProps}` onto the 4 picker components we have to reduce code repetition

louise-davies (dismissed):
These changes now look good to me - please just check over the minor edit to the common types prop though (I made it so we use the actual library to define the prop types and extracted some more common props)

sam-glendenning:
> These changes now look good to me - please just check over the minor edit to the common types prop though (I made it so we use the actual library to define the prop types and extracted some more common props)

This looks much nicer, good spot. You'll have to review again due to a change in the base branch.

louise-davies (approved):

PR #1228, opened 2022-04-20
Adding alert label if file/size limit exceeded in selections table #1183

### LINKED ISSUE #1183: Display warning to user if they exceed number of file/total size limits
**Description**:
TopCAT displays a warning like so:

image

We should also display a warning (either via a notification or like an error message similar to TopCAT) to the user in addition to disabling the button

**Acceptance criteria**:
- [ ] If user has exceeded maximum file count/file size limits, then a warning is visually displayed to the user that they have exceeded a limit

sam-glendenning:
## Description
This implementation is designed to mimic Topcat as much as possible.

Here is Topcat:
160371182-998c26c0-8e54-410b-923e-3589b8358880

~~And here is DataGateway with this feature:~~
Capture

Later amendment - this is the updated screenshot
image

Both the text and the label are open to change if we feel its too wordy, etc. Alternatively, if we don't like the design at all, we can have an error notification appear instead. I wasn't too keen on this idea as it may cause notifications to appear too frequently if the user is excessively amending the selections table. I was also more of a fan of the error message appearing next to the offending item so the user can more easily tell what's wrong.

## Testing instructions
Best run through SciGateway so the selections can be customised. To make things easier, change the `fileCountMax` and `totalSizeMax` values in download settings to much smaller values so that the labels can be displayed more easily. Try adding and removing things from the selections - the labels should appear and disappear as necessary.

- [ ] Review code
- [ ] Check Actions build

## Agile board tracking
Closes #1183

### DIFF
--- a/packages/datagateway-download/src/downloadCart/downloadCartTable.component.tsx
+++ b/packages/datagateway-download/src/downloadCart/downloadCartTable.component.tsx
@@ -24,6 +24,7 @@ import {
   CircularProgress,
 } from '@material-ui/core';
 import { RemoveCircle } from '@material-ui/icons';
+import { Alert } from '@material-ui/lab';
 import {
   useCart,
   useRemoveEntityFromCart,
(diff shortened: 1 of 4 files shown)


packages/datagateway-download/src/downloadCart/downloadCartTable.component.tsx
```diff
@@ -351,51 +352,83 @@ const DownloadCartTable: React.FC<DownloadCartTableProps> = (
...
+              <Grid item>
+                <Alert
+                  id="fileLimitAlert"
+                  variant="filled"
+                  severity="error"
+                  icon={false}
+                  style={{ padding: '0px 8px', lineHeight: 0.6 }}
+                >
+                  Too many files - you have exceeded limit of {fileCountMax}{' '}
+                  files - please remove some files
```
louise-davies, line 388: These strings need to be in the translation file - you should still be able to parameterise `fileCountMax` and `totalSizeMax`

louise-davies (changes requested):
See above comment. Also, if we're using the `Alert` component for the other two warnings, should we use this warning for the `empty_items_warning`? It would make it look more error like and since we've already included the `mui-lab` library...

packages/datagateway-download/src/downloadCart/downloadCartTable.component.tsx
sam-glendenning: I've given this some thought and decided to combine our ideas. I say we provide one `<Alert>` object, like the one I've done, in the place where your `empty_items_warning` is so that they aren't too many error messages on the page. We keep that `<Alert>` there, with the appropriate error message, until no further errors remain on the page.

louise-davies (commented):
Good that the changes are addressed - I'm just wondering if it's better to only display one error or to display the errors individually. I'm thinking of times when I've been frustrated when filling out a form that only displayed one error meant that I felt I was playing whack-a-mole with the errors vs if you show all the errors individually then the user knows everything they need to know at once.

Although I'm not sure it's too big of a deal, since the error message will update on every action the user does to try and remedy the message (i.e. removing items from the cart via the remove item buttons) so perhaps it's fine to only show a single error? Thoughts?

sam-glendenning:
Kinda flip flopping on the idea. Maybe it is nicer to put the error message right next to the offending item for convenience (first screenshot). However, if all errors are present, this could cause:

- error message next to file count
- error message next to size calculation
- error message below all them and above download button if empty files present

To top it all off, I presume we'd want to use <Typography> elements in all error messages as well, which causes a larger height of the entire message, which could push the height of the page down too.

louise-davies:
@sam-glendenning true about the height of the page, but I don't mind that too much as it should be a rare occurrence that the user needs to rectify themselves right? The empty item message should be very rare I'd hope, and not occur at the same time as the max items messages often. Also I wouldn't think we'd need a `Typography` element as the `Alert` is already a MUI component which expects text right? Main reason to use `Typography` is to ensure text inherits MUI props/theming etc.

sam-glendenning:
Fair enough. I'll put the alerts back where they were

sam-glendenning:
@louise-davies back we go again!

image

louise-davies (approved):

PR #1231, opened 2022-04-22
Fix selection errors #1210

### LINKED ISSUE #1210: Errors when multi selecting items for the cart
**Description**:
> During the selection of large number of datasets or large number of files (using shift key and scrolling up or down) I received a network error.
> image
> And showing the same when selecting many datafiles:
> image

**Acceptance criteria**:
- [ ] TDB - awaiting demo of issue

louise-davies:
## Description
Requires a build of specific icatproject/topcat#483/ral-facilities/datagateway-download-api#12 branches.

This PR changes the multi-remove `useRemoveFromCart` method to use the same endpoint as `useAddToCart` with the remove parameter. I also added 431 retries to all TopCAT `useMutation` uses (as `useMutation` doesn't automatically retry 3 times like `useQuery` does). This should hopefully remove all errors.

I didn't bother changing the remove from cart methods in `datagateway-download` as they either supply a single ID or an asterisk and so doesn't have the URL length problem.

## Testing instructions

- [ ] Review code
- [ ] Check Actions build
- [ ] I will be deploying this change and the relevant TopCAT snapshot to DLS preprod so it can be fully tested.

## Agile board tracking
Closes #1210


louise-davies:
Note - the e2e tests will fail on CI until we release the relevant version of TopCAT to deploy on CI

sam-glendenning (approved):
Works great! Good use of `retry` to give the request another go in the unlikely event it fails

sam-glendenning:
Quick final comment: are we still using the `removeFromCart` function is cart.tsx? Or can we remove it?

louise-davies:
I'm pretty sure it's exported and used in `datagateway-download` as when I refactored download I tried to reuse any base axios request functions to use in the new react-query hooks, but I can move it to download if you think that's a better place for it now.

sam-glendenning:
Ah yes I see, it's used in the `useRemoveEntityFromCart` hook, my bad. In my opinion, `removeFromCart` probably doesn't belong in `cart.tsx` anymore, I think it's best we move it to `downloadApi` or something similar.

sam-glendenning (dismissed):

sam-glendenning:
Just a couple of things I wanted to ask, since I've forgotten some of the context around this fix:

1. Error code 431 is quite specific, were we always seeing this code and is it always this specific number?
2. Apologies if I'm missing something but error code 431 corresponds with "The server is unwilling to process the request because either an individual header field, or all the header fields collectively, are too large". If we retry the same request when getting this error, why would it work a second time?

louise-davies:
@sam-glendenning this is an intermittent error that TopCAT sometimes gives - the only way to fix it was to redeploy TopCAT but in the meantime, if you retried any request then it might go through within 3 tries. The same request could fail with 431 and then go through when you retry it. It is always 431 and we don't know why TopCAT was responding like that. It's only significant in that `useMutation` by default doesn't retry requests at all (due to them being used to mutate things on the server I would guess) - so get requests to TopCAT would be fine as the 431 responses would trigger retries but the `useMutations` would always fail.

sam-glendenning (approved):
Thanks for the clarification! Happy with this, works well after testing on Diamond preprod

PR #1232, opened 2022-04-22
Preventing download of empty files #1195

### LINKED ISSUE #1195: Individually downloading empty (0 bytes) files redirects to broken IDS download page
**Description**:
As per [this change](https://github.com/ral-facilities/datagateway/pull/1192), users can no longer download all items in their selections if their selections include an empty file (size of 0 B). In this case, they are presented with a warning message above the download button, which is disabled until they remove the offending selection(s).

However, users can still individually attempt to download these files in Dataview, which results in a redirect to a broken IDS download page, presumably because the file does not exist or is impossible to download in whatever form it exists. We want to prevent the user from doing this if the file is empty. Perhaps either disable or remove the download button? It may also be helpful to provide the user with some information on why they cannot download the file, similarly to the selections cart. Perhaps this could come in the form of a tooltip which displays when the user hovers the mouse over a disabled download button?

**Acceptance criteria**:
- [ ] User can no longer download empty files in table or card view
- [ ] Check site-wide that they are no more possible methods of initiating a download of an empty file

sam-glendenning:
## Description
This requires passing the size of the relevant entity to the download button. In table and card view, we make use of the `sizeQueries` array for this, disabling the download button until we can confirm the entity size is greater than zero (realistically, this happens quickly enough that the user does not notice). If the button is disabled, the user can move their mouse over the button to show a tooltip explaining the file is empty and thus cannot be downloaded.

## Testing instructions
This should be tested with SciGateway and both Dataview and Search. The download button appears as both a box and an icon in many places so it's important to test it enables and disables properly everywhere. If you're hooked up to the production IDS, you'll get a variety of empty and non-empty sets of data to properly test this. Make sure that every time the button is disabled, a tooltip displays over it. Also check to see how this all looks in light/dark mode and give any thoughts on that too if you have any.

- [ ] Review code
- [ ] Check Actions build

## Agile board tracking
Closes #1195

### DIFF
--- a/packages/datagateway-dataview/public/res/default.json
+++ b/packages/datagateway-dataview/public/res/default.json
@@ -211,7 +211,10 @@
   "buttons": {
     "add_to_cart": "Add to selection",
     "remove_from_cart": "Remove from selection",
-    "download": "Download"
+    "download": "Download",
+    "tooltip": {
+      "unable_to_download": "Unable to download - this file is empty"
+    }
   },
   "advanced_filters": {
     "show": "Show Advanced Filters",
--- a/packages/datagateway-dataview/public/res/default.json
+++ b/packages/datagateway-dataview/public/res/default.json
@@ -212,9 +212,7 @@
     "add_to_cart": "Add to selection",
     "remove_from_cart": "Remove from selection",
     "download": "Download",
-    "tooltip": {
-      "unable_to_download": "Unable to download - this file is empty"
-    }
+    "unable_to_download_tooltip": "Unable to download - this item is empty"
   },
   "advanced_filters": {
     "show": "Show Advanced Filters",
--- a/packages/datagateway-common/src/views/downloadButton.component.test.tsx
+++ b/packages/datagateway-common/src/views/downloadButton.component.test.tsx
@@ -60,28 +60,30 @@ describe('Generic download button', () => {
   });
 
   it('renders correctly', () => {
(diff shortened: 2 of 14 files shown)


packages/datagateway-dataview/public/res/default.json
```diff
@@ -211,7 +211,10 @@
   "buttons": {
     "add_to_cart": "Add to selection",
     "remove_from_cart": "Remove from selection",
-    "download": "Download"
+    "download": "Download",
+    "tooltip": {
```
louise-davies, line 215: Not sure on the structure of the language file here - perhaps getting rid of the `tooltip` level and just having `unable_to_download_tooltip` is clearer?

sam-glendenning: Very fair, I flip-flopped on this one too

packages/datagateway-dataview/public/res/default.json
```diff
@@ -211,7 +211,10 @@
...
+      "unable_to_download": "Unable to download - this file is empty"
```
louise-davies, line 216: Additionally, this text is confusing when it's an investigation or dataset in question. How about replacing `file` with `item`?

louise-davies:
Other than those minor points about the tooltip it seems to be working well

louise-davies (approved):

PR #1248, opened 2022-05-09
Customising notification message if it concerns session expiry #1209

### LINKED ISSUE #1209: Improve error message when session expires for anonymous user
**Description**:
At the moment, we have "Unable to find user by sessionid: <session uuid>" as a message for anonymous users when the session expires. It would be better to say "Your session has expired, please reload the page" or something similar. If useful for reporting issues, do add the sessionid too into the message, after a more user-friendly message as before.

Screenshot 2022-04-04 at 18 23 15

**Acceptance criteria**:
- [ ] Add user-friendly message when session expires

sam-glendenning:
## Description
To enable a more helpful notification message for the user, we want to amend the notification the user will receive when their session expires. Instead of displaying the returned error message from TopCAT which references a null or expired session ID which the user doesn't need to know or care about, the user is now informed that their session has expired and is instructed to reload the page to refresh their session.

## Testing instructions
Run through SciGateway. I believe the most reliable way of getting this notification to appear is to use an invalid URL for the DG API. Alternatively, see the below screenshot:

session expired

- [ ] Review code
- [ ] Check Actions build

## Agile board tracking
Closes #1209

### DIFF
--- a/packages/datagateway-common/src/handleICATError.test.ts
+++ b/packages/datagateway-common/src/handleICATError.test.ts
@@ -134,4 +134,33 @@ describe('handleICATError', () => {
       type: InvalidateTokenType,
     });
   });
+
+  it('customises notification message to SciGateway on TopCAT authentication error', () => {
+    error.response.status = 403;
+    handleICATError(error);
+
+    expect(events.length).toBe(2);
+    expect(events[0].detail).toEqual({
+      type: NotificationType,
+      payload: {
+        severity: 'error',
+        message: 'Your session has expired, please reload the page',
+      },
+    });
+
+    events = [];
+    error.response.data = {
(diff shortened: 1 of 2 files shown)


louise-davies (commented):
My only comment is that this message makes sense only in the case of anonymous/autoLogin. If the user is logged in and their session expires, then reloading the page won't really help them login again.

Maybe we need to use the fact that `scigateway` sets the `autoLogin` `localStorage` item, and display different messages to an autoLoggedIn user than we would to a logged in one?

i.e. autoLogin === true 
"Your session has expired, please reload the page"
autoLogin === false
"Your session has expired, please login again" / "Your session has expired, please re-authenticate" etc.

sam-glendenning:
That sounds like a good idea, I'll add it in

PR #1250, opened 2022-05-11
Prevent querying for cart on homepage & fix admin download table queries

louise-davies:
## Description
With the new SciGateway change to allow unauthenticated access to the homepage, it meant the user was seeing errors as the cart was requested on  the homepage. I justed jiggled the code around to create an extra component so that the cart is only requested on "actual" dataview pages.

The second change is to fix some of the querying in the admin download table - there were `UPPER`s being applied too liberally and breaking things (e.g. the sorting) so I only kept the `UPPER`s where necessary - for text filters. Also, I noticed that only setting the to field in a date filter threw an error when done against an Oracle DB due to Oracle not accepting the year being 0 (this issue is also present on TopCAT). So I set it to 1 instead which avoids the error.

## Testing instructions
- [ ] Review code
- [ ] Check Actions build
- [ ] Run dataview in scigateway and verify that no requests are sent when you're on the DG homepage
- [ ] test out the sorting and filtering in the admin download table

## Agile board tracking
No issue as these are things I found whilst doing the DLS demo so I made hotfixes for them.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -26,7 +26,6 @@ npm-debug.log*
 yarn-debug.log*
 yarn-error.log*
 
-**/public/settings.json
-**/public/datagateway-dataview-settings.json
-**/public/datagateway-download-settings.json
-**/public/datagateway-search-settings.json
+**/public/*settings*.json
+!**/public/*settings.example.json
+
--- a/packages/datagateway-dataview/src/page/pageContainer.component.test.tsx
+++ b/packages/datagateway-dataview/src/page/pageContainer.component.test.tsx
@@ -424,4 +424,11 @@ describe('PageContainer - Tests', () => {
       wrapper.find('PageBreadcrumbs').prop('landingPageEntities')
     ).toEqual(['study', 'investigation', 'dataset']);
   });
+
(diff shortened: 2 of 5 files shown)


.gitignore
```diff
@@ -26,7 +26,6 @@ npm-debug.log*
 yarn-debug.log*
 yarn-error.log*
 
-**/public/settings.json
-**/public/datagateway-dataview-settings.json
-**/public/datagateway-download-settings.json
-**/public/datagateway-search-settings.json
+**/public/*settings*.json
+!**/public/*settings.example.json
```
sam-glendenning, line 30: Bit late unfortunately but I may have just stumbled across a tiny mistake here. I tried this in OperationsGateway's .gitignore and line 29 caused the example settings files to be ignored as well, thus ignoring line 30. I believe what would work is changing line 29 to:
```
**/public/*settings.json
```
This ensures actual settings files are ignored and files ending in settings.example.json are not.

sam-glendenning: May not even need line 30 with the above fix...

sam-glendenning: I figured out my problem - I called my file **settings-example.json, not **settings.example.json. Your code is just fine! Apologies for that

PR #1252, opened 2022-05-12
Fix breadcrumbs visual glitch

### LINKED ISSUE #1196: Visual glitch in breadcrumbs
**Description**:
For some reason (subpixel issue?) sometimes the breadcrumbs' text appears with the vertical line at the front - this can be seen when resizing the window. This needs investigating.

Screenshot 2022-03-31 at 13 29 04

**Acceptance criteria**:
- [ ] Fix visual glitch in breadcrumbs

louise-davies:
## Description
I believe I fixed the error I was seeing - which was specifically in Firefox on my second monitor (it wasn't specifically due to the size of the window, just the size of the monitor weirdly). I found through trail and error that having the `borderWidth` property in the `:before` and `:after` bits of the CSS as odd was causing the subpixel issue (presumably because of some division by two not rounding right?). I've set it to use even numbers, adjusted the rest of the numbers to make everything align right and added a comment so we don't make the same mistake further down the line.

@agbeltran does this fix the issue you saw? As your visual glitch looked slightly different to mine.

## Testing instructions

- [ ] Review code
- [ ] Check Actions build
- [ ] I think Sam was able to see the issue in FF? if you run it then it shouldn't occur anymore - compare to one of the dev machines.
- [ ] TODO: test if this fixes the display error Alejandra had (could be Mac specific...)

## Agile board tracking
Closes #1196

### DIFF
--- a/packages/datagateway-dataview/src/page/breadcrumbs.component.tsx
+++ b/packages/datagateway-dataview/src/page/breadcrumbs.component.tsx
@@ -65,7 +65,7 @@ const breadcrumbsStyles = (theme: Theme): StyleRules =>
 
           /* Positions breadcrumb */
           lineHeight: '28px',
-          padding: ' 0 10px 0 20px',
+          padding: ' 0 4px 0 14px',
           textAlign: 'center',
 
           /* Add the arrow between breadcrumbs */
@@ -110,17 +110,14 @@ const breadcrumbsStyles = (theme: Theme): StyleRules =>
       },
       '& li:first-child': {
         '& a, p': {
-          paddingLeft: '16px',
-          '&:before': {
-            left: '14px',
-          },
+          paddingLeft: '14px',
         },
       },
       '& li:last-child': {
         '& a, p': {
           /* Curve the last breadcrumb border */
           borderRadius: '0 5px 5px 0',
-          paddingLeft: '20px',
+          paddingLeft: '14px',
           '&:after': {
             content: 'none',
           },
--- a/packages/datagateway-dataview/src/page/breadcrumbs.component.tsx
+++ b/packages/datagateway-dataview/src/page/breadcrumbs.component.tsx
@@ -64,52 +64,37 @@ const breadcrumbsStyles = (theme: Theme): StyleRules =>
           position: 'relative',
 
           /* Positions breadcrumb */
-          height: '28px',
           lineHeight: '28px',
-          padding: '0 5px 0 2px',
+          padding: ' 0 10px 0 20px',
           textAlign: 'center',
 
-          /* Adds between breadcrumbs */
-          marginRight: '-1px',
-          '&:before, &:after': {
+          /* Add the arrow between breadcrumbs */
+          '&:after': {
             content: '""',
             position: 'absolute',
-            top: 0,
-            border: `0 solid ${theme.palette.primary.light}`,
-            // these need to be even numbers to avoid sub-pixel issue (#1196)
-            borderWidth: '14px 6px',
-            width: 0,
-            height: 0,
-          },
-          '&:before': {
-            left: '-12px',
-            borderLeftColor: 'transparent',
-          },
-          '&:after': {
-            left: '100%',
-
-            /* Gap in between chevrons */
-            borderColor: 'transparent',
-            borderLeftColor: theme.palette.primary.light,
+            top: '3px',
+            // half the width/height
+            right: '-11px',
+            // width/height same as lineHeight - 2* top height
+            height: '22px',
+            width: '22px',
+            // change skew to alter how shallow the arrow is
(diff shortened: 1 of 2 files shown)


sam-glendenning (commented):
I've come to the conclusion that I'm only seeing this issue on Firefox and when my Windows screen magnification (not browser magnification) is greater than 100% (which is basically all the time on all my screens). I've tested this theory with what's currently in production and I'm no longer seeing the glitch when I put my magnification back to 100%. So basically, it's the screen magnification that's the problem, at least for me (which is virtually impossible to do anything about/worth doing anything about). Worth bearing in mind that resizing the browser window does _not_ cause the glitch for me.

So it's _possible_ that what you've changed here isn't even worth putting in. Having said that, I am just one user and I really think it's worth seeing what Alejandra says first because there could genuinely be an underlying problem that I'm not seeing because of my specific setup. Plus if she's running Mac, she can test Safari/Mac in general.

Here's a screenshot of prod at 100% magnification in Firefox for me:
prod, 100

Here's a screenshot of prod at 125% magnification in Firefox
prod, 125

Here's a screenshot of your changes at 100% magnification in Firefox for me:
your changes, 100

Here's a screenshot of your changes at 125% magnification in Firefox
your changes, 125

sam-glendenning:
Another quick comment - resizing the actual browser magnification can cause the issue to come back so it's clearly an issue with the calculations and numbers not rounding out nicely. Is it even possible to accommodate for all of these things?

louise-davies:
An alternative would to be to abandon the current design of the breadcrumbs which with the `:before` and `:after` pseudo-elements seems to be doomed to problems (I found the original demo that Goel used to make our breadcrumbs and it has the same issues on FF when zooming in and out): https://codepen.io/Paulie-D/pen/WvjyNM

sam-glendenning:
I feel the time may be right to retire them in favour of a different design, perhaps using a breadcrumbs library. Plus if we need breadcrumbs in operationsgateway (I can't remember if the wireframes included them or not), this would be good for that project too

louise-davies:
I may have already gotten 99% to a similar looking set of breadcrumbs using different CSS tricks that fixes the issue...

image

sam-glendenning:
Looks like you've managed to fix it! I'll wait until CI passes before approving

louise-davies:
@sam-glendenning There's still a minor issue that I'll take screenshots of - either we deal with having the issue or change the design slightly to not have the issue anymore. I'll screenshot both options in soon

louise-davies:
@sam-glendenning Here is the issue: when the breadcrumbs overflow, the "white space" in between each breadcrumb overlaps onto breadcrumbs above and below.

image

This breaks the illusion that the white spaces are actually gaps in the breadcrumbs and you can see it's just some CSS trickery. The solution is to therefore not have any "gaps" at all, so there's nothing to overflow:

image

Do you like the breadcrumbs with no gaps? It does fix the issue. Or, is the overlap on the "gaps" on overflowing breadcrumbs a minor issue that not many will notice?

sam-glendenning:
Well here's my two cents:

1. I didn't notice the issue which could indicate others won't
2. I prefer the gaps, it looks a little like a border and kind of somehow indicates to me that the breadcrumbs are clickable, at least better than not having a gap does
3. Is there anyway to dynamically change the height of the "gap" based on page magnification (I'm guessing no, or at least not easily)

Based on the above, I think we can get away with the overlapping gaps, I don't think the average user will notice or care very much.

louise-davies:
I agree with 2, and if you didn't notice when you were explicitly reviewing it then maybe it's fine.

As for 3 - the "gap" is actually the shadow of a square which overlaps the next breadcrumb along, and thus the overlapping bits are the corners of the square poking out above and below. It's nothing to do with the magnification - if you had a small window and a long breadcrumb trail you'd see the same thing - I'm just using zoom to cause the overflow.

This screenshot might help show what's going on a little...
image

Changing the height would therefore not help - you either have a "gap" big enough to create the illusion but overlaps when overflowing, or you don't overlap when overflowing but then it's too small to be a true "gap"

sam-glendenning:
In that case, I stand by my view that I like the gaps being there!

louise-davies:
Just looks like a spacing issue (we're styling the existing breadcrumbs component in MUI so probably some defaults changed) - shouldn't be too bad and I'd be happy to look at it

sam-glendenning:
For reference, Louise made further changes in https://github.com/ral-facilities/datagateway/pull/1253 (dark mode fix) and https://github.com/ral-facilities/datagateway/pull/1129/files/0d1527f57e11b1e849b47b02c8537b7aa04c4d39..d499886e4c6dd7044ede34b5dfca823ed303a937 (MUIv5 amendments)

PR #1259, opened 2022-05-23
#1174 - changed Title icon to Subject icon

### LINKED ISSUE #1174: T icon for text fields are too confusing
**Description**:
Feedback from DLS is that the T next to fields like `Title` is too confusing, so we should use another icon. If we wait for MUIv5, I think the [`ABC` icon](https://mui.com/components/material-icons/?query=abc&selected=Abc) might be good, otherwise for MUIv4 icons there's the [`TextFields` icon](https://mui.com/components/material-icons/?query=text&selected=TextFields), or other maybe the [`TextFormat` icon](https://mui.com/components/material-icons/?query=text&selected=TextFormat), or maybe they're all still confusable as actual text instead of an icon and we should go with something like [`TextSnippet` icon](https://mui.com/components/material-icons/?query=text&selected=TextSnippet)

**Acceptance criteria**:
- [ ] Use of the `Title` icon is replaced by a less confusing icon

LunaBarrett:
## Description
Changed the Title icon on data-gateway to the Subject icon

## Testing instructions

- [ ] Review code
- [ ] Check Actions build

## Agile board tracking
Closes #1174

### DIFF
--- a/packages/datagateway-common/src/card/advancedFilter.component.test.tsx
+++ b/packages/datagateway-common/src/card/advancedFilter.component.test.tsx
@@ -1,6 +1,6 @@
 import React from 'react';
 import { createShallow } from '@material-ui/core/test-utils';
-import TitleIcon from '@material-ui/icons/Title';
+import SubjectIcon from '@mui/icons-material/Subject';
 import FingerprintIcon from '@material-ui/icons/Fingerprint';
 import PublicIcon from '@material-ui/icons/Public';
 import ConfirmationNumberIcon from '@material-ui/icons/ConfirmationNumber';
@@ -171,7 +171,7 @@ describe('AdvancedFilter', () => {
     ).toEqual('advanced_filters.hide');
   });
 
-  it('TitleIcon displays correctly', () => {
+  it('SubjectIcon displays correctly', () => {
     const wrapper = shallow(
       <AdvancedFilter
         title={{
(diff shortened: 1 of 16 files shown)


packages/datagateway-common/src/card/advancedFilter.component.test.tsx
```diff
@@ -1,6 +1,6 @@
 import React from 'react';
 import { createShallow } from '@material-ui/core/test-utils';
-import TitleIcon from '@material-ui/icons/Title';
+import SubjectIcon from '@mui/icons-material/Subject';
```
sam-glendenning, line 3: Hi Luna, hope you don't mind me chiming in! I think I see the problem here and it's definitely not your fault.

Basically you've attempted to import the `SubjectIcon` using the Material UI v5 method. However, for the moment, we're still using Material UI v4, which is why all the @material-ui references in the code aren't what you'll see in the up-to-date documentation. Fortunately, the v4 docs are still available at https://v4.mui.com/components/material-icons/ which advise us to use

```
import SubjectIcon from '@material-ui/icons/Subject';
```

So if you change your references to look like the above, there should be no problems at all! For reference, we are in the middle of trying to upgrade to Material UI v5 but this won't be done for at least a month or two yet (waiting on Diamond release first)

sam-glendenning (changes requested):

sam-glendenning (approved):
Looks good! Go ahead and merge it when you're ready

PR #1266, opened 2022-05-27
Bugfix/disable download button #1261

### LINKED ISSUE #1261: Disable HTTPS download button if status is not complete
**Description**:
A small bug - for HTTPS downloads the download button is always clickable, even though it should be disabled if the download is not in the `COMPLETE` state.

**Acceptance criteria**:
- [ ] HTTPS download button only enabled when download status is `COMPLETE`

louise-davies:
## Description
Essentially, I just added a check for whether a download is complete to enable the download button. I also added an extra tooltip message to differentiate from the non-http(s) download message. Also removed the "not-downloadable" aria-label on the button itself as in that case the tooltip is there which labels the button as undownloadable.

## Testing instructions

- [ ] Review code
- [ ] Check Actions build
- [ ] scigateway-preprod currently has a HTTPS download in non-COMPLETE state in the root user's downloads, which is what I used for testing

## Agile board tracking
Closes #1261


packages/datagateway-download/src/downloadStatus/downloadStatusTable.component.tsx
```diff
@@ -300,59 +300,56 @@ const DownloadStatusTable: React.FC<DownloadStatusTableProps> = (
...
                   <BlackTooltip
                     title={
-                      t('downloadStatus.download_disabled_tooltip', {
-                        transport: downloadItem.transport,
-                      }) as string
+                      !isHTTP
+                        ? (t(
+                            'downloadStatus.non_https_download_disabled_tooltip',
+                            { transport: downloadItem.transport }
+                            // for some reason it can't infer these types on its own
```
sam-glendenning, line 321: I'm guessing because the `title` is `NonNullable`, which excludes null and undefined and `useTranslation` can return undefined? 🤷 The solution works though!

sam-glendenning (approved):

packages/datagateway-download/src/downloadStatus/downloadStatusTable.component.tsx
louise-davies: If I called the translation function outside the return and stored it in a variable then I could use that, hence I don't know why it didn't like it 😅

Plus the previous value of `title` was also a translation....

PR #1283, opened 2022-06-15
Improve handling of undefined carts

### LINKED ISSUE #1280: Improve network error message and how to handle cart request failed?
**Description**:
Based on feedback from a DLS user, they were presented with this screen (blur applied for data privacy):

Screenshot 2022-06-06 10 28 02-blurred

Their requests to TopCAT to fetch their cart failed due to low level network errors (i.e. TopCAT didn't recieve any of their requests). We should improve the Axios default `Network Error` to indicate a course of action to the user e.g. refresh the page to try again, try again later, contact support etc.

Additionally, the checkboxes column didn't appear because the `selectedRows` prop passed to the `table` component was `undefined` and thus it doesn't render - should we change this behaviour? If we do, do we just remove the `selectedRows !== undefined` check and assume all the checkboxes should be empty? Should we display the checkboxes but disable them all to indicate that something is very wrong?

**Acceptance criteria**:
- [ ] Resolve the issues raised by the DLS user

louise-davies:
## Description
Added a more friendly error message for network errors, as well as added logic to display the checkboxes even when the cart isn't defined and instead make things disabled & have tooltips.

Also noticed that there was some repeating styling of the base MUI tooltip so I just made one StyledTooltip for use in common

## Testing instructions

- [ ] Review code
- [ ] Check Actions build
- [ ] set your topcat url to nonsense to see the disabled checkboxes, to simulate a network error idk...

## Agile board tracking
Closes #1280


packages/datagateway-dataview/src/views/table/datafileTable.component.tsx
```diff
@@ -161,7 +161,7 @@ const DatafileTable = (props: DatafileTableProps): React.ReactElement => {
 
   return (
     <Table
-      loading={addToCartLoading || removeFromCartLoading}
+      loading={addToCartLoading || removeFromCartLoading || cartLoading}
```
sam-glendenning, line 164: Is it worth adding an `allIdsLoading` as well? Since `allIds` is used in the `SelectHeader` onClick function and it can be undefined, this may lead to the `SelectHeader` being enabled when it doesn't have `allIds`

packages/datagateway-dataview/src/views/table/isis/isisDatasetsTable.component.test.tsx
```diff
@@ -93,6 +93,7 @@ describe('ISIS Dataset table component', () => {
 
     (useCart as jest.Mock).mockReturnValue({
       data: [],
+      isLoading: false,
```
sam-glendenning, line 96: You forgot to add `cartLoading` to the `isisDatasetsTable` component

packages/datagateway-dataview/src/views/table/datasetTable.component.test.tsx
```diff
@@ -253,6 +254,7 @@ describe('Dataset table component', () => {
           parentEntities: [],
         },
       ],
+      isLoading: false,
```
sam-glendenning, line 257: There's another `useCart` mock function that needs `isLoading: false`, on line 274 below this

packages/datagateway-download/src/downloadStatus/downloadStatusTable.component.tsx
```diff
@@ -315,14 +315,14 @@ const DownloadStatusTable: React.FC<DownloadStatusTableProps> = (
                   <BlackTooltip
                     title={
                       !isHTTP
-                        ? (t(
+                        ? t<string, string>(
                             'downloadStatus.non_https_download_disabled_tooltip',
                             { transport: downloadItem.transport }
                             // for some reason it can't infer these types on its own
```
sam-glendenning, line 321: This comment can be removed now

packages/datagateway-search/src/table/investigationSearchTable.component.test.tsx
```diff
@@ -127,6 +127,7 @@ describe('Investigation Search Table component', () => {
     ];
     (useCart as jest.Mock).mockReturnValue({
       data: [],
+      isLoading: false,
```
sam-glendenning, line 130: This line is fine, I just couldn't comment on the line I wanted.

Just for maintainability, in the `afterEach` on line 183, I'd change this all to `jest.clearAllMocks()` to make sure we're not missing anything, instead of calling `mockClear` line after line. If you ever come across any others like this, I'd do the same, it's just easier down the line

packages/datagateway-common/src/table/headerRenderers/selectHeader.component.tsx
```diff
@@ -39,17 +39,21 @@ const SelectHeader = React.memo(
       >
         <Checkbox
```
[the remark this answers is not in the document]
louise-davies: There wasn't any - I can pretty easily add one 😅

packages/datagateway-dataview/src/views/table/datafileTable.component.tsx
louise-davies: allIds already has a fallback where it instead just uses the currently loaded `data` - although it's probably better to just disable it if allIds hasn't loaded. I'll add it to the loading stuff alongside `cartLoading`

sam-glendenning:
It's looking good but unfortunately the select header has been a bit messed up by the tooltip addition. The checkbox is no longer vertically centred, which causes this:

image

Looks like the `<span>` in the select cell component doesn't work here. But removing it totally means the tooltip doesn't render and the checkbox is slightly off centre:

image

louise-davies:
@sam-glendenning we'll leave this PR until after DLS release now - maybe we should just revert and have no tooltip on the select header but I'll try and fix the solution with the `span` - I expect it's just parts of the styling aren't being properly inherited.

sam-glendenning (approved):

PR #1284, opened 2022-06-16
Implement best practices in Docker image 

VKTB:
## Description
This PR implements best practices in the Docker image.

In summary, the changes do the following:
- Removes the lines that sets the `PATH` because the image builds and the container works fine without it.
- Combines the `RUN` commands so that there are less layers.
- Configures the image so that containers run using the non-root `www-data` user account that the `httpd` image defines.
- Uses `setcap` to allow non-root user accounts to bind to privileged ports (port 80 in this case) as these ports are only permitted to `root` by default. I found this solution in [this](https://takac.dev/docker-run-apache-as-non-root-user-based-on-the-official-image/) article.
- Changes the access rights for the Apache's `logs` directory from `root` to `www-data` user account.

I also tried to leverage build cache in the building stage to reduce the build time but I could not achieve this. I faced some `lerna` issues which I could not resolve.

## Testing instructions
- [ ] Review code
- [ ] Check Actions build
- [ ] Image builds successfully
- [ ] Container starts successfully

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,12 @@
+/.github
+/.husky
+/packages/datagateway-*/cypress
+/packages/datagateway-*/cypress.json
+/packages/datagateway-*/server
+/packages/datagateway-*/README.md
+.gitignore
+.prettierrc
+codecov.yml
+CODEOWNERS
+LICENSE.md
+README.md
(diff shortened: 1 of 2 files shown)


moonraker595 (approved):
Looks good 👍

Issue #1328, opened 2022-07-14
Add react-testing-library so we can start migrating tests over


louise-davies:
**Description**:
Add a description here

**Acceptance criteria**:
- [ ] A checklist
- [ ] of criteria to accept this story as done

louise-davies:
This has been added

PR #1352, opened 2022-07-21
Migrate dg-download to `react-query` and `@testing-library`

### LINKED ISSUE #1312: Migrate dg-download to use react-query
**Description**:
See title, also see #1308

**Acceptance criteria**:
- [ ] Admin download table, user download table & download confirm modal all use react-query instead if directly calling

kennethnym:
## Description
This PR includes refactoring of dg-download to use `react-query` and `@testing-library` fully.

### Changes

- I have tweaked the behavior of filtering in admin status table slightly. I added a debounce to table filtering requests in order to prevent excessive calls. Without the debounce, a new request is made on every keystroke, which is not ideal. I can easily extract the debounce logic into a reusable `useDebounce` hook, but unless debounce will be used in other places, I don't think it is necessary.
- Before migration to `react-query`, upon refreshing tables, the data will be set to an empty array so that there is visually a refresh. Since we no longer have control of the data (`react-query` manages them in the internal cache), I cannot replicate the behavior without spaghetti of `useEffect` and `useState`. I don't know if it is worth it considering a "Refreshing..." will be shown anyway when the table is refreshing.
- I have added `role="progressbar"` to `Preloader` so that it is more accessible and also makes it easier to test.

## Testing instructions
- [ ] Review code
- [ ] Check Actions build

## Agile board tracking
connect to #1312, #1328 

Closes #1312


packages/datagateway-download/src/downloadApi.test.ts
```diff
@@ -55,7 +56,7 @@ describe('Download Cart API functions test', () => {
         return Promise.resolve({
           data: {
             facilityName: mockedSettings.facilityName,
-            userName: 'test user',
+            rName: 'test user',
```
louise-davies, line 59: slight typo?

packages/datagateway-download/src/downloadApi.ts
```diff
@@ -98,23 +95,26 @@ export const submitCart: (
       params
     )
     .then((response) => {
-      log.debug(response);
-
       // Get the downloadId that was returned from the IDS server.
+      console.log('response', response);
```
louise-davies, line 99: can this go back to either being a `loglevel` call or be removed all together?

packages/datagateway-download/src/downloadApi.ts
```diff
@@ -98,23 +95,26 @@ export const submitCart: (
...
     });
 };
 
+/**
+ * Settings for {@link fetchDownloads}
+ */
+export interface FetchDownloadsSettings {
+  facilityName: string;
+  downloadApiUrl: string;
+}
```
louise-davies, line 111: It seems like other functions in this file use this same type for the settings - therefore could this be renamed to be more generic and used for other functions like `getDownload`, `getDownloadStatusType` etc

packages/datagateway-download/src/downloadApiHooks.ts
```diff
@@ -121,43 +179,62 @@ export const useIsTwoLevel = (): UseQueryResult<boolean, AxiosError> => {
...
+        fileName,
+        {
+          facilityName,
+          downloadApiUrl,
+        },
+        zipType
+      ),
+    {
+      onError: (error, _, rollback) => {
+        console.log('error', error);
```
louise-davies, line 225: We don't need to log the error here - `handleICATError` already logs the error message

packages/datagateway-download/src/downloadApi.ts
kennethnym: I actually tried to do that, but other functions take in settings with slightly different shapes, (e.g. `getDataFileCount` takes `{ apiUrl: string }`). Although we can use the `Pick` util type to pick the keys we want from the settings file.

packages/datagateway-download/src/downloadApi.ts
kennethnym: This is just for debugging purpose, I'll delete it.

packages/datagateway-download/src/downloadApi.test.ts
kennethnym: Yeah it is, will fix.

packages/datagateway-download/src/downloadApiHooks.ts
```diff
@@ -250,3 +327,434 @@ export const useDatafileCounts = (
...
+
+export interface UseDownloadDeletedParams {
+  downloadId: number;
+  deleted: boolean;
+}
+
+/**
+ * A React query that provides a mutation for deleting a download item.
+ */
+export const useDownloadDeleted = (): UseMutationResult<
```
louise-davies, line 413: Minor comment - but in my head it makes more sense to name mutations after action they perform i.e. this would be `useDeleteDownload` as it implies that this hook can be used to delete downloads.

kennethnym: It was actually named `useDeleteDownload` originally, but then I realized it could also be used to restore a download, so I renamed it to `useDownloadDeleted`. Maybe `useDeleteOrRestoreDownload`?

packages/datagateway-download/src/downloadApiHooks.ts
```diff
@@ -250,3 +327,434 @@ export const useDatafileCounts = (
...
+        downloadApiUrl: downloadSettings.downloadApiUrl,
+      }),
+    onSettled: (_, error) => {
+      loadedQueriesCount.current += 1;
+      if (error) handleQueryError(type);
+    },
+    ...queryOptions,
+  }));
+
+  // TODO: un-ignore ts when upgrading react-query to >=3.28
```
louise-davies, line 558: I believe if you merge develop we have now upgraded the react-query version to the lastest v3

packages/datagateway-download/src/downloadApiHooks.ts
```diff
@@ -250,3 +327,434 @@ export const useDatafileCounts = (
...
+  >[] = downloadTypes.map((type) => ({
+    queryKey: [QueryKey.DOWNLOAD_TYPE_STATUS, type],
+    queryFn: () =>
+      getDownloadTypeStatus(type, {
+        facilityName: downloadSettings.facilityName,
+        downloadApiUrl: downloadSettings.downloadApiUrl,
+      }),
+    onSettled: (_, error) => {
+      loadedQueriesCount.current += 1;
+      if (error) handleQueryError(type);
```
louise-davies, line 553: I believe moving the error handling code into the `onSettled` call changes the original behaviour - before it waited until all the queries returned, and then either sent the `downloadConfirmDialog.access_methods_error` or sent individual `downloadConfirmDialog.access_method_error`. Now, we'll end up sending `downloadConfirmDialog.access_method_error` when each request fails, and then finally send `downloadConfirmDialog.access_methods_error` once they've all returned in an error state. This is potentially problematic as it could spam the user with lots of error messages if they're all unable to be fetched. I wonder if this is therefore better handled in a `useEffect` in `downloadConfirmDialog`...

packages/datagateway-download/src/downloadApiHooks.test.tsx
```diff
@@ -542,4 +739,762 @@ describe('Download Cart API react-query hooks test', () => {
       );
     });
   });
+
+  describe('useDownloads', () => {
+    it('should retrieve user downloads', async () => {
+      const mockFormattedDownloadItems: FormattedDownload[] = [
```
louise-davies, line 745: This is identical to `mockFormattedDownloadItems` set above - maybe it's worth changing the earlier assignments into `let`s and setting them to their defaults in a `beforeEach` and then modify the data as you require in each test? We can definitely shave off some lines of repeated mock definitions ;)

packages/datagateway-download/src/downloadApiHooks.test.tsx
```diff
@@ -542,4 +739,762 @@ describe('Download Cart API react-query hooks test', () => {
...
+      );
+      expect(handleICATError).toHaveBeenCalledWith({
+        message: 'Test error message',
+      });
+    });
+  });
+
+  describe('useDownloadDeleted', () => {
+    it('should delete download with given id and update the download list upon success', async () => {
+      const mockDownloadItems: Download[] = [
```
louise-davies, line 888: Again, couldn't this just reference the global `mockDownloadItems` like `[mockDownloadItems[0]]` to get a single download item mock?

packages/datagateway-download/src/downloadApiHooks.test.tsx
```diff
@@ -542,4 +739,762 @@ describe('Download Cart API react-query hooks test', () => {
...
+        mockFormattedDownloadItems,
+      ]);
+    });
+  });
+
+  describe('useSubmitCart', () => {
+    it('should submit cart and clear cart on success', () => {
+      axios.put = jest.fn().mockResolvedValue(123);
+    });
+  });
```
louise-davies, line 1427: Seems there's a test missing here? We mock `axios` but there's no call of `useSubmitCart` or any assertations

packages/datagateway-download/src/downloadStatus/adminDownloadStatusTable.component.tsx
```diff
@@ -115,110 +110,72 @@ const AdminDownloadStatusTable: React.FC = () => {
...
+    // debounced to avoid excessive calls
+    refetchTimeout.current = setTimeout(() => {
+      refetch();
+    }, 200);
+
+    return () => {
+      const prevTimeout = refetchTimeout.current;
+      if (prevTimeout) clearTimeout();
+    };
+  }, [refetch, filters]);
```
louise-davies, line 162: This code shouldn't be necessary - as `TextFilter` already debounces it's input - see https://github.com/ral-facilities/datagateway/blob/6292c13ff7b41a5f31abad669e97cf29551b3852/packages/datagateway-common/src/table/columnFilters/textColumnFilter.component.tsx#L21.

packages/datagateway-download/src/downloadApiHooks.ts
kennethnym: The `handleQueryError` waits until `loadedQueriesCount` matches the query count to either send individual errors or broadcast the `'downloadConfirmDialog.access_methods_error'`:

```js
if (loadedQueriesCount.current === queryCount) {
}
```

louise-davies: Ah, I missed this so that's good :)

packages/datagateway-download/src/downloadApiHooks.test.tsx
kennethnym: Ahaha yeah I missed it, sorry about that.

packages/datagateway-download/src/downloadStatus/adminDownloadStatusTable.component.tsx
kennethnym: Interesting - the test times out without the debounce, I'll investigate further.

packages/datagateway-download/src/downloadApiHooks.test.tsx
kennethnym: Yeah good point, when I moved the mock items out I must have left out some local variables.

packages/datagateway-download/src/downloadConfirmation/dialogContent.component.ts
```diff
@@ -0,0 +1,7 @@
+import { DialogContent as MuiDialogContent, styled } from '@mui/material';
+
+const DialogContent = styled(MuiDialogContent)(({ theme }) => ({
+  padding: theme.spacing(2),
+}));
+
+export default DialogContent;
```
louise-davies, line 7: Minor comment but does this really need to be in it's own file? I feel as though it's place in `downloadConfirmDialog` was fine as it's such a small component - essentially just applying some styles. `dialogTitle` I understand as it is tested separately and a little bigger...

louise-davies: Ah, I see - it's because it's used by both `downloadConfirmDialog` and `downloadRequestResult` - nevermind this comment then!

kennethnym: Since I needed it in both `DownloadConfirmDialog` and `DownloadRequestResult`, I thought if I was moving `DialogTitle` to its own file, might as well do the same for `DialogContent`, because otherwise I'd end up with `DialogTitle` in a separate file but `DialogContent` being exported by `downloadConfirmDialog.component`, and it just feels weird to me. What do you think?

packages/datagateway-download/src/downloadConfirmation/downloadConfirmDialog.component.tsx
```diff
@@ -438,365 +324,274 @@ const DownloadConfirmDialog: React.FC<DownloadConfirmDialogProps> = (
...
-                          if (method.message) {
-                            return method.message;
-                          } else {
-                            return t(
-                              'downloadConfirmDialog.access_method_helpertext_disabled_error'
+                    {/* Access methods from settings as items for selection */}
+                    {sortedDownloadTypes &&
+                      sortedDownloadTypes.map(
+                        ({ type, ...methodInfo }, index) => {
+                          if (methodInfo.disabled !== undefined) {
```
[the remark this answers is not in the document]
kennethnym: I don't think there's any reason, I think it was like that already when I was making the changes. I think your snippet makes more sense anyway.

louise-davies (approved):
Looks good and seems to function similar to previously - we'll just need to test it extensively when aiming to release.

PR #1371, opened 2022-07-27
Add progress bar UI to download status tables to show download progress

### LINKED ISSUE #1308: Add progress indicators to downloads if the IDS supports it (ids.rd2foo)
**Description**:
Kevin's `ids.r2dfoo` exposes [getPercantageComplete](http://nexus.esc.rl.ac.uk:8081/nexus/content/sites/site/ids.r2dfoo/1.0.0/miredot/index.html#1826526913). This accepts `preparedId` and returns a string that is either parsable as an int `0-100`, or a string status indicating "UNKNOWN".

If the string is int parsable, we should render a progress bar with relevant percentage indicated, and if it's not parsable as an int we should just display the status message. We should only render a progress bar if the download is in the `RESTORING` state (double check this - we definitely don't need to render for `COMPLETE` and `EXPIRED` but are there other states?). We also need to test the case of a `preparedId` has been deleted from the topcat db and so `getPercantageComplete` returns 404.

question: Kevin says that a download can be 100% complete via `getPercantageComplete` but the download Availability is still not `COMPLETE` - this is due to pollcat delays etc. Should we do something to explain this to the user (add extra `Finalising` status to availability column?, add extra state to progress bar saying finalising, or do nothing and leave users to it)

This should also get added to the admin download table - we should investigate whether a bar is too wide for that table and whether to just display the percentage (ask Chris).

**Acceptance criteria**:
- [ ] Add extra column before the Availability column which displays progress bars

kennethnym:
## Description
This PR adds a new progress bar UI to both download status tables to show the progress of a download item. If the server returns a status string rather than a progress percentage, then the status string is shown instead.

I added a new entry in the settings file called `uiFeatures` which allow not only the progress UI to be toggled, but any future UI elements/features that need to be togglable:

```json
{
  "uiFeatures": {
    "downloadProgress": true
  }
}
```

It is an array of string flags that contains feature flags. The flags can then enable certain UI features, in this case download progress.

This should be merged after #1352 

## Testing instructions

- [ ] Review code
- [ ] Check Actions build

## Agile board tracking
connect to #1308 

Closes #1308


packages/datagateway-download/public/datagateway-download-settings.example.json
```diff
@@ -62,5 +62,6 @@
       "order": 1
     }
   ],
-  "pluginHost": "http://localhost:3000"
+  "pluginHost": "http://localhost:3000",
+  "uiFeatures": ["DOWNLOAD_PROGRESS"]
```
louise-davies, line 66: A very minor point, but since this is JSON & Javascript normally the convention is `camelCase` and so `downloadProgress` may make more sense. Also, I dislike that this is just an array which needs a specific string - would an object which has boolean properties be clearer and less likely for someone to mess up by mispelling something?
i.e.

```
"uiFeatures": {
  "downloadProgress": true
}
```

If we did do things like that - I would then suggest that `downloadProgress` be disabled by default as this is a DLS only feature

packages/datagateway-download/src/downloadApi.ts
```diff
@@ -362,3 +362,28 @@ export const getDataUrl = (
...
+  settings: { idsUrl },
+}: {
+  preparedId: string;
+  settings: { idsUrl: string };
+}): Promise<DownloadProgress> => {
+  const { data } = await axios.get(`${idsUrl}/getPercentageComplete`, {
+    params: { preparedId },
+  });
+  // if data is not a number (NaN), it is a status value
+  const isStatus = isNaN(data);
```
louise-davies, line 387: I think `isNaN` is a bit unreliable - it has some odd results (`isNaN("")` is false, so is `isNaN([])`. I think the below code is more robust.

```
const numberData = parseInt(data); // or parseFloat - does the endpoint only return integers?
return Number.isNaN(numberData) ? data : numberData;
```

packages/datagateway-download/src/downloadStatus/downloadProgressIndicator.component.tsx
```diff
@@ -0,0 +1,66 @@
...
+} from '../downloadApiHooks';
+import { useTranslation } from 'react-i18next';
+
+interface DownloadProgressIndicatorProps {
+  downloadId?: number;
+}
+
+/**
+ * A component for showing the download progress of the given download item.
+ * @param download The download item that this component should show the progress of.
```
louise-davies, line 15: This comment doesn't accurately portray that it takes a `downloadId`, not a `download`

packages/datagateway-download/public/datagateway-download-settings.example.json
kennethnym: Since it is a JSON file, I don't think the IDE would provide suggestion for the key anyways (unless we create a JSON schema). In TypeScript, I typed the array as an array of a union string, so in terms of autocompletion I think both are the same.

packages/datagateway-download/src/downloadStatus/downloadProgressIndicator.component.tsx
```diff
@@ -0,0 +1,66 @@
...
+  } = useDownloadPercentageComplete({
+    preparedId: download?.preparedId ?? '',
+    enabled: Boolean(download?.preparedId) && download?.status === 'RESTORING',
+  });
+
+  console.log(
+    'load progress',
+    Boolean(download?.preparedId) && download?.status === 'RESTORING'
+  );
+  console.log('status', download?.status);
```
louise-davies, line 38: Remove console.log statements

packages/datagateway-download/src/downloadStatus/downloadProgressIndicator.component.tsx
```diff
@@ -0,0 +1,66 @@
...
+/**
+ * A component for showing the download progress of the given download item.
+ * @param download The download item that this component should show the progress of.
+ * @constructor
+ */
+function DownloadProgressIndicator({
+  downloadId,
+}: DownloadProgressIndicatorProps): JSX.Element {
+  const [t] = useTranslation();
+  const { data: download, isLoading: isLoadingDownload } = useDownload({
```
louise-davies, line 22: Why do we requery for the download here? Could we not just pass the required props from the status tables to save us requerying for all the downloads in the table?

packages/datagateway-download/src/downloadStatus/downloadProgressIndicator.component.tsx
```diff
@@ -0,0 +1,66 @@
...
+  if (typeof progress === 'number')
+    return (
+      <Box sx={{ display: 'flex', alignItems: 'center' }}>
+        <Box sx={{ width: '100%', mr: 1 }}>
+          <LinearProgress variant="determinate" value={progress} />
+        </Box>
+        <Box sx={{ minWidth: 36 }}>
+          <Typography role="" variant="body2" color="text.secondary">
+            {progress}%
+          </Typography>
```
louise-davies, line 58: I thought that `LinearProgress` can be rendered with a label? https://mui.com/material-ui/react-progress/#linear-with-label

kennethnym: If you open the code example, `LinearProgressWithLabel` is actually a custom component.

packages/datagateway-download/public/datagateway-download-settings.example.json
louise-davies: Yes, but from the perspective from a sysadmin who won't necessarily understand the source code - if they had a settings file with `"uiFeatures: []"` - how would they know what strings to put in? If this was `"uiFeatures": {"downloadProgress":false}` - they'd know to merely flip the boolean

packages/datagateway-download/src/downloadStatus/downloadProgressIndicator.component.tsx
louise-davies: Ah okay - nevermind! I assumed it was a prop 😅

packages/datagateway-download/src/downloadStatus/downloadProgressIndicator.component.tsx
kennethnym: The problem is that the data in the table are formatted (`FormattedDownload`), which makes it awkward to work on the status of the download. In theory I can use a combination of `t` from useTranslation and translation key and use it to check the download status, but it feels wrong.

packages/datagateway-download/public/datagateway-download-settings.example.json
kennethnym: Ahh, I see what you mean now. I agree with you, I'll change it to an object type.

packages/datagateway-download/src/downloadStatus/downloadProgressIndicator.component.tsx
louise-davies: It feels like that would be more efficient than sending out more queries though right? Alternatively, we could do something like not use a select function to transform the downloads and instead do it in a separate variable (like what happens in `adminDownloadStatusTable` and then do something like this in the `cellContentRenderer`:

```
cellContentRenderer: ({
   rowData,
}: TableCellProps) => (
  <DownloadProgressIndicator
    download={data?.pages.flatMap((page) => page).find((download) => download.id === (rowData as FormattedDownload).id)}
  />
)
```

Also, couldn't you just make a hook that does the reverse of `formatDownload` - i.e. to transform from `FormattedDownload` to `Download` which you can use to abstract the `useTranslation` stuff and use the hook in `downloadProgressIndicator`?

kennethnym: Yeah, when I was writing that piece of code I was thinking calling `useDownload` there should have been a cache hit because `DownloadConfimDialog` also calls it, but then I just realized the table could've been loaded before the dialog (so obvious but I missed it).

What about a map that maps download IDs to `Download`, so that we can just look up the map for the unformatted `Download` when it is needed?

packages/datagateway-download/src/downloadApi.ts
kennethnym: I can use `parseFloat` for now, it is easy to switch to `parseInt` if needed.

packages/datagateway-download/src/downloadStatus/downloadProgressIndicator.component.tsx
louise-davies: I don't see why the reverse of `formatDownload` is awkward, as we define both sides of the equation (theoretically there could be weird issues if the syadmin had duplicate string values in the translations but that's a niche concern I think). If you think having a map is more useful then go for it. A third alternative is to create a new field in `FormattedDownload` that's like `systemStatus` which keeps the original form of the status enum? For usages like this?

kennethnym: > theoretically there could be weird issues if the syadmin had duplicate string values in the translations but that's a niche concern I think
That is exactly what I am thinking about! What if the status values map to the same user-facing label? But if it is not a concern, then it will be way easier to just do a reverse of `formatDownload`.

kennethnym: I think I prefer adding an extra field to `FormattedDownload` - I was thinking adding a field called `formattedStatus` that stores the, well, formatted status, and the `status` will store the original value.

kennethnym: I can do the same for `isDeleted`.

louise-davies (commented):
The code for this looks good - lets just have it tested next week & get feedback from the DLS admins

packages/datagateway-download/public/datagateway-download-settings.example.json
louise-davies, line 66: The example settings file & e2e settings haven't been updated to the new format

packages/datagateway-download/src/ConfigProvider.tsx
```diff
@@ -23,9 +23,19 @@ export interface DownloadSettings {
   routes: PluginRoute[];
   helpSteps: { target: string; content: string }[];
   pluginHost?: string;
+
+  /**
+   * A map of UI flags that can toggle certain UI features..
+   */
+  uiFeatures: {
```
louise-davies, line 30: I don't know whether it's worth making this optional & just assuming false in that case for all flags - when I was deploying I forgot to update the settings file and it thus threw an exception...

Maybe this isn't a concern though as once it's in the new settings file it's considered a "standard" config value

packages/datagateway-download/src/downloadApiHooks.ts
```diff
@@ -736,3 +738,35 @@ export const useAdminUpdateDownloadStatus = (): UseMutationResult<
...
+export const useDownloadPercentageComplete = <T = DownloadProgress>({
+  preparedId,
+  ...queryOptions
+}: { preparedId: string } & UseQueryOptions<
+  DownloadProgress,
+  AxiosError,
+  T,
+  string[]
+>): UseQueryResult<T, AxiosError> => {
+  const { idsUrl } = React.useContext(DownloadSettingsContext);
```
louise-davies, line 756: Just had a thought - each download could have different IDSes and so we can't just get the "general" `idsUrl` from the settings file - we need the one corresponding to the download. We'll probably have to match the `transport` item in a download to the access methods array in the settings to get the right URL.

packages/datagateway-download/src/downloadApiHooks.ts
```diff
@@ -736,3 +738,35 @@ export const useAdminUpdateDownloadStatus = (): UseMutationResult<
...
+  return useQuery(
+    [QueryKey.DOWNLOAD_PROGRESS, preparedId],
+    () =>
+      getPercentageComplete({
+        preparedId: preparedId,
+        settings: { idsUrl },
+      }),
+    {
+      onError: (error) => {
+        handleICATError(error);
```
louise-davies, line 767: I'll need to check with Chris & Sabah whether they want progress request errors to be broadcast to the user or not (I've just asked Chris just waiting on a response)

louise-davies:
@kennethnym I confirmed with Sabah that they're happy with this feature and they've been testing it pretty extensively. Once merge conflicts are resolved this can go in :)

packages/datagateway-download/public/res/default.json
```diff
@@ -34,7 +34,11 @@
     "expired": "Expired",
     "paused": "Paused",
     "preparing": "Preparing",
-    "restoring": "Restoring from tape"
+    "restoring": "Restoring from tape",
+    "progress": "Progress",
+    "calculating_progress": "Calculating...",
+    "progress_unavailable": "Unavailable",
```
louise-davies, line 40: A minor comment from yesterday was that `Unavailable` was confusing and they think perhaps `Unknown` is better. Might as well change the default before it merges in

louise-davies (approved):
Looks good and was well tested on DLS preprod :)

PR #1384, opened 2022-08-08
Fix details panel layout issue across mounts

### LINKED ISSUE #1331: Detail panel tab resets when scrolling
**Description**:

Expand an entity (checked it for datafiles and instruments but assume it'll happen for all entities?)
image

Swap detail panel tabs
image

Scroll down so the panel is out of sight
image

When scrolling back up, the tab has reset to the first but still has the height of the second. This either leeds to information being cut off as below, or loads of dead space if the second tab was taller.
image

**Acceptance criteria**:
- [ ] Scrolling back to an expanded details panel doesn't switch tabs (or at least the height is correct when it does?)

kennethnym:
## Description
This PR addresses the issue where details panel forgets their state when being unmounted by `react-virtualized`, but their container remembers their height, which causes mismatch between the displayed tab and the container height. Instead of storing the selected tabs locally in the panel component, it is stored in the `dgcommon` redux store instead.

I have migrated all the affected tests to `@testing-library` as well.

**PS:** I think it is now better to store the heights of the panels in the `dgcommon` store as well instead of passing `detailsPanelResize` prop to each panel component.

## Testing instructions
- [ ] Review code
- [ ] Check Actions build

## Agile board tracking
connect to #1331 

Closes #1331


packages/datagateway-common/package.json
```diff
@@ -45,8 +45,10 @@
   "devDependencies": {
     "@mui/icons-material": "5.5.0",
     "@mui/material": "5.5.0",
+    "@testing-library/jest-dom": "^5.16.4",
     "@testing-library/react": "12.1.3",
     "@testing-library/react-hooks": "8.0.1",
+    "@testing-library/user-event": "^14.4.1",
```
sam-glendenning, line 51: Could these dependencies be pinned? Worth doing because renovate will make a whole PR just to do this anyway if this were to get merged like this

kennethnym: Yep, will fix.

packages/datagateway-common/src/detailsPanels/isis/instrumentDetailsPanel.component.tsx
```diff
@@ -46,8 +90,8 @@ const InstrumentDetailsPanel = (
         textColor="secondary"
         indicatorColor="secondary"
         scrollButtons="auto"
-        value={value}
-        onChange={(event, newValue) => setValue(newValue)}
+        value={selectedTab ?? DEFAULT_TAB}
```
sam-glendenning, line 93: I agree with having the `useEffect` on line 77 but I'm worry of a slight disconnect if we have 
```
selectedTab ?? DEFAULT_TAB
```
here. Just on the off chance (I'm sure it's unlikely) that the `value` here is incorrect because it's out of sync with `selectedTab`? Idk, maybe it doesn't matter but I feel that if we just have 
```
value={selectedTab}
```
here, then we're always in sync with what's in the store and we haven't got to worry. Feel free to dissuade me on this!

kennethnym: Unfortunately MUI doesn't allow `undefined` value. `selectedTab` can only be `undefined` when the details panel is first rendered (and has *never* been rendered before). Upon first render the store will be synced with the selected tab (always `DEFAULT_TAB` for initial render).

sam-glendenning: Oh of course, I knew there must be a little more to it than that. Thanks for letting me know :)

packages/datagateway-common/src/detailsPanels/dls/visitDetailsPanel.component.test.tsx
```diff
@@ -1,31 +1,46 @@
...
+function renderComponent({
+  rowData,
+  detailsPanelResize,
+}: {
+  rowData: Investigation;
+  detailsPanelResize?: () => void;
+}): RenderResult {
+  return render(
+    <Provider
+      store={createStore(
```
sam-glendenning, line 24: I wouldn't say it's a big deal right now but certainly something to consider as we move further down the line of RTL migration. Since `createStore` is deprecated, Louise figured out a good way of [making use of the newer method](https://github.com/ral-facilities/operationsgateway/blob/aefde7d4e3d9c11b2339bbb58eafcb70eb041330/src/state/store.tsx#L19), `configureStore`, on OperationsGateway that also reduces code repetition. It will be worth adding this at some point but I don't think it's necessary until we decide to properly migrate everything over.

kennethnym: Is `configureStore` from redux toolkit compatible with "legacy" redux architecture which is what `datagateway-dataview` is using at the moment?

kennethnym: It seems to be, might be worth changing?

packages/datagateway-common/src/detailsPanels/dls/visitDetailsPanel.component.test.tsx
```diff
@@ -47,25 +62,26 @@ describe('Visit details panel component', () => {
...
-    (useInvestigationSize as jest.Mock).mockReturnValue({
-      data: 1,
-      refetch: jest.fn(),
+    axios.get = jest.fn().mockImplementation((url: string) => {
+      if (/.*\/investigations/.test(url))
+        return Promise.resolve({ data: [rowData] });
+
+      if (/.*\/user\/getSize/.test(url)) return Promise.resolve({ data: 64 });
+
+      return Promise.resolve();
```
sam-glendenning, line 71: I'm personally not convinced these mock implementations _need_ to change. The advantages of the original method were:
1. we can make sure we are stubbing the correct method, without having to resort to regex to ensure the right endpoints with the right return data. Someone looking at this file for the first time might be confused as to what is being done here when `axios` itself isn't actually being imported or used in the `visitDetailsPanel` component
2. by mocking the functions themselves, we don't need to care about the _implementation_ of those functions. As far as `visitDetailsPanel` is concerned, it's just calling a function that returns some data and `axios` isn't involved at all. If we did end up changing the implementation of `useInvestigationDetails` or `useInvestigationSize`, we may also have to change the implementation of their mocks here
3. with the new mocking method, we're involving the inner workings of `useInvestigationDetails` and `useInvestigationSize` too much when we really just want to test `visitDetailsPanel` as a unit. `useInvestigationDetails` and `useInvestigationSize` have their own unit tests where we actually do stub `axios` to test them in this fashion. So this isn't really necessary

packages/datagateway-common/src/detailsPanels/dls/visitDetailsPanel.component.test.tsx
sam-glendenning: I personally couldn't say for sure but if you think it is then great!

Change it if you want to, I wouldn't say it's urgent as there's still a lot of RTL migration to do but doing it with `configureStore` will help as we go further down the line

packages/datagateway-common/src/detailsPanels/dls/visitDetailsPanel.component.test.tsx
```diff
@@ -132,137 +148,144 @@ describe('Visit details panel component', () => {
...
+    renderComponent({ rowData });
     expect(
-      wrapper.find('[data-testid="visit-details-panel-no-name"]').exists()
-    ).toBeTruthy();
-  });
-
-  it('calls useInvestigationDetails and useInvestigationSize hooks on load', () => {
-    createWrapper();
-    expect(useInvestigationDetails).toHaveBeenCalledWith(rowData.id);
-    expect(useInvestigationSize).toHaveBeenCalledWith(rowData.id);
```
sam-glendenning, line 168: And just tagging onto my last comment, because of the changes we can no longer test these are being called (or we can but have chosen not to). So would be nice to have these back

packages/datagateway-common/src/detailsPanels/isis/datafileDetailsPanel.component.test.tsx
```diff
@@ -31,22 +50,65 @@ describe('Datafile details panel component', () => {
...
+
+  it('should show default tab on first render', async () => {
+    renderComponent({
+      rowData,
+    });
+
+    expect(
+      await screen.findByRole('tab', { name: 'datafiles.details.label' })
+    ).toHaveAttribute('aria-selected', 'true');
   });
```
sam-glendenning, line 75: Could be good to have this test in other files with a default tab. Currently I only see it here and in `isis/datasetDetailsPanel`

packages/datagateway-common/src/detailsPanels/isis/datafileDetailsPanel.component.test.tsx
```diff
@@ -31,22 +50,65 @@ describe('Datafile details panel component', () => {
...
+        name: 'datafiles.details.parameters.label',
+      })
+    );
+
+    expect(
+      await screen.findByRole('tabpanel', {
+        name: 'datafiles.details.parameters.label',
+      })
+    ).toBeVisible();
+  });
```
sam-glendenning, line 109: And this one too actually, I only see it in a few

packages/datagateway-common/src/detailsPanels/isis/datafileDetailsPanel.component.test.tsx
```diff
@@ -124,16 +177,28 @@ describe('Datafile details panel component', () => {
...
+    );
 
-    expect(detailsPanelResize).toHaveBeenCalledTimes(2);
+    await waitFor(() => {
+      expect(mockDetailsPanelResize).toHaveBeenCalledTimes(2);
+    });
   });
 
-  it('does not call detailsPanelResize if not provided', () => {
+  it('should not call detailsPanelResize if not provided', async () => {
```
sam-glendenning, line 201: I know `detailsPanelResize` is an optional parameter but do we ever run into a scenario where it's not provided? Because if not, we could make it a mandatory parameter and this test wouldn't even be needed (and maybe would require changing the `useLayoutEffect` hook we have as well).

packages/datagateway-common/src/detailsPanels/isis/instrumentDetailsPanel.component.test.tsx
```diff
@@ -65,37 +83,43 @@ describe('Instrument details panel component', () => {
...
+  it('should let user switch between tabs', async () => {
     rowData.instrumentScientists = [
       {
         id: 4,
         user: {
           id: 5,
-          name: 'Louise',
-          fullName: 'Louise Davies',
+          name: 'Rick',
+          fullName: 'Rick Ashley',
```
sam-glendenning, line 97: Good test data

packages/datagateway-common/src/detailsPanels/dls/visitDetailsPanel.component.test.tsx
kennethnym: React testing library highly encourages integration testing over unit testing. Since the hooks are used in `VisitDetailsPanel`, I think it is worthwhile to see if the hooks work well with the component itself. If we stub the hooks, it encourages testing the implementation details of the component itself. In the migration to RTL for `datagateway-download`, I used the same approach to test the components as well.

packages/datagateway-common/src/detailsPanels/isis/datafileDetailsPanel.component.test.tsx
kennethnym: Good shout! I think I forgot to put it in other tests.

packages/datagateway-common/src/detailsPanels/dls/visitDetailsPanel.component.test.tsx
sam-glendenning: Alright I'll go with it, I just think it makes it harder to figure out what's going wrong if something breaks. And what does this mean for the test files in `common/src/api`? Are some of them no longer necessary?

packages/datagateway-common/src/detailsPanels/isis/investigationDetailsPanel.component.tsx
```diff
@@ -266,7 +311,7 @@ const InvestigationDetailsPanel = (
                 );
               })
             ) : (
-              <Typography data-testid="investigation-details-panel-no-samples">
+              <Typography data-testid="n">
```
sam-glendenning, line 314: Is this a typo?

sam-glendenning:
I have noticed one thing - when you have a tab whose purpose is to viewDatasets or viewDatafiles, that's all good but when you go back to the previous page and expand the panel again, it looks a little funny because that tab is still selected:

image

I think in this scenario (if we can detect it) is to revert back to DEFAULT_TAB when this happens. What do you think?

kennethnym:
I think I know why this is happening - when clicking on "View datasets" it redirects to another page, but it is still considered "selected", so when you go back, the details pane shows it as selected (from the store) even though it is not supposed to be "selectable". I should be able to get around it by making a special case for it - not update the store when it is "selected".

packages/datagateway-common/src/detailsPanels/isis/investigationDetailsPanel.component.tsx
kennethnym: It probably is!

packages/datagateway-common/src/detailsPanels/isis/datafileDetailsPanel.component.test.tsx
kennethnym: I actually don't know. I'll have a look to see if I can make it non-optional, but there has to be a reason why it is optional.

packages/datagateway-common/src/detailsPanels/dls/visitDetailsPanel.component.test.tsx
kennethnym: The tests in `common/src/api` uses `renderHook` from `testing-library` to test the individual hooks, so I think they are fine. I do agree it is sometimes difficult to figure out what went wrong when things break.

packages/datagateway-common/src/detailsPanels/dls/visitDetailsPanel.component.test.tsx
kennethnym: I will keep it for now, I mentioned to Louise about the deprecation, and she said the deprecation is more like "redux toolkit is now preferred" rather than it being actually "deprecated".

kennethnym: https://github.com/reduxjs/redux/releases/tag/v4.2.0 It is deprecated in this release.

sam-glendenning: Ah okay, it was only because my IDE marked it as deprecated that I knew at all!

packages/datagateway-common/src/detailsPanels/isis/datafileDetailsPanel.component.test.tsx
sam-glendenning: Yeah because we have this test which is all well and good but it doesn't actually do anything 😅 like you instantiate a mock function, don't do anything with it and then assert that it's not called 😂 so yeah I think it would be good to see if it truly is non-optional and if so, this test can go

sam-glendenning:
Yeah that's also a valid solution!

packages/datagateway-common/src/detailsPanels/isis/datafileDetailsPanel.component.test.tsx
kennethnym: I think they are already in all the details panels that have tabs
- isis/dataset
- isis/datafile
- isis/instrument
- isis/investigation
- dls/dataset
- dls/visit

so this can be resolved

sam-glendenning:
For reference, `datagateway-table` is an extremely outdated label referring to the original name of dataview! Ask Louise, she'll remember the old days

sam-glendenning:
@kennethnym your reformatting commit is so big I can't actually see what you changed besides it. Could you point me to the change you made to fix the original problem?

sam-glendenning (approved):
Nice fix!

I'm not sure why prettier wanted to reformat so many files though. Not that there's anything wrong with it that I can see

kennethnym:
@sam-glendenning I merged `develop` into this branch to fix some merge conflicts, which introduced some style changes due to `prettier` upgrade. This happened to some of my other PRs as well. Thanks for the review though! I will go ahead and merge this PR.

PR #1388, opened 2022-08-09
Use different settings.json file for running e2e in CI vs locally

### LINKED ISSUE #1370: Detect whether e2e test is run locally or CI and select the correct settings file accordingly
**Description**:
At the moment if e2e tests are run locally they'd fail because the settings file point to URLs that only work in a CI environment. There should be a way to detect whether the tests are run in CI or locally and select the correct settings accordingly. That way we don't have to manually edit the settings file to make the tests run in environments other than CI.

**Acceptance criteria**:
- [ ] The environment is detected correctly and the correct settings file is used accordingly in e2e tests.

kennethnym:
## Description
This PR adds the ability for e2e test servers to select the corresponding `*-settings.json` file depending on whether the e2e tests are running in CI or not. If the tests are running in CI, then the `*-settings.json` in `server/` is used. Otherwise, the one in `public/` is used.

## Testing instructions

- [ ] Review code
- [ ] Check Actions build

## Agile board tracking
connect to #1370 

Closes #1370

### DIFF
--- a/packages/datagateway-dataview/cypress/support/commands.js
+++ b/packages/datagateway-dataview/cypress/support/commands.js
@@ -55,7 +55,8 @@ export const readSciGatewayToken = () => {
 };
 
 Cypress.Commands.add('login', (credentials) => {
-  return cy.readFile('server/e2e-settings.json').then((settings) => {
+  return cy.request('datagateway-dataview-settings.json').then((response) => {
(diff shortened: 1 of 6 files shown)


sam-glendenning (approved):
Nice fix, tested locally and works fine :)

PR #1393, opened 2022-08-18
Add a datafile previewer for previewing datafiles in-situ, implemented for .txt & .log

### LINKED ISSUE #1213: Visualise .txt files in situ
**Description**:
The catalogue has main text files and it would be good to be able to visualise them in the browser without having to download them. We will look at other data file formats next. 

e.g. see txt files in https://data.isis.stfc.ac.uk/browse/instrument/6/facilityCycle/55/investigation/24068500/dataset/24068066/datafile?sort=%7B%22modTime%22%3A%22desc%22%7D

**Acceptance criteria**:
- [ ] Add visualisation of text files in the browser

kennethnym:
## Description
This PR adds the functionality to preview datafiles directly in the website.

The preview UI can be triggered by:
- visiting links to datafiles
- clicking on preview buttons in datafile tables

The previewer only supports some datafile types (`txt` and `log` at the moment`). The previewer first checks the file extension of the datafile. If the given datafile is unsupported, it will show a message telling the user that the datafile cannot be previewed because it is unsupported, and also the file extension of the datafile, so that if they want to request support for a certain datafile type, they can just pass along the extension. Otherwise, it delegates the parsing and rendering of the preview to the appropriate component.

### Functionality

- Download the datafile being previewed
- Copy link to the current datafile. Browsing the copied link will open the previewer again.
- Zoom controls for controlling the size of the preview
- A details side panel showing the details of the datafile

### Possible new features to consider (txt preview only)

- Jump to line number through links
- Text searching

### Extending the previewer

The rendering delegation is defined by a lookup table in `./previewComponents/previewComponents`. To add support for a new file extension, simply add an entry to the table, mapping the new file extension to the corresponding React component that can render a preview of the datafile.

For example, the table maps `txt` to `TxtPreview` which is a React component that previews the given datafile as a `txt` file.

### Demonstration

This video demonstrates the functionality of the previewer:

https://user-images.githubusercontent.com/19359984/185383632-ebbc6496-7dea-4f2e-b06e-84ce508ec52d.mp4

### Misc

This PR also includes reformatting of files in other packages due to prettier upgrade, otherwise e2e tests will fail on the linting step.

## Testing instructions
- [ ] Review code
- [ ] Check Actions build

## Agile board tracking
connect to #1213 

Closes #1213


agbeltran:
this is looking great @kennethnym ! As this PR deals with issue #1213, we can mark it as ready for review. We will deal with the preview of other data types in other issues. Thanks!

kennethnym:
Just a quick note that a lot of the diffs are caused by prettier format changes. Ideally it should not be in this PR, but tests won't pass without the format changes.

packages/datagateway-common/src/api/datafiles.tsx
```diff
@@ -251,13 +338,26 @@ export const downloadDatafile = (
   };
 
   const link = document.createElement('a');
-  link.href = `${idsUrl}/getData?${Object.entries(params)
-    .map(([key, value]) => `${key}=${value}`)
-    .join('&')}`;
+  const objectUrl = content && window.URL.createObjectURL(content);
```
louise-davies, line 341: I'm still wary of using `Blob`s for downloads - have we tested this new method when trying to download large files? e.g. can the 8 GB file here be downloaded via this method? https://datagateway-dev.isis.stfc.ac.uk/browse/instrument/57054468/facilityCycle/84793030/investigation/89234877/dataset/89234894/datafile?sort=%7B%22fileSize%22%3A%22desc%22%7D - via the current method I'm able to download this file (albeit it takes a while!)

louise-davies: It's a bit different for a user to ask to preview a large file - where they'd probably expect it would be slow and impracticable to do so, than a user downloading it - where it happens in the browser's background so they're less affected by the wait

louise-davies: Ah wait nevermind - I see that this is only used when the `content` param is supplied - which is only done in the previewer

kennethnym: Yep, normal downloads won't use Blobs, it is only used for when the previewer has already downloaded the datafile as a Blob, and `downloadDatafile` acts more like save this Blob as a file to the computer.

packages/datagateway-dataview/src/settings.ts
```diff
@@ -6,6 +6,7 @@ export interface DataviewSettings {
   apiUrl: string;
   downloadApiUrl: string;
   idsUrl: string;
+  icatUrl: string;
```
louise-davies, line 9: Can we remove this? We have a issue to track improving the developer login process here: https://github.com/ral-facilities/datagateway/issues/1400. At the very least `icatUrl` should be optional (as it's not required in prod)

packages/datagateway-dataview/src/state/app.types.tsx
```diff
@@ -10,6 +14,7 @@ export interface DGDataViewState {
   settingsLoaded: boolean;
   selectAllSetting: boolean;
   pluginHost: string;
+  isisDatafilePreviewer: DatafilePreviewerState;
```
louise-davies, line 17: small point, but we could probably rename this to be more generically `datafilePreviewer` as it's not technically ISIS specific - it's just singe level IDS specific

packages/datagateway-dataview/src/views/datafilePreview/datafilePreviewer.component.tsx
```diff
@@ -0,0 +1,280 @@
...
+  const isDetailsPaneShown = useSelector<StateType, boolean>(
+    (state) => state.dgdataview.isisDatafilePreviewer.isDetailsPaneShown
+  );
+  const [isDetailsPaneGridVisible, setIsDetailsPaneGridVisible] =
+    React.useState(isDetailsPaneShown);
+  const [isDetailsPaneIn, setIsDetailsPaneIn] =
+    React.useState(isDetailsPaneShown);
+
+  React.useEffect(() => {
+    // This effect controls the appearance of details panel
```
louise-davies, line 148: Is there not an out of the box way to do this via MUI like using `Drawer`? I have examples in OperationsGateway in how to make it not fill the whole screen/fit the container it's in. However, if this works it works, just is a little confusing having 3 variables for the details pane visibility 😅

kennethnym: `Drawer` isn't quite what I'm looking for - it is full height, and it sits on top of the content, not side by side.

packages/datagateway-dataview/src/views/datafilePreview/previewDatafileButton.component.tsx
```diff
@@ -0,0 +1,50 @@
...
+function PreviewDatafileButton({
+  datafile,
+}: PreviewDatafileButtonProps): JSX.Element {
+  const [t] = useTranslation();
+
+  const isSupported = isDatafilePreviewable(datafile);
+
+  return (
+    <Link
+      to={`datafile/${datafile.id}`}
```
louise-davies, line 25: Clicking on the preview button for some reason causes the whole page to reload for me, despite the fact that this is using a `react-router` link. If you instead use the `component` prop of `IconButton` to render that as a `Link`, I think it should work fine

louise-davies (commented):
Also, technically you can preview datafiles from incorrect paths, e.g. using ISIS dev data, if I take the datafile id from `/browse/instrument/24/facilityCycle/87730662/investigation/90027513/dataset/90027523/datafile/90028171` and use it in this path `/browse/instrument/23/facilityCycle/84793030/investigation/87813986/dataset/89475874/datafile/90028171`, the same file loads each time even though the latter's hierarchy is wrong. This can be fixed by using `WithIdCheck`

packages/datagateway-dataview/src/views/datafilePreview/datafilePreviewer.component.tsx
louise-davies: It can be used like this though - see:

OG-plot-config

This is on OG develop. However, the current way works as well - it was just a thought

kennethnym: Interesting. This may sound nitpicky, but when you expand/close the side panel, the content on the right hand side "jumps" instead of expand/shrink itself smoothly, which is why I needed to code in special logic to handle it.

louise-davies: Hmm true, maybe we should steal your code in OG ;)

kennethnym: Feel free to ~~steal~~ adapt it as you wish!

kennethnym:
> Also, technically you can preview datafiles from incorrect paths, e.g. using ISIS dev data, if I take the datafile id from `/browse/instrument/24/facilityCycle/87730662/investigation/90027513/dataset/90027523/datafile/90028171` and use it in this path `/browse/instrument/23/facilityCycle/84793030/investigation/87813986/dataset/89475874/datafile/90028171`, the same file loads each time even though the latter's hierarchy is wrong. This can be fixed by using `WithIdCheck`

Resolved!

kennethnym:
This might need slight changing to allow disable file content fetching done automatically by the datafile previewer. HDF5 is a good example - they can be large in file size, so it is not a good idea to load all the file content to memory. In this case we want to delegate the content fetching to the preview component.

louise-davies (changes requested):
Code looks fine - but I found a bug. If you preview a datafile, then go back to the list of datafiles via the `Other Datafiles` button, and then preview the same file again the control bar at the top is missing. The same thing happens if you navigate away and then back in any other way - e.g. via breadcrumbs back up to dataset and then go back to same file, via browser back button etc. I presume it's because the controls wait until the data has been fetched to display perhaps, but in these cases the react-query already has the result cached so they don't load properly?

Otherwise, my other comments (preview button causing a whole page reload, and the URL not being checked) have both been addressed.

packages/datagateway-dataview/src/views/datafilePreview/previewComponents/textPreview.component.tsx
```diff
@@ -0,0 +1,112 @@
...
+/**
+ * A {@link PreviewComponent} that previews text-based datafiles.
+ * @see PreviewComponent
+ */
+function TextPreview({
+  datafile,
+  datafileContent,
+}: PreviewComponentProps): JSX.Element {
+  // const [isReadingContent, setIsReadingContent] = React.useState(true);
+  // const [textContent, setTextContent] = React.useState('');
```
louise-davies, line 61: Remove commented out state

kennethnym: Oof, forgot about it :p

louise-davies (approved):
Looks good - the previous bug I reported was fixed and the new querying method works well.

PR #1398, opened 2022-08-24
#1397 - make fileCountMax / totalSizeMax optional

### LINKED ISSUE #1397: Cart total file count/size error warnings pop up even if the limits are -1 (disabled)
**Description**:
ISIS doesn't use `fileCountMax`/`totalSizeMax` - but the settings were defined as mandatory in the schema so the way to disable them in practice is to set them to `-1`. However, the new errors just check for `fileCount > fileCountMax` etc., which always triggers if `fileCountMax` is `-1`.

image

We should probably change things so that `fileCountMax` & `totalSizeMax` are optional, and thus it prompts developers to null check things like these checks to ensure they don't trigger if the administrator hasn't set them.

**Acceptance criteria**:
- [ ] Errors in the screenshot don't occur in ISIS where no limits are enforced

louise-davies:
This means we can more easily remember to check if they're set to a valid number via TS and to disable any features related to them if they're undefined - primilary what's fixed here is the "too many/much" alerts.

Because this is needed to create a patch release, it's needs to merge onto main (and it branched from main). I will then merge main into develop + resolve any merge conflicts.

## Description
Enter a description of the changes here

## Testing instructions

- [ ] Review code
- [ ] Check Actions build
- [ ] {more steps here}

## Agile board tracking
Closes #1397

### DIFF
--- a/packages/datagateway-download/src/ConfigProvider.tsx
+++ b/packages/datagateway-download/src/ConfigProvider.tsx
@@ -16,8 +16,8 @@ export interface DownloadSettings {
   downloadApiUrl: string;
   idsUrl: string;
 
-  fileCountMax: number;
-  totalSizeMax: number;
+  fileCountMax?: number;
+  totalSizeMax?: number;
 
   accessMethods: DownloadSettingsAccessMethod;
   routes: PluginRoute[];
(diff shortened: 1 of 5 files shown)


louise-davies:
Note - because this is branched from main, it is before the yarn 2 upgrade and so it will likely bork your existing repo's dependencies. I'd suggest checking this out in a seperate clone of the repo/git worktree so as to not cause as muh trouble when switching back to develop later.

sam-glendenning:
May also want to amend the code at https://github.com/ral-facilities/datagateway/blob/923473a55a04530008210bb47a4f26b7aafd31b3/packages/datagateway-common/src/table/cellRenderers/cellContentRenderers.tsx#L7 since parts of it are no longer relevant

sam-glendenning (approved):

PR #1401, opened 2022-08-24
Lucene search UI changes

### LINKED ISSUE #1363: Implement new UI changes to support new Lucene
**Description**:
Follows on from #1361 - certain UI changes we like what Patrick has done and so we want to port them over exactly - the sort and my data checkboxes in the search controls, removing of the filters from the tables (and removing some padding). These need to either be sorted by #1361 or re-implemented.

**Acceptance criteria**:
- [ ] Patrick's UI changes that are fine as is (i.e. search box controls, removal of table header padding) are implemented

kennethnym:
## Description
This PR includes support for new Lucene search & new UI to support the Lucene search change.

## Testing instructions

- [ ] Review code
- [ ] Check Actions build

## Agile board tracking
connect to #1363 

Closes #1363


kennethnym:
The checks are failing for the E2E tests because the test servers don't support the new Lucene search endpoint yet.

packages/datagateway-download/src/index.tsx
```diff
@@ -93,6 +93,13 @@ export const fetchSettings = (): Promise<DownloadSettings | void> => {
         );
       }
 
+      // Ensure all fileCountMax and totalSizeMax are present.
+      if (!('fileCountMax' in settings && 'totalSizeMax' in settings)) {
+        throw new Error(
+          'fileCountMax or totalSizeMax is undefined in settings'
+        );
+      }
+
```
sam-glendenning, line 102: Following changes made in https://github.com/ral-facilities/datagateway/pull/1398, this needs to be removed

sam-glendenning:
Not entirely sure where this is going wrong right now but the datafile search table is not stretching to fill the width of its container, unlike the other search tables

mismatch table size

kennethnym:
That is weird, I never had this problem before! I'll have a look.

sam-glendenning:
Note for the above: this was in Firefox. Let me know if things are different for you in other browsers, etc

packages/datagateway-common/src/card/parameterFacetList.tsx
```diff
@@ -0,0 +1,80 @@
...
+interface ParameterFacetListProps {
+  entityName: DatasearchType;
+  parameterTypeName: string;
+  facets: ParameterValueFacet[];
+  changeFilter: (key: string, value: SearchFilter, remove?: boolean) => void;
+  setFilterUpdate: React.Dispatch<React.SetStateAction<boolean>>;
+  valueType: 'DATE_AND_TIME' | 'STRING';
+}
+
+export const ParameterFacetList = (
```
sam-glendenning, line 17: I'd support this file being renamed to `parameterFacetList.component.tsx`

packages/datagateway-common/src/detailsPanels/dls/visitDetailsPanel.component.tsx
```diff
@@ -282,6 +295,7 @@ const VisitDetailsPanel = (
                 return (
                   <Grid key={sample.id} item xs>
                     <Typography>
+                      {sample.type?.name ? `${sample.type.name}: ` : ''}
                       <b>{sample.name}</b>
                     </Typography>
                   </Grid>
```
sam-glendenning, line 301: It might be good to refactor this to only return a Grid and Typography element if we have sample.type.name, instead of returning an empty Typography element in a Grid. So basically refactor to something like:
```
investigationData.samples.filter((sample) => !!sample.type?.name).map((sample) => {
    return (
        <Grid key={sample.id} item xs>
            <Typography>
                {sample.type?.name ? `${sample.type.name}: ` : ''}
                <b>{sample.name}</b>
            </Typography>
        </Grid>
    );
})
```

packages/datagateway-search/public/datagateway-search-settings.json.DEV
```diff
@@ -0,0 +1,18 @@
+{
```
sam-glendenning, line 1: This file will need to be removed

packages/datagateway-search/public/datagateway-search-settings.json.TESTING
```diff
@@ -0,0 +1,18 @@
+{
```
sam-glendenning, line 1: As will this one

packages/datagateway-search/cypress/integration/search/datafileSearch.spec.ts
```diff
@@ -32,7 +33,7 @@ describe('Datafile search tab', () => {
 
     cy.get('[aria-label="Submit search"]')
       .click()
-      .wait(['@datafiles', '@datafiles', '@datafilesCount'], {
+      .wait(['@searchDocuments', '@datafilesCount'], {
```
sam-glendenning, line 36: Am I right in saying that the queries to /search/documents now cover all attempts to query individually for investigations, datasets and datafiles? If so, we should remove the intercepts for investigations, datasets and datafiles, since these are no longer required (see lines 16, 18 and 20)

sam-glendenning: Actually, I forgot that individual count queries aren't made either. So all the intercepts between lines 13 and 20 can be removed (also make sure to remove any `cy.wait` calls we make to their respective aliases

packages/datagateway-search/src/searchBoxContainer.component.tsx
```diff
@@ -1,15 +1,15 @@
 import React from 'react';
```
sam-glendenning, line 1: I note that `searchBoxContainer` has been changed but `searchBoxContainerSide` hasn't. I think `containerSide` will also need to be provided the same changes

packages/datagateway-search/src/searchBoxContainer.component.test.tsx
```diff
@@ -14,41 +19,66 @@ jest.mock('react-redux', () => ({
...
+
+  let state: DeepPartial<StateType>;
 
   beforeEach(() => {
-    useSelector.mockImplementation(() => {
-      return initialState;
-    });
+    state = {
+      dgsearch: {
+        sideLayout: true,
```
sam-glendenning, line 41: This should be false, since we are not testing `containerSide` here

packages/datagateway-search/src/searchBoxContainer.component.test.tsx
```diff
@@ -14,41 +19,66 @@ jest.mock('react-redux', () => ({
...
+    };
   });
 
   it('renders searchBoxContainer correctly', () => {
-    const wrapper = createWrapper();
-    expect(wrapper).toMatchSnapshot();
+    const { asFragment } = renderComponent({ initialState: state });
+    expect(asFragment()).toMatchSnapshot();
   });
 });
```
sam-glendenning, line 51: We should add more tests to check for things like the MyDataCheckbox appearing if the user is not logged in  anonymously, if it calls the correct method, etc.

packages/datagateway-search/src/searchTabs.component.tsx
```diff
@@ -13,29 +11,31 @@ import {
...
 import DatasetCardView from './card/datasetSearchCardView.component';
 import DatasetSearchTable from './table/datasetSearchTable.component';
+import SearchTabLabel from './searchTabLabel.component';
+
+export interface SearchTableProps {
+  containerHeight: string;
+  hierarchy: string;
+  onTabChange: (currentTab: string) => void;
+  currentTab: string;
+}
```
sam-glendenning, line 38: I don't think these are being used anywhere and are a duplicate of `SearchTabsProps` so they can be removed

packages/datagateway-search/src/card/datasetSearchCardView.component.tsx
```diff
@@ -28,12 +21,19 @@ import {
...
+  FiltersType,
+  formatBytes,
 } from 'datagateway-common';
 import { useTranslation } from 'react-i18next';
 import { useHistory, useLocation } from 'react-router-dom';
+import { styled, Paper, Typography } from '@mui/material';
 import { useSelector } from 'react-redux';
 import { StateType } from '../state/app.types';
-import { styled } from '@mui/material';
+import { CVCustomFilters } from 'datagateway-common/lib/card/cardView.component';
```
sam-glendenning, line 36: I think it's better to export the `CVCustomFilters` type directly from datagateway-common as a whole instead of importing it like this? So maybe best to remove this line and add this  in `datagateway-common/src/index.tsx`:
```
export type { CVCustomFilters } from './card/cardView.component';
```

packages/datagateway-search/src/card/datasetSearchCardView.component.tsx
```diff
@@ -238,31 +314,33 @@ const DatasetCardView = (props: DatasetCardViewProps): React.ReactElement => {
   // hierarchy === 'isis' ? data : undefined is a 'hack' to only perform
   // the correct calculation queries for each facility
```
sam-glendenning, line 315: This can be removed

packages/datagateway-search/src/card/datasetSearchCardView.component.tsx
```diff
@@ -348,57 +432,69 @@ const DatasetCardView = (props: DatasetCardViewProps): React.ReactElement => {
...
+          data={paginatedSource ?? []}
+          totalDataCount={aggregatedIds?.length + (hasNextPage ? 1 : 0) ?? 0}
+          onPageChange={pushPage}
+          onFilter={pushFilter}
+          onSort={handleSort}
+          onResultsChange={pushResults}
+          loadedData={!isLoading}
+          loadedCount={!isLoading}
+          filters={parsedFilters}
+          sort={{}}
```
sam-glendenning, line 486: I'm just uninformed but why is data not sortable anymore?

packages/datagateway-search/src/search/advancedHelpDialogue.component.test.tsx
```diff
@@ -39,12 +52,8 @@ describe('Advanced help dialogue component tests', () => {
...
-      return dgSearchInitialState;
-    });
-
-    const wrapper = shallow(<AdvancedHelpDialogue />);
-    expect(wrapper).toMatchSnapshot();
+    const { asFragment } = renderComponent({ initialState: state });
+    expect(asFragment()).toMatchSnapshot();
   });
 
   it('can open and close help dialogue', () => {
```
sam-glendenning, line 59: It's a good idea to rewrite this test using RTL (although it may be being done through https://github.com/ral-facilities/datagateway/issues/1440 so if so, ignore this comment)

packages/datagateway-search/src/search/myDataCheckBox.component.tsx
```diff
@@ -0,0 +1,41 @@
...
+import { useTranslation } from 'react-i18next';
+import { Checkbox, FormControlLabel } from '@mui/material';
+import { ArrowTooltip } from 'datagateway-common';
+
+interface MyDataCheckBoxProps {
+  checked: boolean;
+  onChange: (checked: boolean) => void;
+}
+
+const MyDataCheckBox = ({
```
sam-glendenning, line 11: This will need a test file (can be done through https://github.com/ral-facilities/datagateway/issues/1440 if needed)

packages/datagateway-search/src/search/sortSelect.component.tsx
```diff
@@ -0,0 +1,85 @@
...
+const MenuProps = {
+  PaperProps: {
+    style: {
+      maxHeight: ITEM_HEIGHT * 4.5 + ITEM_PADDING_TOP,
+      width: 250,
+    },
+  },
+};
+
+const SortSelectComponent = (): React.ReactElement => {
```
sam-glendenning, line 19: This will also need a test file

packages/datagateway-search/src/state/actions/index.tsx
```diff
@@ -43,6 +45,15 @@ export const loadMaxNumResults = (
...
 
+export const loadMinNumResults = (
+  minNumResults: number
+): ActionType<ConfigureMinNumResultsPayload> => ({
+  type: ConfigureMinNumResultsType,
+  payload: {
+    minNumResults: minNumResults,
+  },
+});
+
```
sam-glendenning, line 56: We'll need a test for this in `actions.test.tsx`

packages/datagateway-search/src/state/reducers/dgsearch.reducer.test.tsx
```diff
@@ -81,7 +81,7 @@ describe('dgsearch reducer', () => {
   });
 
   it('should set maxNumResults property when configuring action is sent', () => {
```
sam-glendenning, line 83: Should add an equivalent test for `minNumResults`

packages/datagateway-search/src/table/cellRenderers.tsx
```diff
@@ -0,0 +1,67 @@
...
+  const result = useDatasetSizes(dataset);
+  return <>{formatCountOrSize(result[0], true)}</>;
+}
+
+export {
+  InvestigationDatasetCountCell,
+  InvestigationSizeCell,
+  DatasetDatafileCountCell,
+  DatasetSizeCell,
+};
```
sam-glendenning, line 67: Can do with a test file for this file

packages/datagateway-search/src/search/advancedHelpDialogue.component.test.tsx
kennethnym: Yep, in fact I have already migrated this test to RTL!

VKTB:
While testing the backend search changes using the UI from this PR, I noticed from the `/icat/search/documents` requests that `restrict` is set to `True` when _neutron AND scattering_ is clicked from the suggested options while the _My data_ checkbox is unchecked. The checkbox in the UI doesn't get updated when `restrict` gets set to `True`.

It is also worth mentioning that deselecting some of the _Types_ and then clicking on  _neutron AND scattering_ results in all types being re-selected, see GIF below.

Search Changes1

sam-glendenning:
Thanks @VKTB, yes so neutron AND scattering is a hard-coded link, putting all the parameters into the address bar. Clearly there's a bug here where the restrict flag is also being mistakenly set so we'll have to fix that. It's the same with instrument calibration. You're right to raise it though because, as these are hard-coded, it's important to make sure they are working correctly since they are examples for the end user

packages/datagateway-search/src/table/datasetSearchTable.component.test.tsx
```diff
@@ -1,662 +1,577 @@
...
-            <DatasetSearchTable hierarchy={hierarchy ?? ''} />
-          </QueryClientProvider>
-        </Router>
-      </Provider>
-    );
+  /**
+   * Mock implementation of axios.get
+   */
+  const mockAxiosGet = (url: string): Promise<Partial<AxiosResponse>> => {
+    console.log('url', url);
```
sam-glendenning, line 96: Can be removed

packages/datagateway-search/src/table/datasetSearchTable.component.test.tsx
```diff
@@ -1,662 +1,577 @@
...
-    );
-
-    filterInput.instance().value = '';
-    filterInput.simulate('change');
-
-    expect(history.length).toBe(3);
-    expect(history.location.search).toBe('?');
-
-    cleanupDatePickerWorkaround();
+    // the checkbox should be checked
```
sam-glendenning, line 236: *not be checked

packages/datagateway-search/src/table/datasetSearchTable.component.test.tsx
```diff
@@ -1,662 +1,577 @@
...
+    const selectAllCheckbox = await screen.findByRole('checkbox', {
+      name: 'select all rows',
     });
-    const wrapper = createWrapper();
 
-    wrapper.find('[aria-label="select row 0"]').last().simulate('click');
-
-    expect(addToCart).toHaveBeenCalledWith([1]);
+    expect(selectAllCheckbox).not.toBeChecked();
+    expect(selectAllCheckbox).toHaveAttribute('data-indeterminate', 'false');
```
sam-glendenning, line 267: Is this right? If only some checkboxes are checked, surely the selectAllCheckbox is in an indeterministic state, right?

packages/datagateway-search/src/table/datafileSearchTable.component.test.tsx
```diff
@@ -4,18 +4,17 @@ import { initialState as dgSearchInitialState } from '../state/reducers/dgsearch
 import configureStore from 'redux-mock-store';
```
sam-glendenning, line 4: This file will need a refactor with RTL

packages/datagateway-search/src/table/datasetSearchTable.component.test.tsx
```diff
@@ -1,662 +1,577 @@
...
-  it('does not render ISIS link when facilityCycleId cannot be found', () => {
-    const wrapper = createWrapper('isis');
-
-    expect(wrapper.find('[aria-colindex=3]').find('a')).toHaveLength(0);
-    expect(wrapper.find('[aria-colindex=3]').text()).toEqual(
-      'Dataset test name'
-    );
-  });
+  it('does not render ISIS link when facilityCycleId cannot be found', async () => {
+    renderComponent('isis');
```
sam-glendenning, line 440: Don't we have to remove the `facilityCycleId` from the response first?

packages/datagateway-search/src/table/investigationSearchTable.component.test.tsx
```diff
@@ -1,622 +1,527 @@
...
+// ====================== END FIXTURE ======================
 
 jest.mock('datagateway-common', () => {
   const originalModule = jest.requireActual('datagateway-common');
 
   return {
     __esModule: true,
     ...originalModule,
+    // fetchDownloadCart: jest.fn(),
+    // addOrRemoveFromCart: jest.fn(),
```
sam-glendenning, line 87: Can be removed

packages/datagateway-search/src/table/investigationSearchTable.component.test.tsx
```diff
@@ -1,622 +1,527 @@
...
+    // create a new array to trigger useMemo update
+    cartItems = [];
 
-    expect(history.length).toBe(3);
-    expect(history.location.search).toBe('?');
+    // clicks on the row checkbox
+    await user.click(screen.getByRole('checkbox', { name: 'select row 0' }));
 
-    cleanupDatePickerWorkaround();
+    // the checkbox should be checked
```
sam-glendenning, line 273: *not be checked

packages/datagateway-search/src/table/investigationSearchTable.component.test.tsx
```diff
@@ -1,622 +1,527 @@
...
+    const selectAllCheckbox = await screen.findByRole('checkbox', {
+      name: 'select all rows',
     });
-    const wrapper = createWrapper();
-
-    wrapper.find('[aria-label="select row 0"]').last().simulate('click');
 
-    expect(addToCart).toHaveBeenCalledWith([1]);
+    expect(selectAllCheckbox).not.toBeChecked();
+    expect(selectAllCheckbox).toHaveAttribute('data-indeterminate', 'false');
```
sam-glendenning, line 304: As above with the earlier data-indeterminate comment

packages/datagateway-search/src/searchPageCardView.component.tsx
```diff
@@ -0,0 +1,435 @@
...
+        ],
+      },
+      filters
+    );
+
+  const countSearchResults = (
+    searchResponses: InfiniteData<SearchResponse> | undefined,
+    hasNextPage: boolean | undefined
+  ): string => {
+    console.log('searchPageCardView countsearchresults', searchResponses);
```
sam-glendenning, line 211: Can be removed

packages/datagateway-search/package.json
```diff
@@ -104,7 +104,9 @@
     ]
   },
   "devDependencies": {
+    "@testing-library/jest-dom": "^5.16.5",
     "@testing-library/react": "12.1.3",
+    "@testing-library/user-event": "^14.4.1",
```
sam-glendenning, line 109: These dependencies need to be pinned for Renovate to properly manage them

sam-glendenning:
Think that's all the comments I have to make for now so that's my first review done! General advice is to address what I've said above, fix merge conflicts, double check we have no remaining enzyme tests and all tests are passing

kennethnym:
Thank you so much for the review! I'll have a look at this today.

packages/datagateway-common/src/detailsPanels/dls/visitDetailsPanel.component.tsx
kennethnym: I think it's fine as it is because `sample.name` is required and is always shown, so `Typography` should never be empty. If we filtered samples with no type names, then even their names will be invisible.

packages/datagateway-search/src/card/datasetSearchCardView.component.tsx
kennethnym: I'm not sure either, I think Patrick added the code in 860e70a9eff620efecf887309912363dc6ecc9d5 which seems to be for demoing, so I don't know if it's meant to be "un-sortable" or if it's just there so that the demo works.

sam-glendenning: Apologies, Kenneth, I later clarified this with someone. Data is sortable, it just happens in a different place. It happens in that dropdown list next to the search button. So that handles all that now. This can be resolved

packages/datagateway-search/src/state/reducers/dgsearch.reducer.test.tsx
kennethnym: I also updated dgsearch reducer because `loadMinNumResults` action wasn't handled.

packages/datagateway-search/src/table/datafileSearchTable.component.tsx
```diff
@@ -45,155 +64,354 @@ const DatafileSearchTable = (
...
+      return Object.entries(filters).map((dimension) => {
+        const dimensionKey = dimension[0].toLocaleLowerCase();
+        const dimensionValue = dimension[1];
+        return {
+          label: t(dimensionKey),
+          dataKey: dimensionKey,
+          dataKeySearch: dimensionKey
+            .replace('investigation.', '')
+            .replace('investigationparameter.', 'investigationparameter ')
+            .replace('sample.', 'sample '),
```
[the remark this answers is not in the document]
kennethnym: I am not too sure but when I tested this again the search seems to work fine - the investigation search table and the datafile search table is showing correct data.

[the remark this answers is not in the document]
sam-glendenning: In that case I am happy with it as is

packages/datagateway-search/src/table/datafileSearchTable.component.test.tsx
kennethnym: I migrated this test to RTL in #1447. Should I port the test to this PR?

packages/datagateway-search/src/table/datasetSearchTable.component.test.tsx
sam-glendenning: Ahh okay, thanks for the clarification

packages/datagateway-search/src/table/datafileSearchTable.component.test.tsx
sam-glendenning: Nah, eventually this PR and yours will need to sync and I'd rather do that all in one go. Good it's been resolved, thanks!

packages/datagateway-search/src/table/datasetSearchTable.component.test.tsx
kennethnym: However, the "wait for query client to finish fetching" technique can be deployed in other tests,  I vaguely remember some tests having the same issue but I can't remember which off the top of my head. I can have a look at it and implement the changes in #1447

packages/datagateway-search/src/table/investigationSearchTable.component.test.tsx
kennethnym: Same comment as above.

packages/datagateway-search/src/table/__snapshots__/investigationSearchTable.component.test.tsx.snap
```diff
@@ -1,158 +1,699 @@
...
+              >
+                <div
+                  class="MuiTableCell-root MuiTableCell-body MuiTableCell-sizeSmall css-188rx6i-MuiTableCell-root"
+                >
+                  <p
+                    aria-label=""
+                    class="MuiTypography-root MuiTypography-body2 MuiTypography-noWrap css-cvfmw9-MuiTypography-root"
+                    data-mui-internal-clone-element="true"
+                  >
+                    Calculating...
```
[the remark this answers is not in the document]
kennethnym: The snapshots are going to be removed in #1447 because I removed snapshot testing in most component tests

packages/datagateway-search/src/table/datasetSearchTable.component.test.tsx
sam-glendenning: Fantastic

kennethnym:
This is a video demonstrating the reworked parameter filter UI that I have been working on:

https://user-images.githubusercontent.com/19359984/214312409-ed689392-c98b-49ce-a1d2-95636d7ea527.mp4

Any feedback is appreciated!

patrick-austin:
> Any feedback is appreciated!

It's more of a general comment rather than being related to the parameter value filters, and ultimately it's a UI/UX thing but seeing the video where you do a lot of selecting, unselecting etc. made me wonder if the `APPLY` button should always be enabled as it is now. Testing on my local branch you can click it even having no facet filters selected, and it will send a request to the backend (needlessly). I might find it more intuitive if it was disabled (if that's the right term - greyed out basically) unless you have made a change to the filters that is yet to be applied?

kennethnym:
Good point - that is definitely an oversight on my part.

patrick-austin:
I noticed a couple of things the other day, just want to make sure they haven't fallen through the cracks:

- [ ] Error handling: Long running searches (how long is configurable on the backend) and searches with bad syntax (e.g. `/`) should display useful errors to the user (i.e how to avoid the issue in future). I wrote some suggestions on https://github.com/ral-facilities/datagateway/issues/1363#issuecomment-1208230279
- [ ] Help text: Similarly, I drafted a more detailed (possibly too detailed) up to date version of the help text on https://github.com/ral-facilities/datagateway/tree/feature/updated-search-help-text-1363. Might want to edit this down, but based on the user sessions we did behaviour of things like file paths can be unexpected so should be documented somewhere.

kennethnym:
Got it, will have a look.

packages/datagateway-common/src/detailsPanels/investigationDetailsPanel.component.tsx
```diff
@@ -49,15 +54,23 @@ const InvestigationDetailsPanel = (
           {t('investigations.details.start_date')}
         </Typography>
         <Typography>
-          <b>{investigationData.startDate}</b>
+          <b>
+            {investigationData.startDate
+              ? new Date(investigationData.startDate).toLocaleDateString()
```
louise-davies, line 59: Is it better to use `date-fns` `format` function to format datestrings here rather than relying on `toLocaleDateString` to ensure a consistent format?

packages/datagateway-common/src/detailsPanels/dls/visitDetailsPanel.component.tsx
```diff
@@ -295,6 +309,70 @@ const VisitDetailsPanel = (
...
+                    case 'DATE_AND_TIME':
+                      return (
+                        <Grid key={parameter.id} item xs>
+                          <Typography variant="overline">
+                            {parameter.type.name}
+                          </Typography>
+                          <Typography>
+                            <b>
+                              {parameter.dateTimeValue &&
+                                parameter.dateTimeValue.split(' ')[0]}
```
louise-davies, line 356: If this is trying to get just the date - again it's probably preferred to use `date-fns` to be more resilient to future timestamp format changes.

packages/datagateway-search/cypress/e2e/search/datafileSearch.cy.ts
```diff
@@ -2,131 +2,411 @@ describe('Datafile search tab', () => {
...
-    cy.intercept('**/investigations/count?where=%7B%22id*').as(
-      'investigationsCount'
-    );
-    cy.intercept('**/investigations?*').as('investigations');
-    cy.intercept('**/datasets/count?where=%7B%22id*').as('datasetsCount');
-    cy.intercept('**/datasets?*').as('datasets');
-    cy.intercept('**/datafiles/count?where=%7B%22id*').as('datafilesCount');
-    cy.intercept('**/datafiles?*').as('datafiles');
-    cy.intercept(`**/topcat/user/cart/${facilityName}/cartItems`).as('topcat');
+    // cy.intercept(`**/topcat/user/cart*`, { statusCode: 200 });
```
louise-davies, line 13: Commented out code

packages/datagateway-search/cypress/support/commands.js
```diff
@@ -42,6 +43,7 @@ export const readSciGatewayToken = () => {
   const token = window.localStorage.getItem('scigateway:token');
   let sessionId = null;
   let username = null;
+  console.log('sessionId', sessionId);
```
louise-davies, line 46: Remove console.log

packages/datagateway-search/src/card/datasetSearchCardView.component.tsx
```diff
@@ -59,79 +66,165 @@ const DatasetCardView = (props: DatasetCardViewProps): React.ReactElement => {
...
+    },
+    [t]
   );
 
-  const textFilter = useTextFilter(filters);
-  const dateFilter = useDateFilter(filters);
+  const parsedFilters = React.useMemo(() => {
+    const parsedFilters = {} as FiltersType;
+    Object.entries(filters).forEach((v) => {
+      parsedFilters[v[0].substring(8)] = v[1]; // "dataset." is 8 characters
```
louise-davies, line 167: Rather than use `.substring(8)` here, is it perhaps clearer to use `.replace("dataset.", "")`? I think those will end up equivalent results (presuming the parsed filter always starts with `dataset.`)

packages/datagateway-search/src/facet/useFacetFilters.ts
```diff
@@ -0,0 +1,137 @@
...
+  };
+
+  const applyFacetFilters = React.useCallback((): void => {
+    const searchParams = new URLSearchParams(location.search);
+    // const filters = Object.entries(selectedFacetFilters).reduce<FiltersType>(
+    //   (obj, [dimension, value]) => {
+    //     obj[dimension.toLocaleLowerCase()] = value;
+    //     return obj;
+    //   },
+    //   {}
```
louise-davies, line 111: Commented out code

packages/datagateway-search/src/searchPageContainer.component.test.tsx
```diff
@@ -121,6 +132,7 @@ describe('SearchPageContainer - Tests', () => {
     };
 
     (axios.get as jest.Mock).mockImplementation((url) => {
+      console.log('url', url);
```
louise-davies, line 135: Remove console.log

packages/datagateway-search/src/table/investigationSearchTable.component.tsx
```diff
@@ -48,115 +52,141 @@ const InvestigationSearchTable = (
...
-      filterType: 'include',
-      filterValue: JSON.stringify({
-        investigationInstruments: 'instrument',
-      }),
-    },
-  ]);
-  const { data: allIds, isLoading: allIdsLoading } = useIds(
-    'investigation',
-    [
+  console.log('new filters', filters);
```
louise-davies, line 73: Remove console.log

packages/datagateway-search/src/table/cellRenderers.tsx
```diff
@@ -0,0 +1,67 @@
...
+  useInvestigationDatasetCount,
+  useInvestigationSize,
+} from 'datagateway-common';
+import React from 'react';
+
+/**
+ * A cell in {@link InvestigationSearchTable} that displays the count of a dataset.
+ * @param investigation
+ */
+function InvestigationDatasetCountCell({
```
louise-davies, line 16: A slight concern I have is over the performance of each size/count query being separate as that has caused performance issues in the past. However, I expct Jounaid's size/count refactor to go into the same release as this search stuff so no major concern - although that's something to keep in mind for later as there's still merge jobs to do.

packages/datagateway-search/src/facet/components/facetPanel/toggleableFilterItem.component.tsx
```diff
@@ -0,0 +1,63 @@
...
+          sx={{ padding: 0, paddingRight: 1 }}
+          checked={selected}
+          inputProps={{
+            'aria-label': `${
+              selected ? 'Remove' : 'Add'
+            } ${classificationLabel} filter`,
+          }}
+          onChange={(event) =>
+            onSelect(classificationLabel, event.target.checked)
+          }
```
louise-davies, line 48: Remove this `onChange` handler in favour of the one on the list item - as otherwise clicking on the checkbox specifically will cause 2 chips to get added

louise-davies (changes requested):
Code is looking good other than my minor comments above. In playing about with the UI - I found a strange interaction with the parameter filters with card view. Neither the type or parameter name facets show up in the card view itself, but the parameter filters do which looks a little strange due to duplication. Also, deleting the card view facet causes the UI to think there are no results at all! See gif for demo

DG-card-chips

Also, as a minor style point it feels as though there's too much empty space on the sides of the cards - especially since there's now extra blank space on the left due to the filters panel.

packages/datagateway-search/src/table/cellRenderers.tsx
kennethnym: I think I added this was because the batch query approach would also query for sizes that are not currently visible in the table, so when they would error out it would cause a flood of errors in the console, completely bogging down the website. Separating the query into its own component has the benefit of preventing unnecessary queries.

With the new investigation size/count refactor though, like you said, I don

packages/datagateway-search/src/card/datasetSearchCardView.component.tsx
kennethnym: I don't think this code is needed anymore, as it is only used for showing the selected filters in `cardView`. Since it is already handled externally, `parsedFilters` can be safely removed.

packages/datagateway-search/src/card/investigationSearchCardView.component.tsx
```diff
@@ -337,54 +467,120 @@ const InvestigationCardView = (
...
+                totalDataCount={
+                  aggregatedIds?.length + (hasNextPage ? 1 : 0) ?? 0
+                }
+                onPageChange={pushPage}
+                onFilter={pushFilter}
+                onSort={handleSort}
+                onResultsChange={pushResults}
+                loadedData={!isLoading}
+                loadedCount={!isLoading}
+                filters={filters}
```
louise-davies, line 568: Need to remove passing of filters to `investigationSearchCardView` as well

kennethnym: Oops, forgot about it.

louise-davies:
Just noticed something whilst improving test coverage - we don't actually use the changes made to card view in this PR right? If I'm correct I'm going to revert them as we don't need it - unless someone things we should keep the new card view - in which I then invite them to write tests for it.

patrick-austin:
You can display search results in table or card view right (unless that's changed since I started on this)? I remember needing to make some changes to how filters were handled, as card view already had the start of a second "faceted" filtering system in place, but it needed some tweaking so it would work with the Lucene filters.

(Appreciate this was directed at Kenneth not me, and whatever initial stuff I did will have been long since overtaken by other changes - so my comments are probably not particularly pertinent.)

louise-davies:
@patrick-austin yeah, but since we needed it to work for both card and table view, there's now a `FacetPanel` component (for adding new filters) and a `SelectedFilterChips` which are used instead so the code is the same for both views. So, I was wondering if it's fine to remove all the card view specific stuff. I believe it is, because if I delete it then filtering/faceting still works, and we can tell we're not using the code since it's uncovered in the tests (hence why I want to remove it!)

patrick-austin:
Yeah that makes sense, I think I started with modifying card view because there was more code in place already to work with, but makes sense that it would've been refactored out at some point for tables too.

patrick-austin:
Have added requests for `instrument.name` and changed `type.name` to `sample.type.name` for the Investigation -> Sample faceting. Also had to add a `?` to the `investigationFacilityCycles` array in the Datafile table to allow for empty arrays (this was present for `investigationInstrument` and the other tables already.

While it should now (with latest backend changes) be possible to facet on `instrument.name` or `instrument.fullName` by configuration settings in `icat.lucene`, the values displayed for Datasets and Datafiles are counter intuitive. Rather than showing the number of the loaded rows which have each instrument set, they instead show the number of distinct investigations. For example:
Screenshot 2023-09-28 102930
We have 1000 Datafiles but since each block 300 belongs to the same Investigation, when joining to the InvestigationInstrument index by investigation.id all 300 (or however many) Datafiles still only count as 1. So the numbers aren't what you might expect, but are an accurate count of how many unique InvestigationInstruments there are linked to those Datafiles. It may be difficult to convey this to users though - I suppose you could either remove the numbers entirely or try and explain the fact each Investigation is counted once in the UI (via tooltips or help text)? Or you could just disable this facet for Datasets are Datafiles (i.e. remove the lines I've added where we request it).

This would not be easy to solve on the backend. In principle you could:
- Build a "DatasetInstrument" and "DatafileInstrument" index, which will allow us to count unique datafile.ids and so get an accurate count. The downside is that this will require a significant amount of special case logic, and create an index as big as Datafile (so potentially several billion for Diamond) which could reduce performance and increase storage needed for the index.
- Represent the Investigation -> Instrument relationship as 1:1 instead of 1:*. This would allow us to "flatten" information onto the Datasets and Datafiles directly, which is more convenient when searching but again will require special logic to handle (e.g. if a second instrument is added, do we overwrite the first?)

But neither is ideal, and I'd be reluctant to add even more complicated special case handling.

louise-davies:
@patrick-austin I might investigate the possibility of disabling the numbers for these specific facets via some config, otherwise I'll just make it so you can only facet on `Instrument` at the `Investigation` level as the numbers are just too confusion IMO

louise-davies (approved):

PR #1404, opened 2022-08-26
Use react query provided timestamps for download table last checked time

### LINKED ISSUE #1395: Last refreshed timestamp is not updated when user refreshes the download table
**Description**:
Sabah reported that the timestamp does not update when he refreshes the admin download table. This is likely due to the react-query refactor, and we should check this bug on both download status tables.

**Acceptance criteria**:
- [ ] Refreshing the table should update the "Last updated" timestamp correctly

kennethnym:
## Description
This PR addresses the issue where the last checked time is not updated when the download tables are refreshed. Instead of tracking the update time manually, `react-query` `useQuery` supplies a `dataUpdatedAt` value that gives the timestamp of when the query is last updated in ms.

## Testing instructions
- [ ] Review code
- [ ] Check Actions build

## Agile board tracking
Closes #1395

### DIFF
--- a/packages/datagateway-download/src/downloadStatus/adminDownloadStatusTable.component.test.tsx
+++ b/packages/datagateway-download/src/downloadStatus/adminDownloadStatusTable.component.test.tsx
@@ -70,11 +70,16 @@ describe('Admin Download Status Table', () => {
     jest.clearAllMocks();
   });
 
-  it('should render correctly', () => {
+  it('should render correctly', async () => {
     const mockedDate = new Date(Date.UTC(2020, 1, 1, 0, 0, 0)).toUTCString();
     global.Date.prototype.toLocaleString = jest.fn(() => mockedDate);
 
     const { asFragment } = renderComponent();
+
+    // wait for data to finish loading
+    expect(
+      await screen.findByText(mockedDate.toLocaleString())
+    ).toBeInTheDocument();
 
     expect(asFragment()).toMatchSnapshot();
   });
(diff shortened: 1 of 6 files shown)


packages/datagateway-download/src/downloadStatus/adminDownloadStatusTable.component.test.tsx
```diff
@@ -70,12 +70,17 @@ describe('Admin Download Status Table', () => {
...
     const mockedDate = new Date(Date.UTC(2020, 1, 1, 0, 0, 0)).toUTCString();
     global.Date.prototype.toLocaleString = jest.fn(() => mockedDate);
 
     const { asFragment } = renderComponent();
 
+    // wait for data to finish loading
+    expect(
+      await screen.findByText(mockedDate.toLocaleString())
+    ).toBeInTheDocument();
+
```
sam-glendenning, line 83: Hi Kenneth, as I'm sure you've seen, this results in a big addition to the snapshot for this test? Am I right in saying this is because the old snapshot was totally wrong and just described an empty table with no rows??? 🤦

kennethnym: Yes, that is correct. The old snapshot doesn't wait for the table to finish loading all the data so it is essentially just a snapshot of its loading state, which is not very useful.

sam-glendenning: Oh dear 🤦 thank you for spotting this. It is so important to religiously check snapshots before changing them because they're our best guess of exactly what the DOM looks like. Plus, it's not like computers can check them!

kennethnym: I think in general [snapshot testing is not recommended](https://kentcdodds.com/blog/effective-snapshot-testing) for any snapshot bigger than a couple lines, but I kept it because it is already there.

sam-glendenning (approved):
Nice change! And I agree about the excessive snapshot but at least you made sure the test for it was now properly running

kennethnym:
Thank you for the review! I'll go ahead and merge this PR.

PR #1412, opened 2022-09-07
Fix 'Date To' Filter in user download table  

### LINKED ISSUE #1399: Requested Date lone to filter not working in user download table
**Description**:
If a user only fills out the `Requested Date To` filter, the filter doesn't work. It does work if you fill out the `From` filter though. Looking at the code, in `sortedAndFilteredData`, we need some more sophisticated logic to allow for all of neither, startDate, endDate and both to be set by the user and apply the correct filtering.

**Acceptance criteria**:
- [ ] A lone To filter in the requested date column in the user download table works

jounaidr:
### Description
Old logic would not get to filter range checking if startDate was null (lines 152/156). Changes remove this check and add a conditional where startTimestamp is set, so that if startDate is null, assign the timestamp to 0.

Unit test updated to test each filter (from/to) independently, as well as general comments added to test file.

### Testing instructions
Can be tested by updating 'Date To' filter text box whilst keeping the 'Date From' box empty:

chrome_V4AFWiDSer

### Agile board tracking
connect to #1399 
closes #1399

### DIFF
--- a/packages/datagateway-download/src/downloadStatus/downloadStatusTable.component.test.tsx
+++ b/packages/datagateway-download/src/downloadStatus/downloadStatusTable.component.test.tsx
@@ -363,16 +363,20 @@ describe('Download Status Table', () => {
       name: 'downloadStatus.createdAt filter to',
     });
 
+    // Type into date from filter textbox
     await user.type(dateFromFilterInput, '2020-01-01 00:00');
 
+    // Should show all files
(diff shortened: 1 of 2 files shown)


louise-davies (commented):
Tested the functionality and it works well with all combos of from & to date filters inputted either via the date picker or via the text box. The code looks good so it's ready to go from my perspective 😄  well done on your first issue 🎉

louise-davies (approved):

Issue #1411, opened 2022-09-05
Switch ISIS facility cycle calculations to use InvestigationFacilityCycle rather than date calculations


louise-davies:
**Description**:
See: https://github.com/ral-facilities/isis-icat-ingestion/issues/76. But essentially, since ICAT 5 has added a relationship between `Investigation` and `FacilityCycle`, and ISIS intends to start backfilling this imminently and wants to use this relationship instead of the date calculation method.

**Acceptance criteria**:
- [ ] We use the `InvestigationFacilityCycle` relationship instead of `Investigation.startDate` between `FacilityCycle` start & end dates in the ISIS hierarchy

PR #1413, opened 2022-09-07
Add links to the corresponding entities for download selections

### LINKED ISSUE #874: Add link to resources in the 'download selection' window
**Description**:
When looking at the download option of a selection, it would be really useful to have a link back to the resource (investigation, dataset, datafile) listed for download

dg-downloadcart

**Acceptance criteria**:
- [ ] Add link back to the resource from the selection download window

kennethnym:
## Description
This PR adds links to the corresponding entity in download selection tables for each download selection. For example, a download selection for an investigation in the download selection table will now link to the corresponding investigation. Datafile download selections will have links to the parent datasets.

The URL building logic is handled by 3 functions, `buildInvestigationUrl`, `buildDatasetUrl`, and `buildDatafileUrl`. The correct builder function is then passed to the link component (`DownloadCartItemLink`), depending on the `entityType` of the download selection. The link component consumes the builder function via `useQuery`, so the URLs are cached by `react-query` for each download selection.

Theoretically, the 3 search tables in `datagateway-search` can use the 3 functions instead of handling it in the table component itself, making the component shorter and more readable. If required, I can move the functions to `dagateway-common` so that it can be used by other datagateway packages, including `datagateway-search`. 

## Testing instructions
- [ ] Review code
- [ ] Check Actions build

## Agile board tracking

Closes #874


kennethnym:
Since we have #1411 coming up, maybe it's worth waiting for it to be done before merging this PR?

kaperoo:
Updated urlBuilders to use `investigationFacilityCycles` instead of calculating FC using dates. Test data has also been updated to include this relationship.

packages/datagateway-download/src/downloadCart/downloadCartTable.component.tsx
```diff
@@ -44,26 +56,36 @@ interface DownloadCartTableProps {
...
-
   const [showConfirmation, setShowConfirmation] = React.useState(false);
 
   const { data: isTwoLevel } = useIsTwoLevel();
   const { mutate: removeDownloadCartItem } = useRemoveEntityFromCart();
   const { mutate: removeAllDownloadCartItems, isLoading: removingAll } =
     useRemoveAllFromCart();
-  const { data, isFetching: dataLoading } = useCart();
+  const { data: cartItems, isFetching: isFetchingCart } = useCart();
+  const { data: facilityCycles, isFetching: isFetchingFacilityCycles } =
```
louise-davies, line 75: I'm pretty sure this is no longer needed right? Since we get facility cycles from the investigation link now

packages/datagateway-download/src/downloadCart/urlBuilders/buildInvestigationUrl.ts
```diff
@@ -0,0 +1,108 @@
...
+      filterType: 'where',
+      filterValue: JSON.stringify({ id: { eq: investigationId } }),
+    },
+  ];
+
+  if (facilityName === 'ISIS') {
+    filters.push({
+      filterType: 'include',
+      filterValue: JSON.stringify(['investigationInstruments.instrument']),
+    });
```
louise-davies, line 28: Don't we also need to `include` `investigationFacilityCycle.facilityCycle`?

packages/datagateway-download/src/downloadCart/urlBuilders/buildDatasetUrl.ts
```diff
@@ -0,0 +1,91 @@
...
+  datasetId,
+}: {
+  apiUrl: string;
+  facilityName: string;
+  datasetId: Dataset['id'];
+}): Promise<Dataset | null> {
+  let includeField: string;
+  switch (facilityName) {
+    case 'ISIS':
+      includeField = 'investigation.investigationInstruments.instrument';
```
louise-davies, line 20: Don't we also need to `include` `investigationFacilityCycle.facilityCycle`?

packages/datagateway-download/src/downloadCart/urlBuilders/buildDatafileUrl.ts
```diff
@@ -0,0 +1,72 @@
...
+}: {
+  apiUrl: string;
+  facilityName: string;
+  datafileId: Datafile['id'];
+}): Promise<Datafile | null> {
+  let includeField: string;
+  switch (facilityName) {
+    case 'ISIS':
+      includeField =
+        'dataset.investigation.investigationInstruments.instrument';
```
louise-davies, line 21: Don't we also need to `include` `investigationFacilityCycle.facilityCycle`?

louise-davies (changes requested):
Need to ensure facility cycle relation is included for ISIS and need to remove the all facility cycles stuff

packages/datagateway-download/src/downloadCart/urlBuilders/buildDatafileUrl.test.ts
```diff
@@ -0,0 +1,67 @@
...
+
+    expect(url).toBe('/browse/investigation/58/dataset/856/datafile');
+  });
+
+  it('should return an ISIS URL to the parent dataset of the datafile', async () => {
+    const url = await buildDatafileUrl({
+      apiUrl: mockedSettings.apiUrl,
+      facilityName: 'ISIS',
+      datafileId: 70,
+      facilityCycles: mockFacilityCycles,
```
louise-davies, line 32: We don't need to pass this in in the tests anymore - leaving it here is potentially confusing

packages/datagateway-download/src/downloadCart/urlBuilders/buildInvestigationUrl.test.ts
```diff
@@ -0,0 +1,171 @@
...
+        data: [investigation],
+      });
+    });
+
+    it('should return a generic URL to it', async () => {
+      const url = await buildInvestigationUrl({
+        investigation,
+        apiUrl: mockedSettings.apiUrl,
+        facilityName: 'SVELTE',
+        facilityCycles: mockFacilityCycles,
```
louise-davies, line 26: We don't need to pass this in in the tests anymore - leaving it here is potentially confusing

louise-davies (changes requested):
It's not technically an issue with this PR, but the `dataset` datafileCount query (downloadApi.ts L322) fails because it's params are not JSON stringified

Otherwise, this seems to be working well :)

louise-davies (approved):
LGTML

PR #1414, opened 2022-09-09
fix timestamps for Requested At columns in download tables

### LINKED ISSUE #1410: Time not displayed in Requested At column in download tables
**Description**:
This can be seen on DLS preprod - but an update to `date-fns-tz` (specifically v1.2.0) broke how we handle the download `createdAt` timestamps. I believe the problem is the dates are returned in a format like `2022-08-31T13:01:37.944Z[UTC]` - which gives the name of the timezone in brackets next to the timezone offset (which is `Z`). If we remove the `[UTC]` bit, via `"2022-08-31T13:01:37.944Z[UTC]".replace(/\[.*\]/, "");` then it seems to work fine again.

**Acceptance criteria**:
- [ ] Times are correctly displayed in downloadStatusTable and adminDownloadStatusTable

jounaidr:
## Description
Bug #1410 fixed by using the regex replace suggested, `.replace(/\[.*]/, '')`, for both admin and normal download tables. More verbose validation could be done around `props.cellData` where we get the timestamp, however I presume validation is handled by date-fns-tz in the `toDate` method when parsing the date/time string (despite it not working right now :sweat_smile:), so adding more would most likely be redundant.

Additional testing is not required, as multiple date/time strings with the `[UTC]` tag are passed through within the snapshot tests. It seems [adminDownloadStatusTable.component.test.tsx.snap](https://github.com/ral-facilities/datagateway/pull/1414/files#diff-ee01cdf6466b0603ef0131094312ad44d569e5965510cbf1791c49f930d12035) was testing incorrect behaviour, and excepted the times to display as 00:00 given a string with the `[UTC]` tag, so I updated it.

This PR breaks [downloadStatus.spec.ts](https://github.com/ral-facilities/datagateway/pull/1414/files#diff-6ad1b3632d22ae29ffe2fdda72b61f79718f313a0b579dc6e9e04cf72f4eb147) e2e test, as the _should be able to filter download items by date between_ test case makes a check using a date object with the time set to 00:00, which now fails as the correct time is displayed. Current fix uses a `contain` to check the date is as expected, without comparing the time. This is because the test files seem to update somewhat inconsistently, and often it was testing against a timestamp significantly earlier than when the test is being run, making any checks against the time ineffective. I think this test case could be reworked (once filtering by time is working), however I believe this is out of scope for this issue.  

## Testing instructions
Can be seen on /download and /admin/download endpoints

download:
image
## Testing instructions
Can be seen on /download and /admin/download endpoints

download:
image

admin download:
image

## Agile board tracking
closes #1410

### DIFF
--- a/packages/datagateway-download/cypress/integration/downloadStatus.spec.ts
+++ b/packages/datagateway-download/cypress/integration/downloadStatus.spec.ts
@@ -165,7 +165,7 @@ describe('Download Status', () => {
       cy.get('[aria-rowcount="0"]').should('exist');
 
       const currDate = new Date();
-      const assertDate = new Date(currDate.toString());
+      const assertDate = currDate;
 
       cy.get('input[id="Requested Date filter from"]').clear();
       cy.get('input[id="Requested Date filter to"]').clear();
@@ -180,10 +180,11 @@ describe('Download Status', () => {
 
       cy.get('[aria-rowcount="4"]').should('exist');
 
-      cy.get('[aria-rowindex="1"] [aria-colindex="4"]').should(
-        'have.text',
-        format(assertDate, 'yyyy-MM-dd HH:mm:ss')
-      );
+      cy.get('[aria-rowindex="1"] [aria-colindex="4"]')
+        .should(($el) => {
+          const actual = new Date(String($el.text()))
+          expect(actual.getTime()).to.be.closeTo(assertDate.getTime(), 10000);
+        })
     });
 
     it('multiple columns', () => {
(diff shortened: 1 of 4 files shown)


packages/datagateway-download/cypress/integration/downloadStatus.spec.ts
```diff
@@ -178,13 +178,13 @@ describe('Download Status', () => {
...
-        'have.text',
-        format(currDate, 'yyyy-MM-dd HH:mm:ss')
+        'contain',
+        format(currDate, 'yyyy-MM-dd')
       );
     });
 
     it('multiple columns', () => {
-      cy.get('[aria-label="Filter by Access Method').first().type('globus');
+      cy.get('[aria-label="Filter by Access Method"]').first().type('globus');
```
jounaidr, line 187: Seems `"]` was missing here

louise-davies (approved):
Looks good - and it's fine to just test the date filtering in the e2e tests for now as to do otherwise requires mocking fake download data

Issue #1408, opened 2022-09-02
Issue when filtering by title


agbeltran:
**Description**:
When navigating experiments and then filtering by title, I seemed to have found a bug.

To reproduce:
- in DG for ISIS, browse by Experiments
- choose the GEM instrument
- in the title filter, enter 'tests'

URL: https://data.isis.stfc.ac.uk/browseStudyHierarchy/instrument/28/study?filters=%7B%22studyInvestigations.investigation.title%22%3A%7B%22value%22%3A%22tests%22%2C%22type%22%3A%22include%22%7D%7D&sort=%7B%22studyInvestigations.investigation.startDate%22%3A%22desc%22%7D

the first 3 results seem OK, but not sure why the fourth result is listed - see screenshot

<img width="1180" alt="Screenshot 2022-09-02 at 16 30 09" src="https://user-images.githubusercontent.com/1417033/188183580-93340b1e-e373-4004-bd4d-f8299444b78d.png">

Now, changing the filter from include to exclude, note that the first result contains 'tests' (and it wasn't given in the previous list)

<img width="1345" alt="Screenshot 2022-09-02 at 16 31 37" src="https://user-images.githubusercontent.com/1417033/188183790-6b5136ab-5c91-4c23-bf1b-bc4e4fe0a4cc.png">

URL: https://data.isis.stfc.ac.uk/browseStudyHierarchy/instrument/28/study?filters=%7B%22studyInvestigations.investigation.title%22%3A%7B%22value%22%3A%22tests%22%2C%22type%22%3A%22exclude%22%7D%7D&sort=%7B%22studyInvestigations.investigation.startDate%22%3A%22desc%22%7D

Could there be hidding characters that are affecting these filtering results?

**Acceptance criteria**:
- [ ] Investigate and fix filtering issue

louise-davies:
@agbeltran we think this may be related to #1390 - which is a problem with the current implementation of the study view

PR #1416, opened 2022-09-15
Fix incorrect study dates shown in ISIS study views with date filters

### LINKED ISSUE #1390: Incorrect study dates and duplicate rows appearing
## Description:
I noticed an issue on ISIS prod whereby filtering on date parameters for studies can result in rows being displayed that show results outside the user's filtered date range.

See https://data.isis.stfc.ac.uk/browseStudyHierarchy/instrument/23/study?filters=%7B%22studyInvestigations.investigation.startDate%22%3A%7B%22endDate%22%3A%222018-08-08%22%2C%22startDate%22%3A%222017-08-08%22%7D%7D&sort=%7B%22studyInvestigations.investigation.startDate%22%3A%22desc%22%7D for an example. You will notice we are filtering on start date between 2017 and 2018 and a result has appeared for 2015.

After discussing this with @MRichards99, who was also in the room, we came to the conclusion that this study is being displayed because it contains _at least one_ investigation matching the user's query. However, instead of displaying the start date for that matching investigation, it displays the start date for the _first_ investigation in that study's list of investigations (which was 2015 in this case).

I've also ascertained that this can lead to what looks like duplicate rows appearing. Here's what happens:
1. for each investigation that matches the user's search, a row is created in the table, containing its study info
2. because of the above, _the start date of the first investigation of that study_ is inserted into the study start date field (not the study start date)
3. therefore, if a study has, for example, two investigations that match the user's search, two rows for that study will be displayed that look exactly the same. The start date will contain the start date of the _first_ investigation, ignoring the start date of the other completely

The studies view (two rows, one for each matching investigation of that study)
image

The investigation view of that study (they both have different start dates but only the first is used in the study table)
image

Because these two issues are intrinsically linked, I recommend both be tackled at the same time.

## Tackling the problem:
I'd recommend starting here - https://github.com/ral-facilities/datagateway/blob/4a727092ba3badab5e0c460783f70acf903a2523/packages/datagateway-common/src/table/cellRenderers/cellContentRenderers.tsx#L38
This could be where the incorrect study info is obtained. We also need to determine what other views could be affected by this. Check if this affects facility cycle investigations and DataGateway Search.

@MRichards99 and I discussed the overall solution a bit but could be worth discussing further. Start with making sure the study start date field in the studies table is correct (so insert the start date for the matching investigation in question). Next to talk about is how to tackle the duplicate rows. Do we want this to happen, especially since both link to the same page anyway? Perhaps it is best to determine if we have two duplicate rows and only display one (perhaps the first matching one).

**Acceptance criteria**:
- [ ] A checklist

kennethnym:
## Description
This PR addresses an issue where the dates of studies are sometimes incorrect when date filters are applied.

Each `Study` object has a `studyInvestigations` field which is a list of `StudyInvestigation`s that belong to the `Study`. Currently, for `Study`s with multiple `StudyInvestigation`s, only the info of the first one is shown (e.g. start date and end date). With the filters are applied, the first `StudyInvestigation` in the list may not necessarily match the active filters, causing a mismatch between the info that is shown and the active filters.

I created a new function that, given a list of `StudyInvestigation`s and the active filters, returns the `StudyInvestigations` that match the filters. Then, to show each matched `StudyInvestigation`s properly, for each matched `StudyInvestigation`s, a new `Study` object is created and added to the list of `Study`s. Finally, the aggregated list is passed to either the table view or the card view. This means a `Study` can appear as "duplicate" rows, but each row will show info of a different `StudyInvestigation`.

## Testing instructions

- [x] Review code
- [x] Check Actions build

## Agile board tracking
Closes #1390

### DIFF
--- a/packages/datagateway-common/src/table/cellRenderers/cellContentRenderers.test.tsx
+++ b/packages/datagateway-common/src/table/cellRenderers/cellContentRenderers.test.tsx
@@ -280,7 +280,7 @@ describe('Cell content renderers', () => {
 
     it('should return a list of study investigations that match the given end date filter with no start date', () => {
       const endDateFilter: DateFilter = {
-        endDate: '2010-03-01',
+        endDate: '2022-03-01',
       };
 
       const filters: FiltersType = {
--- a/packages/datagateway-common/src/table/cellRenderers/cellContentRenderers.test.tsx
+++ b/packages/datagateway-common/src/table/cellRenderers/cellContentRenderers.test.tsx
@@ -9,7 +9,14 @@ import {
   tableLink,
   formatCountOrSize,
   getStudyInfoInvestigation,
+  filterStudyInfoInvestigations,
 } from './cellContentRenderers';
+import type {
+  DateFilter,
+  FiltersType,
+  Study,
+  TextFilter,
+} from '../../app.types';
 
 describe('Cell content renderers', () => {
   describe('formatBytes', () => {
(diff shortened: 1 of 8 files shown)


louise-davies:
@kennethnym did we decide what to do re: duplicating rows causing a mismatch between the count and the table? I think I preferred the solution where we don't add any extra rows and just instead try and ensure that the investigation displayed matches the filters? This could still end up with "duplicate" rows unless we somehow keep track of which investigations for a specific study ID have already been displayed and use that to decide between multiple matching investigations...

However, I think the duplicate rows issue as described in the original issue isn't as bad as mismatching the count or the filtering not working as expected... Thoughts?

kennethnym:
Now that I think about it, the mismatch between displayed rows and row count should never happen - if there are more than 1 investigation matched for a study, then for the row count it should return the correct number accordingly, e.g. if a study has 2 matched investigations, the row count should be 2, as shown in the screenshots attached in the original issue.

louise-davies (approved):
Tested this. It fixes #1408, and partially fixes #1390 (we still have duplicate rows, but hopefully better "matches" appear when filtering). The results match the results count, and I think this is the best solution until we get `DataPublications`.

PR #1419, opened 2022-09-21
Add label text for include/exclude filters to make them more explicit

### LINKED ISSUE #1409: Filters include/exclude
**Description**:
Related to #1408, when including or excluding keywors for filtering, the UI doesn't show what of the options is chosen - it looks exactly the same if we are including or excluding the keyword.

Can we change the UI to reflect what option is selected?

@mattjSTFC may have suggestions

**Acceptance criteria**:
- [ ] change the filtering UI to make clear if exclude or include option is selected

jounaidr:
## Description
Currently, the include/exclude text filters are not explicit as to which filter is being applied once text has been entered into the field:
image

This PR replaces the current placeholder text with an InputLabel, which will place the label text above the text box once text has been entered. See:

image

#### Unit/e2e tests

[table.component.test.tsx](https://github.com/ral-facilities/datagateway/pull/1419/files#diff-3276aaeec1948cc1c371c3982cdf9690079f30cc6b9cf7a18d7d2641e45b4360) has been updated, as the test that would check if the first column is rendered correctly (which passes a label 'Test 1') now also has the 'Include' text from the InputLabel. To keep this specific assert testing what it was intended to, I have just updated the expected text to what is now being returned.

To test this feature specifically, a new assert has been added to the _it renders data columns correctly_ test case, which looks for the label tag and checks its text render is as expected (should say 'Include' by default).

As no extra functionality is being introduced (only visual changes), I believe extra e2e tests are not required.

## Testing instructions
View any include/exclude text filter.

## Agile board tracking
closes #1409


louise-davies:
Worth exploring the use of a label instead of a placeholder - in MUI when the textbox is active these labels go to the top of the textbox

louise-davies:
@mattjSTFC we do have the ability to set a secondary colour for things like this if we'd prefer a colour other than blue ;)

mattjSTFC:
Interesting ... what would that be? In order to get this out the door maybe leave the icon colour change for now.

jounaidr:
include-exclude

mattjSTFC:
From a UX point of view the label certainly helps improve whether your are including or excluding text. There are some presentation issues with spacing / alignment but I imagine these can be addressed another day? 

How does this look in situ? It would be useful to see how it looks alongside the From and To date fields. Especially as there is now more vertical space.

jounaidr:
Thanks @mattjSTFC, here is a screenshot of the entire page:

image

Do you think this option in general is better than the icon adornments? If the icons are preferred I can create an issue for the colour change of actionable icons, and prepare the PR around that. Otherwise ill prepare the PR around the label solution. Thanks :)

mattjSTFC:
I would say that the icon adornments primarily helped reinforce whether it was included or excluded as a visual clue. The plus icon alongside include and the minus icon alongside excluded worked as one. However the critical part was the textual element: include or exclude. If it's going to create more work, then I don't think it is going to make a dramatic improvement in the experience having the icons included as well so would leave it as you have now.

sam-glendenning (approved):
Hope you don't mind me jumping in, this needed a review! I'm happy with this, all looks good. One thing to bear in mind is that this obviously makes tables with more columns look a bit worse, such as the admin download table, and makes it harder to click on the text input.

image

However, there really isn't much we can do about that and it looks better with the drawer on the left closed so the table has more breathing room. Nice improvement!

mattjSTFC:
Hi @sam-glendenning Not seen this page before but also agree it doesn't look the best. Maybe some spacing right would help keep the icon away from the text input content.

At some point the filter work will remove this issue anyway so maybe not such a priority to keep tweaking it.

sam-glendenning:
Thanks for the input, Matt :) as long as we have some sort of plan to address this in the future, then I'm happy

PR #1431, opened 2022-10-04
Authenticate locally with datagateway-api

### LINKED ISSUE #1400: Authenticate Dataview and Download with DataGateway API
**Description**:
When running DataGateway locally, Dataview and Download have problems when submitting login requests because they do not make use of an `icatUrl` to submit a login request. The current code assumes Topcat and ICAT are on the same host but this is not the case in production. 

The current method is done using the `downloadApiUrl`, which is usually the topcat /sessions endpoint, but this results in a 404. Therefore, functionality that makes use of test data belonging to the root user, such as in Download, doesn't work.

image

It's probably best to migrate this old method to authenticating with DataGateway API instead. For consistency, we should do this with DataGateway Search as well.

**Acceptance criteria**:
- [ ] A checklist

jounaidr:
## Description
Changes to make datagateway-search and datagateway-dataview authenticate locally using datagateway-api.

Code has been duplicated across the packages. An attempt has been made to refactor the code into datagateway-common, however, since each `settings` object has a different and specific implementation per package, only about half of the code block (from where the axios request starts) can be easily refactored. Therefore I don't think its worth attempting further, especially since this just developer code.

As this is just dev code, additional testing is not required.

## Testing instructions
Check that a `sessionID` is returned when refreshing the page, and that no 404 errors (as was previously happening) are thrown.

## Agile board tracking
closes #1400


packages/datagateway-dataview/src/index.tsx
```diff
@@ -171,62 +171,36 @@ if (
...
-              });
-          })
-          .catch((error) =>
-            log.error(`Can't log in to ICAT: ${error.message}`)
+  settings.then((settingsResult) => {
+    if (settingsResult) {
+      const splitUrl = settingsResult.apiUrl.split('/');
+      const dataGatewayUrl = `${splitUrl
+        .slice(0, splitUrl.length - 1)
+        .join('/')}/datagateway-api`;
```
louise-davies, line 179: Can't we just use `apiUrl` from settings? We don't need to split and then add on `datagateway-api` again - we only had to do that previously because often TopCAT and ICAT were on the same server in dev, so you could derive the ICAT url from the TopCAT url (`downloadApiUrl`)

packages/datagateway-dataview/src/index.tsx
```diff
@@ -171,62 +171,36 @@ if (
   render();
   log.setDefaultLevel(log.levels.DEBUG);
 
-  if (process.env.NODE_ENV === `development`) {
```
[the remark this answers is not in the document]
jounaidr: Ahh okay, I misunderstood what that second if statement was trying to do and thought it was redundant, I have reverted the change, thanks.

packages/datagateway-dataview/src/index.tsx
jounaidr: Yes, this is much better thanks. Changes made.

louise-davies (approved):

PR #1434, opened 2022-10-06
Migrate lerna to yarn workspaces

### LINKED ISSUE #1114: Remove lerna dependency and investigate yarn 2
**Description**:
Lerna is pretty much unmaintained, and we only use it to run commands in all packages at once. Yarn 2 also has this feature, so we should investigate upgrading (using node_modules at first - not sure about PnP). If that doesn't work, we should still remove lerna and live with just running the commands in all packages manually

**Acceptance criteria**:
- [ ] Lerna removed
- [ ] Yarn 2 investigated

jounaidr:
## Description
Previously, we used lerna for monorepo support in the top level `package.json`, however it seems lerna is becoming unmaintained. This PR migrates all the commands previously implemented using lerna, to use yarn workspaces instead. 

The yarn plugin `workspace-tools` has been installed and used for the [`foreach`](https://yarnpkg.com/cli/workspaces/foreach) option, which runs the command for each package specified by the `workspaces` section in `package.json`. Note that this plugin is installed by default with yarn 4, so this may need to be removed when upgrading yarn ([see](https://www.npmjs.com/package/@yarnpkg/plugin-workspace-tools)).

The `pre-commit` script also used lerna, so it [has also been updated](https://github.com/ral-facilities/datagateway/pull/1434/files#diff-d2bc4bbf14eadc292d84e0ef5f7a93115c23557f533809fc80b896934291529d).

I have not removed `lerna-update-wizard` since it is a dev dependency and [works with yarn workspaces](https://www.npmjs.com/package/lerna-update-wizard).

Using yarn workspaces seems to generally be faster than with lerna, however execution times vary... 

## Testing instructions
Check the following commands inside top level `datagateway` package:

- [ ] `yarn run build`
- [ ] `yarn run lint`
- [ ] `yarn run tsc`
- [ ] `yarn run test:unit`
- [ ] `yarn run test:e2e`

Also check the `pre-commit` script is working.

## Agile board tracking
closes #1114


sam-glendenning:
Worth noting that, according to https://github.com/yarnpkg/berry/pull/4253 and https://github.com/yarnpkg/berry/blob/HEAD/packages/plugin-workspace-tools/README.md, `plugin-workspace-tools` will be included by default with Yarn from v4 onwards. So this is something to try and remember when we do upgrade and may mean we can remove/amend/not worry about any longer the `.yarn/plugins/@yarnpkg/plugin-workspace-tools.cjs` file from then on.

jounaidr:
> Worth noting that, according to [yarnpkg/berry#4253](https://github.com/yarnpkg/berry/pull/4253) and https://github.com/yarnpkg/berry/blob/HEAD/packages/plugin-workspace-tools/README.md, `plugin-workspace-tools` will be included by default with Yarn from v4 onwards. So this is something to try and remember when we do upgrade and may mean we can remove/amend/not worry about any longer the `.yarn/plugins/@yarnpkg/plugin-workspace-tools.cjs` file from then on.

Yeh, I came across [this](https://github.com/yarnpkg/berry/issues/735) before so there possibly might be issues if the plugin is duplicated. Do you think its worth creating an issue for this?

sam-glendenning:
Interesting. I would very much like to think that by the time v4 comes out, they would have thought of this - if we're upgrading in a place which already has plugins manually installed, what do we do with those manual installs? I would hope that during the upgrade, v4 automatically removes them and changes references etc. to point to the new location. 

That issue is nearly 3 years old so I'm hoping this has been thought of. Only way to tell at the moment is to try upgrading to a beta version of v4 to see if it already has a fix for this. But I don't think it's urgent and I don't think it needs an issue on our end yet. It would be more of a note on the issue when we do eventually make one to upgrade to v4, as something to watch out for in case it does happen.

package.json
```diff
@@ -6,24 +6,22 @@
...
+    "husky": "8.0.1"
   },
   "devDependencies": {
     "lerna-update-wizard": "1.1.0"
   },
   "scripts": {
-    "build": "lerna run build --ignore datagateway-common",
-    "lint": "lerna run lint:js -- --max-warnings=0",
-    "tsc": "lerna run tsc",
+    "build": "yarn workspaces foreach --verbose --parallel --exclude datagateway-common run build",
```
sam-glendenning, line 15: - Can we add the `--interlaced` flag to this so we can see the output?
- If we add `--jobs 3` to this, I think it might be slightly faster (it seemed that way on my machine)

package.json
```diff
@@ -6,24 +6,22 @@
...
+    "lint": "yarn workspaces foreach --verbose --parallel run lint:js --max-warnings=0",
```
sam-glendenning, line 16: - Can we add `--interlaced`?
- If we add `--jobs 3` to this, it may run faster? I think we only need 3 instead of 4 because common needs to be linted first before the other packages

package.json
```diff
@@ -6,24 +6,22 @@
...
+    "tsc": "yarn workspace datagateway-common tsc",
     "test": "yarn test:unit && yarn test:e2e",
-    "test:unit": "lerna run test --stream",
-    "test:e2e": "cd packages/datagateway-dataview && yarn run e2e && cd ../datagateway-download && yarn run e2e && cd ../datagateway-search && yarn run e2e",
-    "test:e2e:lerna": "lerna run e2e --stream --concurrency 1",
+    "test:unit": "yarn workspaces foreach --interlaced --verbose --exclude datagateway run test",
+    "test:e2e": "yarn workspaces foreach --interlaced --verbose --jobs 1 run e2e",
```
sam-glendenning, line 20: Can we add `--parallel` and `--jobs 3` to this?

package.json
```diff
@@ -6,24 +6,22 @@
...
-    "lint": "lerna run lint:js -- --max-warnings=0",
-    "tsc": "lerna run tsc",
+    "build": "yarn workspaces foreach --verbose --parallel --exclude datagateway-common run build",
+    "lint": "yarn workspaces foreach --verbose --parallel run lint:js --max-warnings=0",
+    "tsc": "yarn workspace datagateway-common tsc",
     "test": "yarn test:unit && yarn test:e2e",
-    "test:unit": "lerna run test --stream",
-    "test:e2e": "cd packages/datagateway-dataview && yarn run e2e && cd ../datagateway-download && yarn run e2e && cd ../datagateway-search && yarn run e2e",
-    "test:e2e:lerna": "lerna run e2e --stream --concurrency 1",
+    "test:unit": "yarn workspaces foreach --interlaced --verbose --exclude datagateway run test",
```
sam-glendenning, line 19: This command seems to run significantly slower than we used to with lerna. Could we add `--parallel` and `--jobs 3` to it?

sam-glendenning (commented):
Okay, I've read the docs and understood what I can of it but the below comments are all based on my interpretation that the flags `--parallel` and `--jobs` go together, but please correct me on this if you think I'm wrong here. I'm also worried about one command running in 3 parallel jobs each spawning their own set of 3 jobs and having too many threads of execution running at the same time but I'm unsure of the underlying workings. 

So anyway, feel free to give your interpretation of it since I'm probably missing one or two details!

sam-glendenning:
Basically, if you compare the runtime of the CI before and after these changes, it's slower so I'm trying to see how to improve that

package.json
sam-glendenning: Also, what is the `--exclude` flag doing here?

package.json
```diff
@@ -6,24 +6,22 @@
     "packages/*"
   ],
   "dependencies": {
-    "husky": "8.0.1",
-    "lerna": "5.5.0"
+    "husky": "8.0.1"
   },
   "devDependencies": {
     "lerna-update-wizard": "1.1.0"
```
sam-glendenning, line 12: I think we may be able to remove this? I don't think renovate relies on it to do its thing

sam-glendenning (changes requested):

sam-glendenning:
Final comment for now: lerna is also mentioned in the README and I think we can remove it with this PR :)

package.json
sam-glendenning: I don't think so... apparently it's used by running `lernaupdate` and it comes up with a window in the terminal requiring user input and stuff but I can't see this command being run anywhere at any point and I don't believe renovate uses it either. Just to double check though, we'll quickly confirm with @louise-davies if she doesn't mind chiming in here?

package.json
jounaidr: So with the exclude tag, I was having issues with that command as the `foreach` option also was picking up the top level package (rather than just the ones in `packages/*`) so it would basically result in a recursion mess (as it also would run `"test": "yarn test:unit && yarn test:e2e"`). We can either specify which packages to run explicitly, or exclude the datagateway package, and I think excluding is a tidier solution (although I haven't tested which preforms better)

package.json
louise-davies: Yeah, `lerna-update-wizard` can be removed - I used to use it before we had automated dependency updates

package.json
sam-glendenning: I thought it might be that. Your solution is just fine!

package.json
jounaidr: Okay cool, thanks.

package.json
jounaidr: Okay nice, thanks. And on the performance, I have been performance testing locally mainly and my PC didn't handle most of these commands in parallel well. Obviously this should be fine when running in the CI (as I assume its running on a beefy CPU) but I thought it would be best to optimise around running the commands locally, however it makes more sense to optimise around the CI. Ill look into this, thanks :)

sam-glendenning: Yes, I agree, it's hard to compare the two and hard to tell if I had a good run on my laptop by chance, doesn't necessarily mean it's definitely performing better. I would say try running the commands on the CI and compare the two runs - if one is better than the other, use that. Like you say, I'm personally never running the global commands, they're more for the CI

sam-glendenning:
Don't think it's the thread limits making this fail. I've seen these download tests fail before, ever since [this PR](https://github.com/ral-facilities/datagateway/pull/1404/) (not sure if that was the cause of it or not). I'll help investigate because I think there's an underlying issue here

jounaidr:
Ahh okay I see, yeh I thought it was odd this test was failing as it didn't seem related to and of the changes made in this PR, cheers.

jounaidr:
And on the yarn workspaces concurrency, by default it seems if you don't specify a max number number of threads using the `--jobs` tag it will default to half the number of CPU cores available. And I believe the CI host only has 2 cores (think its running `20.04`, specs can be found (here)[https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners]). I think its a good idea to set the thread limit to 3 as that is at most what it would need, however, at most I think it can only utilise 2 threads effectively (unless there is some sort of virtual threading I'm aware of). I don't think there is much more optimisation we could do other than upgrading the CI.

sam-glendenning:
Right okay, I wouldn't worry about it, 2 or 3 is fine for some form of concurrency. What's annoying about these failing tests is that they pass locally for me, not sure about you. Still investigating

jounaidr:
Yeh, it's working for me locally aswell :/ which is why I thought it was caused by the CI running too many threads

sam-glendenning:
I'm not sure. I think there's a lot of long-running tests on the failing test file in particular, which is why Kenneth added some longer timeouts and tried to use fake timers. I'm experimenting a bit with that right now and adding those things to the failing tests that are timing out to see if that makes a difference

sam-glendenning (approved):
Success! Double check the yarn commands and if you're happy with them, go ahead and merge :)

jounaidr:
Seems to be all good now, all passing with 3 max jobs. Think this PR is good to go

PR #1447, opened 2022-10-21
Migrate every test to use testing-library

### LINKED ISSUE #1440: Migrate everything to @testing-library
**Description**:
Migrate all unit tests to use `@testing-library` to allow React 18 upgrade.

**Acceptance criteria**:
- All the tests are using `@testing-library` and coverage remains the same.

kennethnym:
## Description
This PR migrates every test in datagateway to use `@testing-library`. I also improved some existing tests to focus on the behavior of the components rather than the implementation details of them so that they [mimic how users interact with the components](https://testing-library.com/docs/guiding-principles/).

## Testing instructions

- [ ] Review code
- [ ] Check Actions build

## Agile board tracking
Closes #1440


packages/datagateway-common/src/detailsPanels/datasetDetailsPanel.component.tsx
```diff
@@ -29,11 +29,17 @@ const DatasetDetailsPanel = (
...
+      id="details-panel"
+      container
+      direction="column"
+    >
       <Grid item xs>
         <Typography variant="h6">
           <b>{datasetData.name}</b>
-        </Typography>
+        </Typography>{' '}
+        T
```
louise-davies, line 42: I think this is a typo? Can't imagine why we suddenly have a the letter `T` here otherwise 😅

packages/datagateway-common/src/card/cardView.component.test.tsx
```diff
@@ -379,28 +399,29 @@ describe('Card View', () => {
...
-      wrapper.find('[data-testid="card-info-data-Name"]').first().text()
-    ).toEqual('Content');
+    render(<CardView {...updatedProps} />);
+
+    for (const element of await screen.findAllByText('visitId:')) {
+      expect(element).toBeInTheDocument();
+    }
+    for (const element of await screen.findAllByText('1')) {
+      expect(element).toBeInTheDocument();
+    }
```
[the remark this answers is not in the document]
kennethnym: Yeah I agree, should probably improve the test data so that the types and facilities have better, unique names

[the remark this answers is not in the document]
louise-davies: Yeah that's the other solution - ensure that any test data we check is more explicitly test data and uniquely named so `findByText` is a sufficient test

packages/datagateway-common/src/table/columnFilters/dateColumnFilter.component.test.tsx
```diff
@@ -382,15 +418,15 @@ describe('Date filter component', () => {
...
-        'Invalid date range'
-      );
+      const errorMessages = await screen.findAllByText('Invalid date range');
+      for (const element of errorMessages) {
+        expect(element).toBeInTheDocument();
+      }
     });
 
-    it('useTextFilter hook returns a function which can generate a working text filter', () => {
+    it('useTextFilter hook returns a function which can generate a working text filter', async () => {
```
louise-davies, line 429: Minor comment - but this test has the wrong name - presumably because it was copied from the text filter tests and was forgotten

packages/datagateway-common/src/views/viewButton.component.test.tsx
```diff
@@ -69,47 +65,32 @@ describe('Generic view button', () => {
...
-      .find('[aria-label="page view app.view_table"]')
-      .last()
-      .simulate('click');
-
-    expect(handleButtonChange).toHaveBeenCalledTimes(2);
-
-    expect(
-      wrapper.find('[aria-label="page view app.view_cards"]')
-    ).toBeTruthy();
   });
```
louise-davies, line 81: Is it possible to add back in the testing statements that it switches state to then be a `page view app.view_cards` button that when called switches it back to the first button?

packages/datagateway-dataview/src/page/pageContainer.component.test.tsx
```diff
@@ -399,35 +375,25 @@ describe('PageContainer - Tests', () => {
...
-    );
-
-    history.replace(paths.studyHierarchy.toggle.isisInvestigation);
-    wrapper = createWrapper();
-
-    expect(wrapper.find('PageBreadcrumbs').prop('landingPageEntities')).toEqual(
-      ['study', 'investigation', 'dataset']
-    );
+    expect(await screen.findByText('breadcrumbs.home')).toBeInTheDocument();
   });
```
louise-davies, line 392: I don't think this test is testing much here really as it is (unless we're changing this test to just check that the breadcrumbs render in pageContainer) - if we actually wanted to test this then we'd either have to mock breadcrumbs to just display the landing page entities so we can verify they're correct, or do something different (e.g. check that when we're in the study hierarchy that study breadcrumbs have links etc.)

packages/datagateway-dataview/src/views/card/datasetCardView.component.test.tsx
```diff
@@ -89,104 +88,90 @@ describe('Dataset - Card View', () => {
...
   });
 
-  it('renders correctly', () => {
-    const wrapper = createWrapper();
-    expect(wrapper.find('CardView').props()).toMatchSnapshot();
+  // TODO: this results in a huge snapshot diff, so this is skipped for now
+  it.skip('renders correctly', () => {
+    const { asFragment } = renderComponent();
+    expect(asFragment()).toMatchSnapshot();
   });
```
louise-davies, line 95: We could just change this test to test that some `Card`s are rendered as expected (e.g. with Dataset specific things)

packages/datagateway-dataview/src/views/card/isis/isisInvestigationsCardView.component.test.tsx
```diff
@@ -115,165 +112,128 @@ describe('ISIS Investigations - Card View', () => {
     jest.clearAllMocks();
   });
 
-  it('renders correctly', () => {
-    const wrapper = createWrapper();
-    expect(wrapper.find('CardView').props()).toMatchSnapshot();
-  });
-
-  it('calls required query, filter and sort functions on page load', () => {
```
louise-davies, line 123: Is there a reason why these tests were removed? Just curious

kennethnym: My reason is they are implementation details so they can break easily. But on that basis I should probably mock axios calls instead of the hook calls too, which I have done for some tests but not others, because I'm not sure whether THAT is also considered implementation details (ie API calls). My feeling is that mocking API calls is not as "implementation details" as mocking hook calls, because hypothetically if we moved on from hooks API to another API, the tests will still pass if we had mocked API calls rather than hook calls. What do you think?

packages/datagateway-dataview/src/views/card/datasetCardView.component.test.tsx
kennethnym: Yeah I am working on moving away from snapshot testing, so expect this to be redone soon.

packages/datagateway-dataview/src/page/breadcrumbs.component.test.tsx
```diff
@@ -154,33 +142,20 @@ describe('PageBreadcrumbs tests (Generic, DLS, ISIS)', () => {
...
     expect(
-      wrapper.find('[data-testid="Breadcrumb-hierarchy-1"] p').text()
-    ).toEqual('Title 1');
-    expect(wrapper.find('[data-testid="Breadcrumb-last"] p').text()).toEqual(
-      'breadcrumbs.dataset'
-    );
+      await screen.findByRole('link', { name: 'breadcrumbs.investigation' })
+    ).toHaveAttribute('href', '/browse/investigation?view=card');
+    expect(await screen.findByText('Title 1')).toBeInTheDocument();
+    expect(await screen.findByText('breadcrumbs.dataset')).toBeInTheDocument();
```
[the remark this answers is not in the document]
kennethnym: Yeah, good point. I just noticed that `Breadcrumb`s have individual indexed test IDs, so I can use that with a `getAllByTestId` regex query, and then use the resulting array to verify order.

packages/datagateway-dataview/src/views/card/isis/isisInvestigationsCardView.component.test.tsx
louise-davies: I'm fine with moving to mocking axios calls instead - just means more regex path matching. Part of the reason to test these hooks is to test that we're passing the correct params in the query i.e. includes and filters etc. - but I suppose these details are tested in the e2e tests instead and that's probably the better place for it.

packages/datagateway-common/src/card/cardView.component.test.tsx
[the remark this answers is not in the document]
kennethnym: I believe the relevant piece of code is removed in #1401, so this test is no longer relevant.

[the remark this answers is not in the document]
kennethnym: Actually, nvm, looking at the wrong piece of code.

packages/datagateway-common/src/api/retryICATErrors.ts
```diff
@@ -4,6 +4,7 @@ const retryICATErrors = (failureCount: number, error: AxiosError): boolean => {
   const message =
     (error as AxiosError<{ message?: string }>).response?.data?.message ??
     error.message;
+  console.log('error', error);
```
louise-davies, line 7: Remove logging statement

louise-davies (changes requested):
The search e2e tests are failing - we should merge in `develop` and fix the search e2e tests after merging Reilly's e2e test changes from develop

louise-davies:
@kennethnym if you wanna sanity check my last commit which fixed the search e2e tests, then feel free - otherwise I think this is good to go :)

PR #1458, opened 2022-11-10
Misc fixes

louise-davies:
## Description
A branch primarily motiviated to try and fix the download test failures, but is now doing a few misc fixes/improvements

## Testing instructions

- [ ] Review code
- [ ] Check Actions build
- [ ] {more steps here}

### DIFF
--- a/package.json
+++ b/package.json
@@ -15,9 +15,9 @@
     "test": "yarn test:unit && yarn test:e2e",
     "test:unit": "yarn workspaces foreach --interlaced --verbose --parallel --jobs 3 --exclude datagateway run test",
     "test:e2e": "yarn workspaces foreach --interlaced --verbose --parallel --jobs 3 run e2e",
-    "datagateway-dataview": "cd packages/datagateway-dataview && yarn start",
-    "datagateway-download": "cd packages/datagateway-download && yarn start",
-    "datagateway-search": "cd packages/datagateway-search && yarn start",
+    "datagateway-dataview": "yarn workspace datagateway-dataview start",
+    "datagateway-download": "yarn workspace datagateway-download start",
+    "datagateway-search": "yarn workspace datagateway-search start",
     "postinstall": "husky install"
   },
   "packageManager": "yarn@3.2.0"
--- a/packages/datagateway-dataview/src/index.tsx
+++ b/packages/datagateway-dataview/src/index.tsx
@@ -184,7 +184,7 @@ if (
           .then((response) => {
             const jwtHeader = { alg: 'HS256', typ: 'JWT' };
             const payload = {
-              sessionID: response.data.sessionID,
+              sessionId: response.data.sessionID,
               username: 'Thomas409',
             };
             const jwt = jsrsasign.KJUR.jws.JWS.sign(
--- a/packages/datagateway-search/src/index.tsx
+++ b/packages/datagateway-search/src/index.tsx
@@ -166,7 +166,7 @@ if (
           .then((response) => {
             const jwtHeader = { alg: 'HS256', typ: 'JWT' };
             const payload = {
-              sessionID: response.data.sessionID,
+              sessionId: response.data.sessionID,
               username: 'dev',
             };
             const jwt = jsrsasign.KJUR.jws.JWS.sign(


louise-davies:
Between the last 2 commits I did 7 runs of the unit tests and they passed every time, which seems more stable than previously

packages/datagateway-common/src/setupTests.tsx
```diff
@@ -15,6 +15,8 @@ import thunk from 'redux-thunk';
 import configureStore from 'redux-mock-store';
 import { createMemoryHistory, History } from 'history';
 
+jest.setTimeout(15000);
```
kennethnym, line 18: What does this `setTimeout` here do? Does it set the timeout of every test in `datagateway-common` to be 15 seconds? If you could add a comment explaining why it is needed here that would be very helpful.

packages/datagateway-dataview/src/__mocks__/loglevel.ts
```diff
@@ -0,0 +1,3 @@
+// TODO: move __mocks__ folder back to package root once facebook/create-react-app#7539 is fixed
+
+export default jest.createMockFromModule('loglevel');
```
kennethnym, line 3: I would assume this is to avoid excessive warning/error logs in the console? 😆

packages/datagateway-dataview/src/setupTests.ts
```diff
@@ -10,6 +10,8 @@ import { StateType } from './state/app.types';
 import { initialState as dgDataViewInitialState } from './state/reducers/dgdataview.reducer';
 import { dGCommonInitialState } from 'datagateway-common';
 
+jest.setTimeout(15000);
+
```
kennethnym, line 14: Same comment as above.

packages/datagateway-download/src/downloadStatus/__snapshots__/adminDownloadStatusTable.component.test.tsx.snap
```diff
@@ -1092,13 +1092,13 @@ exports[`Admin Download Status Table should render correctly 1`] = `
                           <div
```
kennethnym, line 1092: Would be nice to remove this snapshot altogether and use normal queries instead to test the table component. I wrote some util functions in #1447 (`src/setupTests.ts`) that makes querying for table elements easier. I use them extensively in testing data tables in `dg-dataview`!

packages/datagateway-download/src/setupTests.ts
```diff
@@ -1,6 +1,8 @@
 /* eslint-disable @typescript-eslint/no-empty-function */
 import '@testing-library/jest-dom';
 
+jest.setTimeout(15000);
```
kennethnym, line 4: Same comment as above!

packages/datagateway-search/src/setupTests.ts
```diff
@@ -6,6 +6,8 @@ import { StateType } from './state/app.types';
 import { dGCommonInitialState } from 'datagateway-common';
 import { initialState as dgSearchInitialState } from './state/reducers/dgsearch.reducer';
 
+jest.setTimeout(15000);
```
kennethnym, line 9: Same comment as above!

kennethnym (changes requested):
I kind of skipped over the enzyme tests changes and the snapshot changes, as I have already rewritten all of them in RTL in #1447.

packages/datagateway-dataview/src/__mocks__/loglevel.ts
louise-davies: This is to avoid the errors from "handleICATError" that are triggered when certain tests test failing request behaviour. Note I'm not mocking `console.error` so any actual errors from other libraries etc. should come through. Also, you can test that `log.error` etc. is called in your tests if you want (and some do) to test that errors are being logged - just not dirtying up the console!

packages/datagateway-download/src/downloadStatus/__snapshots__/adminDownloadStatusTable.component.test.tsx.snap
kennethnym: [These are the util functions](https://github.com/ral-facilities/datagateway/blob/enhancement/rtl-migration/packages/datagateway-dataview/src/setupTests.ts#L77-L137).

packages/datagateway-common/src/setupTests.tsx
louise-davies: I figured we occasionally have tests failing due to timeout issues, and then people add longer timeouts to those specific tests. I just kinda figured increasing the global timeout is fine - as in if a test fails it'll just take longer to fail (if it's an async) test, and it sorts out any unstable tests that are hard to figure out because the reason they're failing is due to timeouts which depend on the system at the time of running.

I could theoretically flag to only increase the timeouts in CI, but I figured it's not bad to just increase them generally for all packages.

packages/datagateway-download/src/downloadStatus/__snapshots__/adminDownloadStatusTable.component.test.tsx.snap
louise-davies: I was only here to stop the download unit tests from failing intermittently and got distracted by unit test warnings/console logging! 😅 I don't really fancy fundamentally changing any tests in this PR as that's not what it's for but I can be convinced

kennethnym: No worries, I can do it in #1447 instead 👍

packages/datagateway-common/src/setupTests.tsx
kennethnym: I see. I do wonder why tests take longer than 5 seconds sometimes. I know JSDOM isn't exactly the fastest, but it can't be that slow to the point where some tests can take more than 5 seconds to run. It just seems excessive to me.

packages/datagateway-dataview/src/__mocks__/loglevel.ts
kennethnym: I see. I agree with mocking it globally as it makes the most sense to me as well.

packages/datagateway-common/src/setupTests.tsx
louise-davies: I think the download tests specifically which were the ones that were failing often are as I detailed in the commit, failing because the tests are doing a lot in one test. e.g. in the tests they're typing in filters 2/3/4 times, and `user.type` simulates every keystroke, so potentially if stuff rerenders when the filters change that's a lot of stuff to do in 5 seconds

As I said, the alternative is to refactor those tests potentially, but there will likely always be tests that can occasionally breach 5 secs, and do we want that to then fail our CI?

louise-davies: Also, as a side effect this will help those of us who have terrible laptops (🙋‍♀️) - sometimes when running too many tests at once tests that don't normally time out do - so the global increase timeout helps as well 😅

kennethnym: Good point! I think the code is pretty self-explanatory, so I don't think adding comment is necessary.

kennethnym (approved):
LGTM!

PR #1497, opened 2023-01-09
Disable checkboxes when parent entity is in cart

### LINKED ISSUE #1472: Entity checkboxes are non-interactable if parent entity in cart selection
**Description**:
When attempting to add an entity to the selection (cart), if any of its parent entities are already in the selection, the checkbox will do nothing when we try to tick it. This is due to the way the TopCAT backend handles cart items. For example, if a dataset is in the cart, so are all of its datafiles. This is by design and is not being changed.

However, confusion arises because the children of the entity in the cart do not have their checkboxes ticked, so when visiting them, the user may be confused about whether or not they are in the cart in the first place. And when attempting to interact with these checkboxes, nothing visible happens (a POST request is actually made to the backend but it doesn't modify the cart).

To fix this visual flaw, we should add some additional logic to the checkboxes. We need to verify if any of the entity's parents (and any of its parents, and so on) are currently in the cart. If so, the checkboxes should be ticked and greyed out to prevent user interaction (since nothing happens if they are interacted with anyway). To try and provide transparency, we should also add a tooltip over the checkbox to explain that it can't be interacted with while one of its parent entities is in the cart (we can work on the wording of this). 

Similar logic should also be done in card view (and as such, the add to cart button should say "remove from cart", be greyed out and have a tooltip in the same scenario).

This issue is only present on Diamond. ISIS's backend process each checkbox tick as a separate add to cart action, meaning an entity can be downloaded alongside its parent entity as well.

[Video.webm](https://user-images.githubusercontent.com/71134574/203378113-c7b2e84c-128d-4965-97dd-7f6f6b8e142f.webm)

**Acceptance criteria**:
- [ ] A checklist

jounaidr:
## Description

This PR adds additional logic to the selection checkboxes which disables said checkboxes if the related entities parent entity is already selected. When the checkboxes are disabled, they appear greyed out and are unelectable, and a tooltip can be seen if the user hovers their mouse over the checkbox giving info as to why the checkbox is disabled.

Additional testing has been added within `selectCell.component.test.tsx` which tests the correct tooltip is displayed and that the checkbox is disabled. Despite the Codecov report indicating one line is missing coverage within [datafileTable.component.tsx](https://codecov.io/gh/ral-facilities/datagateway/pull/1497?src=pr&el=tree&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=ral-facilities#diff-cGFja2FnZXMvZGF0YWdhdGV3YXktZGF0YXZpZXcvc3JjL3ZpZXdzL3RhYmxlL2RhdGFmaWxlVGFibGUuY29tcG9uZW50LnRzeA==), it seems to be covered fine when ran locally.

**Functionality:**
chrome_2aNA68Xwj2

**Tooltip:**
image

## Testing instructions
Within `datagateway-dataview` test the following:

Investigations:
1. select investigation
2. view related datasets
3. selection checkboxes for datasets should be disabled

Datasets:
1. select dataset
2. view related datafiles
3.  selection checkboxes for datafiles should be disabled

## Agile board tracking
closes #1472


packages/datagateway-dataview/src/views/table/datafileTable.component.tsx
```diff
@@ -161,6 +161,17 @@ const DatafileTable = (props: DatafileTableProps): React.ReactElement => {
     [cartItems, selectAllSetting, allIds]
   );
 
+  const isParentSelected = React.useMemo(() => {
```
jounaidr, line 164: Could be tidied up by having a single `.some` or `.includes` rather than checking both separately with OR ...

louise-davies (changes requested):
The changes have only been applied to the "generic" tables - it also needs to be added to the ISIS and DLS tables.

Also, we probably also need similar functionality in the card view to disable the add to selection buttons when the parent entity is selected (this only affects Datasets though, as there's no datafile card views)

packages/datagateway-dataview/src/views/table/datafileTable.component.tsx
```diff
@@ -161,6 +161,21 @@ const DatafileTable = (props: DatafileTableProps): React.ReactElement => {
...
+      cartItems?.some(
+        (cartItem) =>
+          cartItem.entityType === 'dataset' &&
+          cartItem.entityId.toString() === datasetId
+      ) ||
+      cartItems?.some(
+        (cartItem) =>
+          cartItem.entityType === 'investigation' &&
+          cartItem.entityId.toString() === investigationId
+      )
```
louise-davies, line 175: Yeah, we're looping over the `cartItems` array twice here unecessarily - the following code should work equivalently:

```
cartItems?.some((cartItem) => 
  (cartItem.entityType === 'dataset' &&
   cartItem.entityId.toString() === datasetId) ||
  (cartItem.entityType === 'investigation' &&
   cartItem.entityId.toString() === investigationId)
)
```

louise-davies (changes requested):
One more thing - the "select all" checkbox in the table header also has to be disabled!

jounaidr:
Ohh you're right I didn't spot that thanks

packages/datagateway-common/src/table/cellRenderers/selectCell.component.test.tsx
```diff
@@ -72,6 +72,13 @@ describe('Select cell component', () => {
     expect(wrapper).toMatchSnapshot();
   });
 
+  it('renders correctly when selectedRows parentSelected is true', () => {
+    const wrapper = shallow(
+      <SelectCell {...selectCellProps} parentSelected={true} />
+    );
+    expect(wrapper).toMatchSnapshot();
+  });
+
```
louise-davies, line 81: Can we port this test over to `selectHeader` as well? Just for completeness :)

louise-davies (changes requested):

packages/datagateway-common/src/table/cellRenderers/selectCell.component.test.tsx
jounaidr: Oh yes sorry ill do that now

louise-davies (approved):
Looks good and it works well :)

PR #1499, opened 2023-01-17
Migrate file size/count queries to use ICAT v5 properties for investigation/dataset entities

### LINKED ISSUE #1457: Switch to using ICAT 5 filesize/counts
**Description**:
With ICAT 5, there's the new fields filesize & filecount in the `Investigation` and `Dataset` entities. These are more likely to be correct and are way more efficient that us querying `getSize`/`count` endpoints for every `Investigation`/`Dataset` we load. ISIS has already migrated to ICAT 5 and have calculated the filesizes, but DLS is still to update. Whilst we could run a hybrid model of having a switch to enable the old behaviour, if we envision fully migrating to ICAT 5 filesizes/counts - I'd prefer to just wait until DLS migrates and then we can remove all the manual size and count code.

**Acceptance criteria**:
- [ ] Use `filesize`/`filecount` fields instead of querying `getSize`/`count` endpoints

jounaidr:
## Description

With ICAT V5.0.0 the underlying file sizes and counts for dataset and investigation entities are included as properties. Previously when a file size/count was required for any table/card views, it was calculated on the fly using a query. This PR removes said query methods and where they are used and replaces them with the values directly from the received entitles.

Note that I was not able to test the changes made to `dls` table and card views, so the changes made to the respective files are assumed to work but should be double checked before this PR is approved. Also, the `fetchSize()` size buttons that were previously implemented within [dls/datasetDetailsPanel.component.tsx](https://github.com/ral-facilities/datagateway/pull/1499/files#diff-f2737f103655923da7f14c677756354076cad14649339cba8bdc5959c461ffda) and [dls/visitDetailsPanel.component.tsx](https://github.com/ral-facilities/datagateway/pull/1499/files#diff-b2f0682ecb600ea6c99ed3d329b01a96ce490d29919d58abd60dc34971362c92) have been removed since the size is no longer calculated using the query methods mentioned. Related e2e tests have been removed as well. Maybe the file count columns should be replaced with the file sizes in dls views?

The `formatCountOrSize()` method within [cellContentRenderers.tsx](https://github.com/ral-facilities/datagateway/pull/1499/files#diff-71898a7c10c7c575aec03db7f180b1dd27760e6a3b41f3248cd080dad6972aec) is now somewhat redundant as the `formatBytes()` method is now imported directly when used, so this method could possibly be refactored to only format the count (i.e `formatCount()`). However the method is used elsewhere so I think this is out of scope for this issue.

## To Finish PR

As mentioned in review comments, some of datagateway-download downloadCartTable can also be refactored to only use one query per item in the cart. An attempt has been made, with new API to request the entire entities, and to use the new fileSize/fileCount properties within downloadCartTable. Using multiple types for Datafile Datasets and Investigations seems to be working, however as expected the file count method needs changes to identify if the entity is a datafile as count property doesn't exist. [see here](https://github.com/ral-facilities/datagateway/pull/1499/files#diff-de3e3622530050ab1d21b517c13902961dcbb2d81d36a647da195f1ad3d4380cR66-R76)

## Testing instructions
Check each page where a file size or count has been refactors still correctly displays the respective info. 

## Agile board tracking
Closes #1457


packages/datagateway-common/src/app.types.tsx
```diff
@@ -9,6 +9,8 @@ export interface Investigation {
   title: string;
   name: string;
   visitId: string;
+  fileSize: number;
+  fileCount: number;
```
jounaidr, line 13: Are these okay being not possibly undefined with '?'

packages/datagateway-common/src/detailsPanels/dls/visitDetailsPanel.component.tsx
```diff
@@ -47,6 +47,7 @@ const VisitDetailsPanel = (
   const { rowData, detailsPanelResize } = props;
   const [t] = useTranslation();
   const { data } = useInvestigationDetails(rowData.id);
+  //TODO: Migrate to using entity size prop once dls has migrated to ICAT v5
```
jounaidr, line 50: This is the only place useInvestigationSize() query is used, can be removed once dls has migrated to ICAT V5

packages/datagateway-common/src/detailsPanels/dls/visitDetailsPanel.component.tsx
```diff
@@ -205,23 +195,7 @@ const VisitDetailsPanel = (
               {t('investigations.details.size')}
             </Typography>
             <Typography>
-              <b>
```
jounaidr, line 208: `fetch()` size button removed as redundant now, however I haven't test dls pages

louise-davies (changes requested):
There's also some refactoring needed in `datagateway-download` - as `downloadCartTable` displays the size + file count of each entity in the cart to both display and accumulate the total size + file count. There's some efficiency  to be had in only going through and fetching each entity in the cart once, and using the resulting `Datafiles`/`Datasets`/`Investigations` to derive both the sizes and counts - whereas currently we're sending two queries for each entity. lmk if you want to chat with me to explain this better.

Otherwise, just some minor comments and some things pending on knowing what DLS want

louise-davies:
I've addressed the review comments I had previously made and ensured all the tests are passing - we're now just waiting on a decision from Diamond on what they want to show in their views

louise-davies:
Dataset count replaced with investigation filesize

Dataset datafile count replaced with the ICAT 5 equivalent, and add a size column

Waiting confimation from Kirsty and Greg

kaperoo (approved):
Looks good to me 👍

louise-davies (approved):

Issue #1473, opened 2022-11-23
Refreshing user download table doesn't refresh progress bars


louise-davies:
**Description**:
This bug was fixed for the admin download table, but wasn't fixed for the user download table. The same invalidation of the download progress queries needs to happen in the user download table.

**Acceptance criteria**:
- [ ] Clicking the refresh button in the user download table refetches the download progress

PR #1507, opened 2023-01-31
Fix refresh button in user download table

jounaidr:
## Description
This PR applies the refresh progress bar fix that was implemented within the admin download table previously, into the user download table as well. Fix was tested via mocking the server response and returning dummy data, so further testing with an actual download object on dls would be useful. See:

chrome_UdKaUiwmAt

## Testing instructions
With download items in progress, refresh the user download table and the progress bar should update appropriately. 

## Agile board tracking
connect to #1473

### DIFF
--- a/packages/datagateway-download/src/downloadStatus/downloadStatusTable.component.test.tsx
+++ b/packages/datagateway-download/src/downloadStatus/downloadStatusTable.component.test.tsx
@@ -342,7 +342,7 @@ describe('Download Status Table', () => {
     cleanupDatePickerWorkaround();
   });
 
-  it('should show download progress ui if enabled', async () => {
+  it('should display download progress ui if enabled', async () => {
     (
       getPercentageComplete as jest.MockedFunction<typeof getPercentageComplete>
     ).mockResolvedValue(20);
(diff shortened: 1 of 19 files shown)


packages/datagateway-download/src/downloadStatus/downloadStatusTable.component.test.tsx
```diff
@@ -369,4 +369,52 @@ describe('Download Status Table', () => {
...
+
+    // pretend table is refreshed
+    renderComponent({
+      settings: {
+        ...mockedSettings,
+        uiFeatures: {
+          downloadProgress: true,
+        },
+      },
+    });
```
louise-davies, line 409: This test isn't testing the refresh button behaviour? Instead this is just rendering a new table, which is why the next assertation passes. Also, there's already a unit test for refresh behaviour - we should piggy back that test to test the progress bars as well. I'll commit the changes as I was fiddling with the code to make sure I was right that this test wasn't triggering the refresh behaviour.

jounaidr: Yeh I was struggling with that yesterday as the refresh button for the user table is in a separate component I think so couldn't be pressed, but you are right I think the way I wrote the test makes it redundant. Thanks for the changes having the test there makes more sense

louise-davies (approved):

PR #1514, opened 2023-02-10
Feature/use isis data publications #1464

### LINKED ISSUE #1464: Use DataPublications instead of Studies in ISIS landing pages
**Description**:
ISIS have/are in the progress of populating DataPublications (see https://github.com/ral-facilities/isis-icat-ingestion/issues/671), which means we can stop using the Study object for DOI landing pages and instead use DataPublications - which are all public by default and contain all the necessary info for the landing pages, which removes the permissioning issues.

We need to refactor both the study view(s) and study landing pages to use DataPublications instead.

**Acceptance criteria**:
- [ ] Migrate study landing pages & study views to DataPublications

jounaidr:
## Description
This PR migrates Study landing/table/card views to use new ICAT 5 DataPublication endpoints. The new page hierarchy can be viewed using `/browseDataPublications/instrument`, and navigating through to see the new landing/table/card views. Note that for the linked investigation, the parent DOI was removed as it seemed to now be always a duplicate.

## To finish PR

Currently all the features are implemented, however some e2e tests are failing. These are mostly to do with investigation doi links, as to fix another test and for consistency the doi's were taken from the investigations datacollection property rather than the previous studyInvestigations. An attempt at fixing this can seen [here](https://github.com/ral-facilities/datagateway/pull/1514/files#diff-ab635b7d03fae2c8d027d6a2505d445b1bf36d0a6ecff9e21e62f5617351d72eR123-R145), however i'm unsure if this will actually work. 

## Testing instructions

Check that each dataPublication page in the hierarchy loads as expected.

## Agile board tracking
closes #1464


joshuadkitenge (approved):
LGTM

PR #1522, opened 2023-04-20
Suggest similar investigations based on the currently displayed investigation on investigation landing page

### LINKED ISSUE #1521: Add suggested/similar investigations list to investigation landing pages
**Description**:
Saiful has been working on a ML reccomendation system for investigations based on their `summary` fields. We'd like to display a list of "similar" investigations when looking at an investigation (so on the investigation landing pages). See screenshot below for where the inspiration for this came from. Basic idea would be to query the API with the investigation id, and display the top 3/4/5? results in the form of a link with the Investigation title as the text and perhaps the DOI as the link address. However, some investigations don't have summaries so we should not display this for them (as otherwise it would just be nonsense recommendations)

image

See Viktor for more info

**Acceptance criteria**:
- [ ] List of suggested investigations is available on an investigation landing page (only for investigations with `summary`s)

kennethnym:
## Description
This PR adds a section on investigation landing pages that suggests similar investigations based on the currently displayed investigations. Investigation suggestions and extra info for the investigation are grouped into 2 accordions for better visual clarity. This video demonstrates the new UI:

https://user-images.githubusercontent.com/19359984/233374428-8d1c78f0-f6e1-46aa-88eb-0516023f35c4.mp4

## Testing instructions

- [ ] Review code
- [ ] Check Actions build

## Agile board tracking
Closes #1521

### DIFF
--- a/packages/datagateway-dataview/src/views/landing/isis/isisInvestigationLanding.component.tsx
+++ b/packages/datagateway-dataview/src/views/landing/isis/isisInvestigationLanding.component.tsx
@@ -1,9 +1,13 @@
 import {
+  Accordion,
+  AccordionDetails,
+  AccordionSummary,
   Box,
   Divider,
   Grid,
   Link as MuiLink,
   Paper,
+  Stack,
   styled,
   Tab,
   Tabs,
@@ -17,10 +21,15 @@ import {
   Public,
   Save,
   Storage,
+  ExpandMore,
 } from '@mui/icons-material';
 import {
+  AddToCartButton,
+  ArrowTooltip,
   Dataset,
+  DownloadButton,
   formatCountOrSize,
+  getTooltipText,
   Investigation,
   InvestigationUser,
   parseSearchToQuery,
@@ -29,16 +38,13 @@ import {
   tableLink,
   useInvestigation,
   useInvestigationSizes,
-  AddToCartButton,
-  DownloadButton,
-  ArrowTooltip,
-  getTooltipText,
 } from 'datagateway-common';
 import React from 'react';
 import { useTranslation } from 'react-i18next';
 import { useHistory, useLocation } from 'react-router-dom';
 import CitationFormatter from '../../citationFormatter.component';
 import Branding from './isisBranding.component';
+import SuggestedInvestigationsSection from './suggestedInvestigationsSection.component';
 
 const Subheading = styled(Typography)(({ theme }) => ({
   marginTop: theme.spacing(1),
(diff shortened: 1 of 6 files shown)


kennethnym:
Absolutely, do you have a particular number in mind? Obviously we don't want to show too much items, otherwise it'll crowd the page too much. Is 5 a good number to you?

agbeltran:
maybe let's show 10 - or @saifulkhan can comment on how many would be better to show at this stage (notice that this is not the final UI, it is just for demo purposes)

kennethnym:
Addressing feedback provided by @saifulkhan, I have added topic chips & pagination to the suggestion list. Demo below:

https://user-images.githubusercontent.com/19359984/234890482-e98afd44-155c-4a2c-b757-6519ef67400a.mp4

I've updated the original comment as well with the new demo.

kennethnym:
Placing the pagination control at the bottom causes it to be unstable, i.e. the control doesn't stay in the same position when switching between pages, which means the users would have to re-aim the mouse cursor between page changes.

louise-davies (commented):
Code looks good to me, maybe just add a TODO comment on the hardcoded URL to remind us for later that it needs to be changed. Only commenting for now instead of approving because we're awaiting a demo of this feature which would unblock "doing it properly" e.g. not hardcoding the URL

packages/datagateway-dataview/src/views/landing/isis/isisInvestigationLanding.component.tsx
```diff
@@ -415,27 +420,38 @@ const LandingPage = (props: LandingPageProps): React.ReactElement => {
...
+                        <ArrowTooltip
+                          title={getTooltipText(
+                            field.content(data[0] as Investigation)
+                          )}
+                        >
+                          <ShortInfoValue>
+                            {field.content(data[0] as Investigation)}
+                          </ShortInfoValue>
+                        </ArrowTooltip>
+                        z
```
louise-davies, line 449: Remove this random `z` character

louise-davies (commented):
I think I prefer the look of the links - it makes it much clearer that they're clickable and that they're links to other web pages. Theoretically, you could keep the whole button clickable and just make the text "look" like a link, but tbh I think it's just easier to have them as links.

But otherwise, I really like the chips to show the similarity score and the functionality works well :)

packages/datagateway-dataview/src/views/landing/isis/isisInvestigationLanding.component.tsx
kaperoo: I thought the z character was intentional 😅

kaperoo:
> I think I prefer the look of the links - it makes it much clearer that they're clickable and that they're links to other web pages. Theoretically, you could keep the whole button clickable and just make the text "look" like a link, but tbh I think it's just easier to have them as links.
> 
> But otherwise, I really like the chips to show the similarity score and the functionality works well :)

Changed the buttons back to links and removed the random 'z' in extra info.

louise-davies:
@VKTB this is ready from our end again

PR #1534, opened 2023-05-30
Fix date-time filtering not working correctly in download table

### LINKED ISSUE #1418: Filtering by time in download table doesn't work
**Description**:
Currently, when attempting to filter by time in the Requested Date column, the files are not filtered as expected (filtering by date works fine). See:

chrome_v4fuDDufH7
 
One possible fix is to replace lines [160](https://github.com/ral-facilities/datagateway/blob/e97d6c0e14a01a5887f04fdab584e4a5fda17bc4/packages/datagateway-download/src/downloadStatus/downloadStatusTable.component.tsx#L160) and [163](https://github.com/ral-facilities/datagateway/blob/e97d6c0e14a01a5887f04fdab584e4a5fda17bc4/packages/datagateway-download/src/downloadStatus/downloadStatusTable.component.tsx#L163) with `new Date(value.startDate).getTime() + (new Date(value.startDate).getTimezoneOffset() * 60000)`, as it could be a problem with UTC time conversion, for which this change will get the unix timestamp in the local timezone, and add/subtract the diff from UTC. However this seems to only work up to 1:00am

**Acceptance criteria**:
- [ ] Time filtering for download status table working as expected

kennethnym:
## Description
This PR addresses the issue of date filtering not working correctly in download table. Instead of converting dates to timestamps for comparison, I used util functions from `date-fns` for date comparison, although the original issue seems to be that the date string from the server contains an extra `[UTC]` at the end. The extra bit is handled correctly when displaying the dates, but seems to be ignored in the filtering logic.

## Testing instructions

- [ ] Review code
- [ ] Check Actions build

## Agile board tracking
Closes #1418

### DIFF
--- a/packages/datagateway-download/src/downloadStatus/downloadStatusTable.component.test.tsx
+++ b/packages/datagateway-download/src/downloadStatus/downloadStatusTable.component.test.tsx
@@ -1,6 +1,5 @@
 import { render, RenderResult, screen, waitFor } from '@testing-library/react';
 import userEvent from '@testing-library/user-event';
-import { UserEvent } from '@testing-library/user-event/dist/types/setup';
 import * as React from 'react';
 import { QueryClient, QueryClientProvider } from 'react-query';
 import {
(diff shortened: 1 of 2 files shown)


louise-davies (approved):
LGTM

Issue #1471, opened 2022-11-22
Keep search history for logged in users


agbeltran:
**Description**:
It would be useful, especially for users who are logged in, to keep the history (with a configurable number of search strings) of the searchers that they performed.

**Acceptance criteria**:
- [ ] The Search omponent keeps the history of the searchers previously performed by a logged in user

louise-davies:
@MRichards99 it depends on what we mean by "keep" a search history. There's a few things we could potentially do
1. leverage browser autocompleting (see https://www.w3schools.com/html/html_forms.asp) - once you submit a form then a browser will "remember" previously submitted values. This persists across browser refreshes, but is locked to a single browser (although might "migrate" across PCs if a user is logged into their browser a la Google or Mozilla accounts)
2. store in the app - then yes, we could store this in Redux or perhaps even just local component state or something, but this would be info lost on refresh or browser change.
3. Use localstorage, similar to the first solution but again doesn't persist across browsers but persists across refreshes. Means we can make our own UI though unlike the 1st solution, but that's also a downside as we have to code it rather than rely on the browser.
4. Store it on the server, which is the ultimate "user can restore their history across PCs and browsers"
   - in this case, storing it in the topcat DB makes sense

I would assume we'd have an autocomplete field if we're implementing it ourselves like how google does it, or if we go with 1 then it's the browsers that determine the UI.

louise-davies:
@MRichards99 the latter case is what we want - you don't actually need to set the `autocomplete` field at all, just having a name on the input is what matters as we aren;t any of the autocomplete types (like address, email or name). I think though that in order for the browser to "remember" submissions, we have to add a name to the input, surround the input in a form and have a button with type submit also in the form (from https://stackoverflow.com/a/68661046/7458681). `event.preventDefault()` stops the browser from reloading the page as it normally does with `form` tags. This snippet below works for me (i.e. pressing enter or clicking the button does nothing, but the browser remembers the string)

```
<form
    onSubmit={(event) => {
       event.preventDefault();
    }}
>
   <TextField name="dg-search" />
   <button type="submit" />
</form>
```

I also just added the `type="submit"` to the search button, a name to the search input, and wrapped `id='container-searchbox'` with a form with event.preventDefault onSubmit in `searchBoxContainer` and it works 🎉

MRichards99:
Ah ok, I'll try that on Friday to see if I can get that to work! 

[...]

PR #1539, opened 2023-06-21
Autocomplete Recent Searches

MRichards99:
## Description
This PR partially implements #1471. As discussed in the issue, I have implemented a way for a user to access their recent searches using [browser autocompleting](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/autocomplete):

image

As Louise stated in the issue, "this persists across browser refreshes, but is locked to a single browser". This may not be enough functionality to satisfy DataGateway users (i.e. they might want recent searches to be the same across browsers) but given the low code change in this PR, it seems this is a good first step.

I updated the snapshots for the unit tests and tried to write an e2e test but I wasn't able to get the autofill field to complete in Cypress during the test (it would appear if I did the steps manually after the test) so I was unable to write a test for that.

It's rather simple, but it seems to work.

## Testing instructions

- [ ] Review code
- [ ] Check Actions build

## Agile board tracking
connect to #1471


louise-davies (approved):
LGTM - it might cause some merge conflicts with the lucene PR but nothing too bad and we'll want this change for when we release the search stuff :)

MRichards99:
Brilliant - I think I'll merge this now so I can get (sneakily) ahead of the Lucene PR :)

Issue #1540, opened 2023-06-23
Add exact filter


louise-davies:
**Description**:
A use case in DLS has come up where it would be useful to be able to do an exact match on a text field. We should add this as an additional filter like include & exclude.

**Acceptance criteria**:
- [ ] Equals filter added as an available text filter

PR #1574, opened 2023-08-14
Feature/add exact filter #1540

kaperoo:
## Description
This PR adds an `Exact` text filter. This filter returns only the results with fields exactly matching the query.

image image

- Added e2e tests to test `include`, `exclude` and `exact` filters (datagateway-dataview).
- Had to add `// eslint-disable-next-line cypress/unsafe-to-chain-command` before some of the existing lines in `investigations.cy.ts`, otherwise cypress wouldn't let me commit.
- Added unit tests for testing `exact` filter (datagateway-common).
- Changed `aria-label` of text filters button from "include or exclude" to "include, exclude or exact"

## Testing instructions
- [ ] Review code
- [ ] Check Actions build

## Agile board tracking
connect to #1540

### DIFF
--- a/packages/datagateway-dataview/cypress/e2e/table/investigations.cy.ts
+++ b/packages/datagateway-dataview/cypress/e2e/table/investigations.cy.ts
@@ -174,50 +174,6 @@ describe('Investigations Table', () => {
   });
 
   describe('should be able to filter by', () => {
-    it('text', () => {
-      cy.get('[aria-label="Filter by Title"]').first().type('wide');
-
-      cy.get('[aria-rowcount="4"]').should('exist');
-      cy.get('[aria-rowindex="1"] [aria-colindex="4"]').contains('85');
-
-      // check that size is correct after filtering
-      cy.get('[aria-rowindex="1"] [aria-colindex="7"]').contains('3.34 GB');
-    });
-
-    it('date between', () => {
-      cy.get('input[id="Start Date filter from"]').type('2012-01-01');
-
-      cy.get('input[aria-label="Start Date filter to"]')
-        .parent()
-        .find('button')
-        .click();
-
-      cy.get('.MuiPickersDay-root[type="button"]').first().click();
-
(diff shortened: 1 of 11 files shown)


packages/datagateway-dataview/cypress/e2e/table/investigations.cy.ts
```diff
@@ -22,7 +22,7 @@ describe('Investigations Table', () => {
 
   it('should disable the hover tool tip by pressing escape', () => {
     // The hover tool tip has a enter delay of 500ms.
-    // eslint-disable-next-line cypress/no-unnecessary-waiting
+    // eslint-disable-next-line cypress/no-unnecessary-waiting, cypress/unsafe-to-chain-command
```
louise-davies, line 25: We should probably refactor the cypress tests to avoid the linting error, as it seems like a sound warning, but for now this eslint ignore is fine.

packages/datagateway-dataview/cypress/e2e/table/investigations.cy.ts
```diff
@@ -215,6 +217,58 @@ describe('Investigations Table', () => {
     });
   });
 
+  describe('should be able to use text filters', () => {
```
louise-davies, line 220: A minor point - these tests should probably be in the `should be able to filter by` section, and replace the `text` test with this describe block (as we don't need the existing text filter test as this is tested by the include test case in this new block)

louise-davies (commented):
Functionality works as expected - just a minor comment on rearranging some of the cypress tests :)

packages/datagateway-dataview/cypress/e2e/table/investigations.cy.ts
kaperoo: Moved the tests to `should be able to filter by` section and removed `text` test.

louise-davies (approved):
LGTM :)

PR #1575, opened 2023-08-18
Improve e2e tests & fix sorting bug

louise-davies:
## Description

- Fix sorting bug where React Query considered two requests for columns sorted in a different order (e.g. name asc first, title desc second and title desc first, name asc second) the same request, and so would incorrectly get the data from cache rather than querying
- Fix sorting bug in that the formatted columns in the admin download table couldn't be sorted
- Fix filtering bug in download status table where only the first filter you applied was checked. Now, you need to satisfy all the applied filters to see the row
- Fix filtering bug in download status table where invalid date ranges threw exceptions and crashed the app. Now, this no longer happens
- Remove linting errors
- Delete unneeded tests
- Improve efficiency by consolidating tests into fewer, longer tests
(see https://docs.cypress.io/guides/references/best-practices#Creating-Tiny-Tests-With-A-Single-Assertion)
  - This has significantly affected the length of the dataview tests - probably around a 33% reduction I would estimate
- lint the cypress code in the CI job
- Fixed some tests that just didn't pass locally against sg-preprod for various reasons - now all tests should pass both on CI and locally

## Testing instructions

- [ ] Review code
- [ ] Check Actions build
- [ ] {more steps here}

### DIFF
--- a/packages/datagateway-common/package.json
+++ b/packages/datagateway-common/package.json
@@ -43,6 +43,7 @@
     "react-router-dom": ">= 5.2.0 < 6"
   },
   "devDependencies": {
+    "@babel/eslint-parser": "7.22.5",
     "@mui/icons-material": "5.11.0",
     "@mui/material": "5.11.0",
     "@testing-library/jest-dom": "5.16.4",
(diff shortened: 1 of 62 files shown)


joshuadkitenge (approved):
All the bug described are fixed. code LGTM

Issue #1582, opened 2023-09-13
Extra request is sent when a table is loaded before the default sort is applied


louise-davies:
**Description**:
We need to change how we implement the default sort functionality - as currently the page loads and sends a request without the sort applied and then shortly after it sends it with the default sort. We shouldn't be sending both requests - we shouldn't send the "unsorted" request at all. This may be causing performance issues on DLS. The same issue occurs in CardView (mainly because both use similar logic)

Steps to reproduce:
If you're looking at a table with a default sort (e.g. DLS datafile table), delete the sort param from the URL and refresh the page. Look at the networking tab, and see that two requests are being sent.

**Acceptance criteria**:
- [ ] When a table/card view with a default sort loads, it only sends one query to the backend for the list of entities, and it's the query with the default sort applied

PR #1583, opened 2023-09-19
Fix double requests on default sort

kaperoo:
## Fixed double requests bug on default sort
Any table/card view that uses default sorting now checks whether it has been mounted to prevent data requests being sent twice.
The `isMounted` variable is used to disable or enable queries in datagateway-common/api.

## Testing instructions
Updated unit tests to check for double requests when default sort is specified.

- [ ] Review code
- [ ] Check Actions build

## Agile board tracking
connect to #1582

### DIFF
--- a/packages/datagateway-common/src/api/dataPublications.tsx
+++ b/packages/datagateway-common/src/api/dataPublications.tsx
@@ -102,6 +102,7 @@ export const useDataPublicationsPaginated = (
         handleICATError(error);
       },
       retry: retryICATErrors,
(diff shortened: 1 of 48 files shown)


louise-davies (approved):
LGTM - tried it out and it only sends the default sort-ed request, fixes the issue in an efficient way, and it's tested in the unit tests. @kevinphippsstfc for reference, this is the bug you raised and it's now got a fix if we need to hotfix this before the next DG release then let me know.

PR #1584, opened 2023-09-28
Release v1.1.2

### LINKED ISSUE #1582: Extra request is sent when a table is loaded before the default sort is applied

louise-davies:
## Description
Hotfix release to fix #1582 on DLS production.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,13 @@
 # Changelog
 
+## [v1.1.2](https://github.com/ral-facilities/datagateway/tree/v1.1.2) (2023-09-28)
+
+[Full Changelog](https://github.com/ral-facilities/datagateway/compare/v1.1.1...v1.1.2)
+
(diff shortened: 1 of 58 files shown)


kaperoo (approved):
Looks good to me 😊

PR #1594, opened 2023-11-17
Improve unit tests & fix text previewer styling

louise-davies:
## Description
See title

The top level test:unit command now runs all packages in band in Jest (i.e. Jest runs tests serially not in parallel) - this is recommended on CI and may reduce failure rates. Also increased jobs from 3 to 4 to run all suites in parallel.

Also, I got annoyed with console errors again so I fixed those and added a library that will fail a test if it generates logging.

Also, in fixing some logging in datafilePreviewer, I noticed that the styling for the text previewer was messed up. Line numbers took up half the view. I refactored so that the line numbers take up only the space they need, with the rest going to the actual text content.

## Testing instructions

- [ ] Review code
- [ ] Check Actions build

### DIFF
--- a/packages/datagateway-dataview/src/views/table/investigationTable.component.test.tsx
+++ b/packages/datagateway-dataview/src/views/table/investigationTable.component.test.tsx
@@ -140,7 +140,7 @@ describe('Investigation table component', () => {
 
               return Promise.resolve({
                 data: {
-                  cardItems: [],
+                  cartItems: [],
                 },
               });
             }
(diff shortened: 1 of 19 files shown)


packages/datagateway-dataview/src/views/table/investigationTable.component.test.tsx
```diff
@@ -171,7 +171,7 @@ describe('Investigation table component', () => {
     jest.clearAllMocks();
   });
 
-  it('renders correctly', async () => {
+  it.skip('renders correctly', async () => {
```
kaperoo, line 174: Do we keep these skips in?

louise-davies: oops no - they were left over from my debugging

kaperoo (approved):
Looks good to me :)

Issue #1596, opened 2023-11-20
Add upload functionality to DLS tables


louise-davies:
**Description**:
Add the ability to upload new data to ICAT for DLS via Alex's upload server: https://github.com/ral-facilities/user-upload-service

This requires using the Uppy framework: https://uppy.io/docs/react/

Wireframes are on Balsamiq: https://balsamiq.cloud/sq9hezy/p1wt1o8/rC2E7

But basic idea - Visits table should let you upload a new Dataset to a Visit, Dataset table should allow uploading new Datafiles to that Dataset. This should be done via the `actions` prop of table/cardview components - see `datagateway-common/src/views/downloadButton` for an example. Also, on the Datafiles table should also be able to upload - although this is via a separate button near the clear filters button (saves the user having to go back up to the dataset table).

**Acceptance criteria**:
- [ ] DLS users can upload data to ICAT via the user-upload-service on all the appropriate views

PR #1597, opened 2023-11-29
Feature/dls user upload #1596

kaperoo:
## Description
Enter a description of the changes here

## Testing instructions

- [ ] Review code
- [ ] Check Actions build
- [ ] {more steps here}

## Agile board tracking
connect to #{issue number}


moonraker595:
Hi, I'm going to add some comments here while I integrate the FE back into a containerised environment.

It would be good to have the upload URL set as an environment variable. At the moment I believe its hard coded to `http://127.0.0.1:8181/upload/`

Something like:
```
datagateway:
    image: harbor.stfc.ac.uk/datagateway/plugins:pr-1597
    container_name: datgateway_container
    environment:
      - DOWNLOAD_API_URL=https://scigateway-preprod.esc.rl.ac.uk:8181/topcat
      - ICAT_URL=https://icat_payara_container:8181/icat
      - IDS_URL=https://scigateway-preprod.esc.rl.ac.uk:8181/ids
      - UPLOAD_URL=https://whatever_my_upload_backend_is:8080/
```

kaperoo:
`uploadUrl` should now be set as an environment variable.

moonraker595:
Hi, Great work on this and having the environment variable configurable makes deployment from a docker-compose file so much easier. 

As we're getting closer to demoing this for Diamond. I think we should smarten the UI up a bit so that it looks a bit more professional. The reviewers are likely to be Steve and Kirsty who aren't technical folk, meaning all they'll mostly care about is look and feel. So with that in mind, I've added some comments below. I remember there were some limitations which we'll need to work on but I can't remember them exactly, so it'll be good to get these down. 

The plan is to have a brief demo but most of the reviewing I expect to happen offline. I plan on forwarding the pre-prod url for them to play around with.

I have this working in pre-prod now: http://host-172-16-114-165.nubes.stfc.ac.uk/ . This is what I'm basing these comments on.

1. Is there any chance we can get rid of the errors that keep on popping up? Originally, the error was a facility one based on topcat, so I removed the add-on but now I get this error. Is there any kind of configuration that would suppress these errors? Would it be easier to deploy an instance of topcat? Has one been dokerised? 
2. The "Upload datafiles to dataset" works but I can't get the "Upload Dataset to Investigation" to work. Looks like an uncaught promise due to an icat error. I'll double check things on my side, maybe it's something I've changed.
3. If I close this, go into the investigation and upload datafiles to datasets, it correctly tries to restore my session to resume the upload but fails the upload itself. Starting the process again works.
4. I think the biggest issue is the upload popup, there's no canvas and the css doesn't seem to be laid out correctly. I've attached some pictures. 

I remember we discussed this briefly but these are the things they'll want to see. If we can't use any of the fancy uppy features, maybe we could smarten up what we have.

It was mentioned in one of the Facilities data pipeline meetings that Kacper wanted to do some backend work (?) so I'm happy to get stuck into the front end under your guidance @louise-davies . Let me know what suits you both. Thanks.

Screenshot 2024-03-07 at 17 15 37
Screenshot 2024-03-07 at 16 34 22

louise-davies:
The Uppy CSS wasn't properly being included in the production build, hence the UI looking very strange! I've pushed a fix for that now,

packages/datagateway-common/src/api/datasets.tsx
```diff
@@ -284,3 +284,26 @@ export const downloadDataset = (
...
+
+  const params = {
+    investigationId: investigationId,
+    datasetName: name,
+    datasetDescription: description,
+  };
+
+  //TODO: remove hardcoded url
+  return axios
+    .post('http://127.0.0.1:8181/dataset', params, {
```
louise-davies, line 303: There's another hardcoded url here - I presume this is meant to be `uploadUrl`?

kaperoo:
I removed the hardcoded dataset creation URL. 
Note that the upload URL in the settings/environment variable should now just be the domain without the `/upload/` path (i.e. `http://127.0.0.1:8181` instead of `http://127.0.0.1:8181/upload/`).

kaperoo:
Hi, we should probably all catch up sometime soon to discuss what would be the best way to check for the existing datasets/datafiles.

Also, I have noticed some problems with the current implementation:
- On the pre-prod provided by @moonraker595 it is currently impossible to upload files larger than 1mb.
image

- We still have the bug where whenever you try to re-select a larger file (i.e. after the browser crashes) we get the "exceeds maximum allowed size" error. This is probably due to the way the ["ghost files"](https://uppy.io/blog/2017/07/golden-retriever/) and `maxTotalFileSize` check are handled in `@uppy/core/src/Restricter.js`, and there is no way we can fix this (unless we make a patch? @louise-davies ).

- Even if we are able to re-select the file, the upload will fail because a partially uploaded file with the same name already exists in the dataset. Maybe we should consider doing rollbacks on failed uploads?

moonraker595:
Hi, thanks for this. I'll put something in our calendars for next week to discuss further.

Some notes:
- Point 1: I had tested the uploading of files **_locally_** up to 50GB so this looks to be a deployment/reverse proxy issue. I'll look into it.

- Point 2: Is this where the Golden Retriever plugin came in? I don't think we'd want to start patching third-party tools unless this is something you'd regularly do. Hopefully, this issue will get pushed into an edge case when we fix the above issue, as in, a user would need to have a failed upload at the same time as hitting the `maxTotalFileSize` / 2.

- Point 3: We would need to think about the workflow:

1. We should check that the datasets and datafiles are unique first as it shouldn't take long. Using the suggested idea of some new endpoint to check this, we could only allow the upload button to be active once this has been cleared. If not, we'd need to display a message containing the offending items. I think this would be separate from uppy so would need somewhere on the page to indicate this.
2. For the actual uploading of a set of files: We could upload the files first, if all succeed, then create the ICAT items. At the moment this is happening per each uploaded item. This would mean we'd need to separate this functionality by having another endpoint to call once the files have been successfully uploaded, something like `mint_upload` or similar. The end goal is to move this data from s3 to tape. For this, we'd check ICAT for successfully uploaded items, get the location and do the transfer. If the upload has failed for whatever reason, the items won't exist in ICAT and so won't get transferred. The S3 store will get wiped periodically so there'd be no need to undo (or delete) the uploaded files. However, If one file fails to upload, where others succeed, It shouldn't be a size issue because this is calculated beforehand by uppy and it shouldn't be a naming issue because this would have been checked beforehand. The user would then have the option to "retry" that particular file, and the session should be retrievable through the uppy framework (I think!) 

I'm assuming all of this is possible from the FE which it may not be.

Thanks,

Alex

moonraker595:
Outcome from FE/Backend integration meeting on 19/03/24

### Problem 1: Filenames already exist in ICAT
Because the ICAT filenames are unique, the user may upload a file that already exists in the catalogue. If this is part of a group of files to be uploaded, this would result in a partial upload and would require the user to go back into the dataset and re-upload the files with a new name. It is also difficult to relay this specific information back up to the front end.

**Solution:**
It would be better to inform the user upfront if any files they select already exist in ICAT.

- Kacper to see if file names can be checked using the DGWapi
- Otherwise a new endpoint /datafiles/verify will be needed to check an array of filenames and see if they already exist in ICAT.

 ### Problem 2: ICAT entries but no files in the object store.
If the upload fails for any reason (bandwidth, object store issues, basically any problem other than the above problem 1), then the ICAT entities will exist in the catalogue but not in the object store. This is because the icat entities are done tus.io's `onUploadCreate` which is the only blocking call to create them with.

**Solution:**
To prevent any complicated rollback procedure, the ICAT entities should only be created once all files have successfully been uploaded. If the upload fails, then the uploaded files that have made it through will be wiped when the object store (essentially a cache) data is transferred to tape.

-  Alex to create a new endpoint `upload/commit` which would enter the ICAT datafiles separately. It would need to know the upload ID (for its location), the file size and the file name to pass to the backend.

### Problem 3: DGW signing users out before large uploads complete
The speed at which files are uploaded is largely dependent on the internet connection. Users with good/high-speed connections shouldn't be limited to smaller sizes to accommodate those that don't. 

**Solution:**
- Kacper to refresh the user session before the upload, ensuring that the user has the full 2 hours (currently believed to be the icat session timeout) to upload larger files

moonraker595:
Outcome from FE/Backend integration meeting on 19/03/24

### Problem 1: Filenames already exist in ICAT
Because the ICAT filenames are unique, the user may upload a file that already exists in the catalogue. If this is part of a group of files to be uploaded, this would result in a partial upload and would require the user to go back into the dataset and re-upload the files with a new name. It is also difficult to relay this specific information back up to the front end.

**Solution:**
It would be better to inform the user upfront if any files they select already exist in ICAT.

- Kacper to see if file names can be checked using the DGWapi
- Otherwise a new endpoint /datafiles/verify will be needed to check an array of filenames and see if they already exist in ICAT.

 ### Problem 2: ICAT entries but no files in the object store.
If the upload fails for any reason (bandwidth, object store issues, basically any problem other than the above problem 1), then the ICAT entities will exist in the catalogue but not in the object store. This is because the icat entities are done tus.io's `onUploadCreate` which is the only blocking call to create them with.

**Solution:**
To prevent any complicated rollback procedure, the ICAT entities should only be created once all files have successfully been uploaded. If the upload fails, then the uploaded files that have made it through will be wiped when the object store (essentially a cache) data is transferred to tape.

-  Alex to create a new endpoint `upload/commit` which would enter the ICAT datafiles separately. It would need to know the upload ID (for its location), the file size and the file name to pass to the backend.

### Problem 3: DGW signing users out before large uploads complete
The speed at which files are uploaded is largely dependent on the internet connection. Users with good/high-speed connections shouldn't be limited to smaller sizes to accommodate those that don't. 

**Solution:**
- Kacper to refresh the user session before the upload, ensuring that the user has the full 2 hours (currently believed to be the icat session timeout) to upload larger files

moonraker595:
Diamond have agreed on the first set of restricted files for uploads: `.xml, .exe, .dll, .bat, .sh, .sqlite, .js, .vbs, .PHP, .wmv, mp3, .flv` file types. I'm sure more will follow. Can you add these to the list please @kaperoo?

moonraker595:
Hi @kaperoo & @louise-davies . I've made the changes to the upload service as discussed above. In summary, these are:
- I've removed the `/dataset` endpoint
- now the `/commit` endpoint is to be called after the successful upload of datafiles or datasets
- the `/commit` schema is [here](https://github.com/ral-facilities/user-upload-service/blob/275c380bc34c8a8947003ae3fc343d239577d48f/backend/schemas.js#L29). It expects a dataset object OR datasetIds in the dataFiles object, but not both. 
- If a dataset object is provided, it'll create a dataset and use that for the datafiles
- Both the `/commit` endpoint and the `/upload` endpoint now expect the user session to be in the auth headers bearer token.
- The upload endpoint doesn't need any extra metadata now

You should be able to run everything up for testing using the [docker-compose file](https://github.com/ral-facilities/user-upload-service/blob/commit_endpoint/docker-compose.yaml), making sure to uncomment the `upload_service:` section and point it to the branch tag in harbor `:commit_endpoint`.

Let me know if you run into any trouble 👍

moonraker595:
Hi @kaperoo, @louise-davies  . Are you able to get the last modified time/date as well as the creation date for each datafile? There are [entities](https://repo.icatproject.org/site/icat/server/6.0.0/schema.html#Datafile) for them in the icat schema and that information might be useful down the line. I've found [this](https://developer.mozilla.org/en-US/docs/Web/API/File) which would allow you to do something like:
```
uppy.on('file-added', (file) => {
  // Access the lastModified property
  const lastModifiedDate = new Date(file.data.lastModified);
});
```
But not sure about the creation date

kaperoo:
A couple of changes and questions:
1) I have updated the list of restricted file types.
If there are more restricted file types in the future, maybe we could use the allowedFileTypes from uppy and only specify the ones we allow?

2) We now use the `/commit` endpoint after a successful upload.
If we upload multiple files and some fail, should we only commit the successful ones or none at all?

3) Added checks for dataset name and duplicate files in the upload queue.

4) We can easily add the _Last Modified_ date of the file, but the _Creation Date_ may [not](https://stackoverflow.com/questions/27800376/how-to-get-file-creation-date-on-browser-using-javascript-or-jquery) be possible. 
If I were to send the LM date, should I do it in the file meta or in the call to `/commit`?

moonraker595:
That’s great, thanks.


  1.  allowedFileTypes sounds like a better way of doing this. For now, I’d say stick to the following. It would be good if these were configurable through some sort of config file, or do you have enough config to worry about?
     *   images files

                                          i.    JPEG (.jpg or .jpeg)

                                         ii.    PNG (.png)

                                        iii.    GIF (.gif)

                                        iv.     BMP (.bmp)

                                         v.    TIFF (.tif or .tiff)

                                        vi.    WebP (.webp)

                                       vii.    HEIF (.heic or .heif)

     *   hdf files

                                          i.    .hd5, .hdf5

     *   txt files
                                          i.    .txt -
                                         ii.    .csv
                                        iii.    .json
                                        iv.    .md - Markdown file
                                         v.    .log - Log file

  1.  It depends, I’m concerned about the “retry” function, will that try and upload everything again?, if so then we could get duplicate entries in ICAT. Maybe it’s safer to not commit anything if any uploads fail.
  2.  -
  3.  in the call to /commit please, let me know what’s definitely possible and I’ll update the schema to accommodate.

From: Kacper Roemer ***@***.***>
Date: Wednesday, 10 April 2024 at 12:04
To: ral-facilities/datagateway ***@***.***>
Cc: Kemp, Alexander (STFC,RAL,SC) ***@***.***>, Mention ***@***.***>
Subject: Re: [ral-facilities/datagateway] Feature/dls user upload #1596 (PR #1597)

A couple of changes and questions:

  1.  I have updated the list of restricted file types.
If there are more restricted file types in the future, maybe we could use the allowedFileTypes from uppy and only specify the ones we allow?
  2.  We now use the /commit endpoint after a successful upload.
If we upload multiple files and some fail, should we only commit the successful ones or none at all?
  3.  Added checks for dataset name and duplicate files in the upload queue.
  4.  We can easily add the Last Modified date of the file, but the Creation Date may not<https://stackoverflow.com/questions/27800376/how-to-get-file-creation-date-on-browser-using-javascript-or-jquery> be possible.
If I were to send the LM date, should I do it in the file meta or in the call to /commit?

moonraker595:
> A couple of changes and questions:
> 
>     1. I have updated the list of restricted file types.
>        If there are more restricted file types in the future, maybe we could use the allowedFileTypes from uppy and only specify the ones we allow?
> 
>     2. We now use the `/commit` endpoint after a successful upload.
>        If we upload multiple files and some fail, should we only commit the successful ones or none at all?
> 
>     3. Added checks for dataset name and duplicate files in the upload queue.
> 
>     4. We can easily add the _Last Modified_ date of the file, but the _Creation Date_ may [not](https://stackoverflow.com/questions/27800376/how-to-get-file-creation-date-on-browser-using-javascript-or-jquery) be possible.
>        If I were to send the LM date, should I do it in the file meta or in the call to `/commit`?

Have these changes been pushed to harbor? I'm not seeing the call to `/commit`

kaperoo:
Hi, thanks for your comment @moonraker595 . 

We've moved the call to `/commit` to the upload post-processing. This means that the files will not "turn green" until the `/commit` call has been successful. This also fixes the table refresh issue.

Because of this (but mostly because of another annoying quirk of uppy), we need to make separate commit calls for the initial upload and (if some files failed) for each successful retry. To do this, we need the response from the initial call to include the id of the created dataset.
Currently we have made this work by slicing the id from the response message.

moonraker595:
Thanks @kaperoo. The most recent change to the backend code doesn't return any dataset ids, just the new datafile ids that were created ( i didn't think you needed it at the time). So:
- do you want me to return the dataset id if one was created? So something like:
`{
  "datasetId": 12345,
  "datafiles": [
    1,
    2,
    3
  ]
}
` 
or `{
  "datafiles": [
    1,
    2,
    3
  ]
}
` for when no dataset has been created
- Or i can split out the commit endpoint into 2 if thats easier? So one to create a dataset and one to commit.

kaperoo:
I don't think we need a separate endpoint for that. The first option you mentioned sounds great. 
If it is easier for you, you can always return the existing/newly created datasetId, but we only need it for failed dataset upload retries.

Thanks

moonraker595:
Hi, this has been implemented and should be in the latest image in harbor. Note : It'll always return the `datasetId`.  The `datafiles` will always be in an array, even if there's one: `{ "datasetId": 12345, "datafiles": [ 1 ] }`

PR #1640, opened 2024-02-02
Release v1.1.3

louise-davies:
## Description
Fix performance issue of querying for investigation id in datafile tables by nstead checking the investigation id in id check functions

This release has already been deployed on DLS preprod and prod and works fine - just need to check that CI is happy


louise-davies:
One e2e test is failing, but since it's just checking by desc order on the studies table, and everything else works, it's probably a quirk in the old test data somehow

Issue #1662, opened 2024-08-09
increase TextFilter debounce timer to 500ms


akintomiwa:
the timeout needs increasing from 250 to 500

PR #1666, opened 2024-08-09
increase TextFilter debounce timer to 500ms #1662

akintomiwa:
## Description
increase TextFilter debounce timer to 500ms

## Testing instructions

- [ ] Review code
- [ ] Check Actions build
- [ ] {more steps here}

## Agile board tracking
connect to #1662

### DIFF
--- a/packages/datagateway-common/src/table/columnFilters/textColumnFilter.component.test.tsx
+++ b/packages/datagateway-common/src/table/columnFilters/textColumnFilter.component.test.tsx
@@ -12,7 +12,7 @@ import userEvent from '@testing-library/user-event';
 
 jest.mock('../../api');
 jest.useFakeTimers('modern');
-const DEBOUNCE_DELAY = 250;
+const DEBOUNCE_DELAY = 500;
 
 describe('Text filter component', () => {
   let user: UserEvent;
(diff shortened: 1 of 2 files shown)


packages/datagateway-common/src/table/columnFilters/textColumnFilter.component.test.tsx
```diff
@@ -12,7 +12,7 @@ import userEvent from '@testing-library/user-event';
 
 jest.mock('../../api');
 jest.useFakeTimers('modern');
-const DEBOUNCE_DELAY = 250;
+const DEBOUNCE_DELAY = 500;
```
louise-davies, line 15: It might be worth moving this variable to the actual component, and export it to use it in both the tests and when passing the actual debounce - it means in the future if we edit the debounce we only have to change it in one place :) Just an opportunity to refactor the code to be a little more future-proof

Otherwise good work!

akintomiwa: Thanks Louise! Okay I understand, I've made and pushed the change.

louise-davies (approved):
Looks good to me (LGTM) :)

PR #1677, opened 2024-09-25
Allow query retry setting to be configurable

louise-davies:
## Description
Adds a setting to all plugins that allows the sysadmin to set the react query query retry behaviour to whatever they want. I just let them set any number, including 0 to effectively disable retries even though setting false would have the same effect as keeping it as a number made typing easier.

This was requested by DLS as they have problems with queries timing out and they generally don't want to retry them, at the very least not 3 times like the current default. I opted away from limiting it to just problematic queries or from trying to figure out error messages indicating a query timeout and just made it a global config option.

## Testing instructions

- [ ] Review code
- [ ] Check Actions build

### DIFF
--- a/packages/datagateway-common/src/api/retryICATErrors.ts
+++ b/packages/datagateway-common/src/api/retryICATErrors.ts
@@ -1,7 +1,7 @@
 import { AxiosError } from 'axios';
 import { useQueryClient } from 'react-query';
 
-export const createRetryICATErrors = (
+const createRetryICATErrors = (
   retries: number
 ): ((failureCount: number, error: AxiosError) => boolean) => {
   return (failureCount: number, error: AxiosError) =>
@@ -26,7 +26,7 @@ const baseRetryICATErrors = (
   return true;
 };
 
-const useRetryICATErrors = (): ((
+export const useRetryICATErrors = (): ((
   failureCount: number,
   error: AxiosError
 ) => boolean) => {
@@ -34,9 +34,12 @@ const useRetryICATErrors = (): ((
   const opts = queryClient.getDefaultOptions();
   // TODO: do we want to be more elegant in handling other types of retry...
   const retries =
-    typeof opts.queries?.retry === 'number' ? opts.queries.retry : 3;
+    typeof opts.queries?.retry === 'number'
+      ? opts.queries.retry
+      : // explicitly handle boolean case as we set this in tests
+      opts.queries?.retry === false
+      ? 0
+      : 3;
 
   return createRetryICATErrors(retries);
 };
-
-export { useRetryICATErrors };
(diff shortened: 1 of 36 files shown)


packages/datagateway-common/src/api/retryICATErrors.ts
```diff
@@ -1,17 +1,45 @@
...
 };
 
-export default retryICATErrors;
+export const useRetryICATErrors = (): ((
+  failureCount: number,
+  error: AxiosError
+) => boolean) => {
+  const queryClient = useQueryClient();
+  const opts = queryClient.getDefaultOptions();
+  // TODO: do we want to be more elegant in handling other types of retry...
```
joelvdavies, line 35: Is this todo still needed?  I think this is fine, since as you say 0 is the same as false here, and it looks like the only other possible type is a function which we can't set in the config, unless I am missing something?

joelvdavies (commented):
Tested locally worked well for me, just one query.

packages/datagateway-common/src/api/retryICATErrors.ts
louise-davies: Yeah I wrote this TODO before I handled false explicitly, and handling true makes little sense (unless I set it to infinity) and same with functions. Plus assuming retry of 3 is what the old function did. I'll remove the TODO

packages/datagateway-dataview/src/App.tsx
```diff
@@ -106,6 +106,27 @@ const queryClient = new QueryClient({
...
+      const opts = queryClient.getDefaultOptions();
+      queryClient.setDefaultOptions({
+        ...opts,
+        queries: { ...opts.queries, retry: queryRetries },
+      });
+    }
+  }, [queryClient, queryRetries]);
+
+  return null;
+};
```
joelvdavies, line 128: Could this be put in common instead of replicating it in each package?

louise-davies: Technically I can extract the search and dataview ones as they're the same, but download is different due to not storing the settings in redux. At that point I wondered if it's just easier to keep each one defined explicitly where it's used, but happy to extract out search & dataview's implementation to common if you think it makes things tidier

joelvdavies: Ah right, I didn't see that. I think it would be nice to at least have the identical ones in common. Not sure what you mean by the second option.

joshuadkitenge: It might be cleaner to extract the shared functionality into a common component and pass the queryRetries as a prop, keeping Redux and Context-specific logic separate.

Then it can used like this in the relvant App.tsx 
```
export const QueryClientSettingUpdaterContext: React.FC<{
  queryClient: QueryClient;
}> = ({ queryClient }) => {
  const { queryRetries } = useContext(DownloadSettingsContext);

  return (
    <QueryClientSettingUpdaterCommon
      queryClient={queryClient}
      queryRetries={queryRetries}
    />
  );
};
```

louise-davies: @joshuadkitenge yep that was my current plan, and maybe even have the hook up to redux in common so search & dataview can have the same one but download can still share most of the code

joelvdavies (approved):
LGTM

PR #1697, opened 2025-02-20
Fix default filter on my data DLS table

louise-davies:
## Description
The default filter only partially worked right - because the datepickers have internal state, updating the filter as we did applied the filter i.e. to the search params in the URL & thus to react-query - but the date picker itself showed blank.

Thus I made the code similar to how `defaultSort` works - it's specified at the column definition level, and it gets passed to the filter components to ensure they initialise with the default value if provided.

Also removed the filtering query from the default JSON file as this was not dynamic and set the date to some point in 2022.

https://github.com/user-attachments/assets/0f52e4c5-d1e4-4252-89ce-4c7d42ef7b03

## Testing instructions

- [ ] Review code
- [ ] Check Actions build
- [ ] Check default filtering in my data DLS table works correctly

### DIFF
--- a/packages/datagateway-common/src/api/index.tsx
+++ b/packages/datagateway-common/src/api/index.tsx
@@ -364,10 +364,11 @@ export const useSingleSort = (): ((
   );
 };
 
-export const usePushFilter = (
+const useFilter = (
+  updateMethod: UpdateMethod,
   filterPrefix?: string
 ): ((filterKey: string, filter: Filter | null) => void) => {
-  const { push } = useHistory();
+  const { push, replace } = useHistory();
 
   return React.useCallback(
     (filterKey: string, filter: Filter | null) => {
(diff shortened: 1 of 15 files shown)


joelvdavies (commented):
Mostly worked well for me. I did find that changing the start date to field, then refreshing the page reset it to the default again so a user sharing a link would not recover the same state.

louise-davies:
@joelvdavies hmm yeah that's a problem - I need to fix the code where we apply the filter to check for any existing filters - good spot!

louise-davies:
@joelvdavies I just noticed the same thing happens with `defaultSort` - we only check if there's not already a sort on the existing column, so if e.g. you have it sorted by size it will override back to the default sort. I'll fix that issue as well as they're both quite similar. We should just only apply default sorts/filters when the sort/filter state is completely empty

joelvdavies (dismissed):
LGTM

joelvdavies:
It still has the same issue if you clear the date filter leaving the url as http://localhost:3000/my-data/DLS?sort=%7B%22visitId%22%3A%22asc%22%7D and then refreshing, but that's minor compared to before.

louise-davies:
@joelvdavies that's where we have an "issue" on how the default sort/filter is supposed to work. We could make it so we only apply the default sort if the filters are also empty and vice versa for default filters  - which has the effect of only applying the filters if the URL is blank. However, we do optimise the default sorting by hard coding the sort into the links, so we'd have to remove those as otherwise default filtering would not work.

I think this is fine for now - it's no worse than the previous behaviour. If users complain then we can ensure it only applies to blank query params

PR #1701, opened 2025-03-03
Fix exact filters & add is deleted filter

louise-davies:
## Description
Exact filters were not working in the admin download table &  the download cart, so fixed that and refactored all the places we check for the type of the filter to ensure we are handling all use cases.

Also added a is deleted filter to the Deleted column on the admin download table - requested by Sabah. Implemented as a simple yes/no dropdown.

Also I noticed some tables/card views had `disableSort` still true for sizes/count columns which is no longer necessary

## Testing instructions

- [x] Review code
- [x] Check Actions build
- [x] In dg-download, test exact filters & the new download filter

### DIFF
--- a/packages/datagateway-common/src/api/index.tsx
+++ b/packages/datagateway-common/src/api/index.tsx
@@ -249,21 +249,29 @@ export const getApiParams = (
           );
         }
         if ('type' in filter && filter.type) {
-          if (filter.type === 'include') {
-            searchParams.append(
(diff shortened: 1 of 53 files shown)


joelvdavies (dismissed):
Worked well for me

louise-davies:
@joelvdavies sorry, added a cheeky extra commit as we noticed a another tiny issue when testing!

PR #1707, opened 2025-03-14
Fix slow size calculation of large carts

louise-davies:
## Description
react-query 3.33.6 introduces a performance issue with `useQueries`. I believe this will be resolved in v4, but for now just downgrade

Also mess around with p-limit stuff with the download cart item links - improves initial load performance

Also did some dependency fixing on here for security alerts to test that they work with the fix

## Testing instructions

- [ ] Review code
- [ ] Check Actions build
- [ ] You need a cart with over 4000 items in it to see the performance issue, but I have tested it on dgwdiamond-preprod with a cart of ~6500 datasets and this only takes ~30 mins to calculate instead of hours. Feel free to ask me and I can demo the difference


joshdimanteto (approved):
LGTM! I couldn't recreate the slowness locally, but this branch works well 👍. If there's a way to show the difference in a GIF or a quick demo, it would be useful for future reference.

louise-davies:
@joshdimanteto probably better off demoing as the cart has DLS data in that I can only see because I'm an admin so I don't want to be posting that publicly 😅

PR #1714, opened 2025-03-27
Restore visit button #1694

### LINKED ISSUE #1694: Add ability for priviledged users to restore a single visit
Requirements from meeting on 04/02/2025:
- Use same/similar dialogue to the normal download cart (i.e. select an IDS, give an email etc.)
- Place the button in the Visit Details Panel
- Only show the button when the user has permission (this will be an endpoint in dg-dl-api)
- Have an info message in the dialogue that explains the differences to the normal download process (i.e. the queueing & that it may take a while)

Later maybe requirement:
- Might need a metadata check to ensure people are restoring the correct visit. Actually, details panel has metadata so initially we'll leave it as it is - maybe add a warning/check if after testing we see accidental restores

Other requirements: (aka from the backend)
- There will be a special endpoint that these requests will go do instead of the normal submit cart endpoint

louise-davies:
## Description
Adds a button to the DLS visit details panel that checks if the user has permission to use the queue functionality and if so lets them submit a visit to the queue.

This required refactoring the downloadConfirmation code into dg-common, and making it slightly more generic so it serves the functionality of both submit cart and queue visit.

## Testing instructions

- [ ] Review code
- [ ] Check Actions build
- [ ] Allows priviledged users to queue visits for restoration
- [ ] Does not allow non-priviledged users to see the button
- [ ] Normal download cart works as normal

## Agile board tracking
Closes #1694


joshdimanteto (approved):
LGTM

PR #1715, opened 2025-03-28
Handle isDeleted and no prepared id cases in progress bars

louise-davies:
## Description

Patrick highlighted that we send requests for progress to the IDS when we have no prepared ID - which always returns 500. Kevin also noticed we display progress bars for deleted downloads. So this PR addresses both points.

Accurately type it so that `preparedId` can be `undefined` - and then don't send download progress requests for deleted or no prepared ID downloads. Also, handle deleted downloads the same as completed or expired downloads.

Also add additional handling for a new usecase brought up by Patrick's queueing code - if a download is `PAUSED` and has no `preparedId` that means it has been queued.

## Testing instructions

- [ ] Review code
- [ ] Check Actions build
- [ ] {more steps here}

## Agile board tracking
connect to #{issue number}

### DIFF
--- a/packages/datagateway-common/src/app.types.tsx
+++ b/packages/datagateway-common/src/app.types.tsx
@@ -266,7 +266,7 @@ export interface Download {
   isDeleted: boolean;
   isEmailSent: boolean;
   isTwoLevel: boolean;
-  preparedId: string;
+  preparedId?: string;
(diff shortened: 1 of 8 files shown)


joshdimanteto (approved):
LGTM

PR #1754, opened 2025-06-19
Add ability to redirect to investigations/datasets/datafiles using a single field value #1753

### LINKED ISSUE #1753: Add ability to redirect to a visit by supplying just the visit Id
See: https://stfc.atlassian.net/browse/DDS-69

But essentially, DLS want to be able to construct URLs to DataGateway using just the visitId as a parameter

louise-davies:
## Description
I've implemented #1753 via a generic redirect component. This means that if in the future if facilities want to be able to identify e.g. a Dataset by it's name, or redirect to a investigation by some other value (e.g. for ISIS a name redirect for an RB number redirect) then it is possible just by configuring the route config.

The URL is in the structure of `/redirect/:facilityName/:entityName/:entityField/:fieldValue`:
- `facilityName` is just to ensure we're building the right URLs for ISIS vs DLS
- `entityName` is currently just `investigation`, `dataset` or `datafile`
- `entityField` what it says on the tin. Note this is case sensitive so `visitId` has to be camelCase
- `fieldValue` the value to search for.

I had it redirect to the homepage with an error if the field is non-unique, but otherwise it works like the DOI redirect did. I can't really change the DOI redirect as that URL is different at the moment so we leave it for now but they share the same base component.

Also, I refactored a few other hooks to reuse this "try and find a single entity" as this was done for things like the details panels, so it reduces a bit of code.

## Testing instructions

- [ ] Review code
- [ ] Check Actions build
- [ ] Test out a few redirect links, including invalid ones/fields with non-unique values

## Agile board tracking
Closes #1753


joelvdavies (commented):
Tested with LILS/ISIS, worked well for me with id, fileSize, fileCount, visitId, erroring where they weren't unique/invalid. Nice that its generic.

PR #1773, opened 2025-07-23
Actually handle unauthorised route config

louise-davies:
## Description
We weren't ever actually passing the `unauthorised` option to the event dispatch function - so this fixes that. This didn't cause issues with DGW at first glace because it was tested with `autoLogin` on, but this would have affected e.g. DLS who don't have autoLogin at the moment. 

## Testing instructions

- [ ] Review code
- [ ] Check Actions build
- [ ] Test that homepage is visible in SGW even when autoLogin is off (or just compare this code to IMS and test it there)


joelvdavies (approved):
Worked for me 🥳

PR #1880, opened 2026-03-09
Migrate to React Query 5

louise-davies:
## Description
See title. Main breaking changes were the changes to `useInfiniteQuery`, the removal of `exact: false` to `setQueryData`/`getQueryData` and thus having to use `setQueriesData`/`getQueriesData`, removing onError handlers and instead using query meta and a global error handler i nthe query cache, some changes to `useMutation`, renaming `loading` state to `pending` and `isInitialLoading` to `isLoading`.

Plus I added the eslint plugin for react query so fixing a lot of things there, mostly because URLs were not in query keys. We could just ignore these errors, but that means we don't get the full power of the linter, and in most cases added them to the query key is fine. It makes finding exact query keys harder but I think I've worked around it sufficiently. Due to adding the linter I based this branch off of the eslint branch.

## Testing instructions

- [ ] Review code
- [ ] Check Actions build
- [ ] Give a good general test of e.g. infinite queries, mutations etc. as they're the ones with the most refactoring work done to them


joshdimanteto (approved):
LGTM, there are a couple of files where the line endings has changed but it's changing from windows to linux so I'm happy 😊

Image

PR #1896, opened 2026-03-25
Upgrade React Router #1110

### LINKED ISSUE #1110: Upgrade to react-router v6
**Description**:
react-router v6 has quite a few breaking changes which we need to refactor.

**Acceptance criteria**:
- [ ] All apps running on react-router v6

louise-davies:
## Description
Biggest refactors are:
- refactoring `datagateway-dataview` so that `PageRouting` is the "top level" component that gets imported via `App`, and `PageContainer` becomes a "layout route" that uses `Outlet` to render the actual route contents and `matchPath` to turn on/off different UI elements
- refactoring many tests as they can't use history anymore
- default sort/filtering broke in tests (but not in dev mode from what I could see?). Added an extra check to wait for the location to update with the default search params before performing the search. At this point it would likely easier to remove default sort/filtering functionality from `Table` and instead just link/route with the default params provided - but that's a larger refactor I didn't want to mix up in here

Otherwise just "standard" refactoring as according to the migration guides

## Testing instructions

- [x] Review code
- [x] Check Actions build
- [x] Test with SciGateway and check routing works

## Agile board tracking
Closes #1110


joshdimanteto (approved):
LGTM

PR #1942, opened 2026-05-27
Fix ismintable error handling to account for API changes

louise-davies:
## Description
See https://github.com/ral-facilities/icat-doi-minter-api/pull/137 for corresponding API changes.

But now, all ismintable "expected errors", aka ones telling us "no this isn't mintable" are now 403, and they have a different format which splits up the 3 different entity types and gives individual reasons per ID. We don't need the reasons, but need to handle the changed format to extract out the IDs and entity types to highlight the correct things.

## Testing instructions

- [x] Review code
- [x] Check Actions build
- [ ] datagateway-dseg has updated api, check that adding unmintable data highlights rows (should be easy to find any data with no session DOI, or that you're not the PI for)


joshdimanteto (approved):
LGTM

PR #1948, opened 2026-06-03
Improve UX of user generated DOIs

louise-davies:
## Description
- improve translations to use Trans components to allow line breaks & links
- add help text to creators & related items
- change generate DOI button name to review DOI metadata
- Fallback to user givenName + familyName or name if fullName is null
- add autoSelect prop to subject picker to select value on blur
- change form error checking, so now show the button as clickable, but once clicked highlight any errors
- make links open in new tabs

## Testing instructions

- [x] Review code
- [x] Check Actions build

### DIFF
--- a/packages/datagateway-download/cypress.config.ts
+++ b/packages/datagateway-download/cypress.config.ts
@@ -4,7 +4,7 @@ export default defineConfig({
   video: false,
   retries: {
     runMode: 3,
-    openMode: 1,
+    openMode: 0,
   },
   e2e: {
     setupNodeEvents(on, config) {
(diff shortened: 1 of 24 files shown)


packages/datagateway-download/cypress.config.ts
```diff
@@ -4,7 +4,7 @@ export default defineConfig({
   video: false,
   retries: {
     runMode: 3,
-    openMode: 1,
+    openMode: 0,
```
joshdimanteto, line 7: was this meant to be updated ??

joshdimanteto (commented):
LGTM, just minor one comment

packages/datagateway-download/cypress.config.ts
louise-davies: I personally prefer this to be set to `0` - so that automatic retries don't happen when developing tests (as often whilst developing a test it is definitely wrong and not flaky, and so repeating takes a while) so I'd prefer this to be set here (as I keep having to change it locally!)

joshdimanteto (approved):

PR #1967, opened 2026-07-01
Clear query caches when the user logs out

louise-davies:
## Description
Chris reported an issue where he was seeing the `WithIdCheck` "oops" page, even when he switched users. After some digging in the logs, it was likely that he was initially looking at the URL anonymously, which meant he saw the oops page as the anon user didn't have permission, and then when he switched to his user account the error persisted as the caches aren't cleared in that case. So, two fixes:

- ensure all query caches are cleared on sign out
- adjust the WithIdCheck oops page to clarify that a reason you could be seeing that message is you don't have permissions, and to check which user you are logged in as

## Testing instructions

- [x] Review code
- [x] Check Actions build

### DIFF
--- a/packages/datagateway-dataview/public/res/default.json
+++ b/packages/datagateway-dataview/public/res/default.json
@@ -304,7 +304,7 @@
   "loading": {
     "verifying": "Verifying URL",
     "oops": "Oops!",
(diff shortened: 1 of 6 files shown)


joshdimanteto (dismissed):
LGTM, there is just 2 failing snapshot tests

joshdimanteto (approved):

Issue #1956, opened 2026-06-10
Add Is PI button switch for My DOIs table


louise-davies:
Remove the existing option `minter`, and instead have a separate toggle for PI/Any which can be used to filter both user defined and session DOIs. Then we should probably reword/rename the remaining options in the my doi page to just be User-Defined/Session DOI and change the label wording to match what the All DOI table uses (this language will be simpler and more intuitive as well)

xkyue:
<img width="1198" height="138" alt="Image" src="https://github.com/user-attachments/assets/34dfa423-12d0-4bf3-a9d0-60facf996408" />

Hi @louise-davies for the PI/Any I wanted to clarify if you wanted only two column one for PI and one for all DOIs including the user, or three columns: PI, others, (then at the beginning default to all of their DOIs)?

PR #1970, opened 2026-07-15
Add Is PI button switch for My DOIs table

xkyue:
## Description
Remove the existing option minter, and instead have a separate toggle for PI/Any which can be used to filter both user defined and session DOIs. Then we should probably reword/rename the remaining options in the my doi page to just be User-Defined/Session DOI and change the label wording to match what the All DOI table uses (this language will be simpler and more intuitive as well)

## Testing instructions
- [ ] Review code
- [ ] Check Actions build

## Agile board tracking
closes to #1956

### DIFF
--- a/packages/datagateway-dataview/public/res/default.json
+++ b/packages/datagateway-dataview/public/res/default.json
@@ -281,25 +281,32 @@
     "all_roles": "All"
   },
   "my_doi_table": {
-    "type_button_group_aria_label": "Select DOIs where I",
-    "minter": "Minted the user-created DOI",
-    "user": "Am listed on the user-created DOI",
-    "session": "Am listed on the session DOI",
-    "open_button_group_aria_label": "Select Session DOIs which are",
-    "open_or_closed": "Open or Closed",
-    "open": "Open",
-    "closed": "Closed",
-    "all": "All"
-  },
-  "all_doi_table": {
-    "type_button_group_aria_label": "Select DOI type",
-    "user": "User-created DOIs",
+    "type_button_group_aria_label": "Select DOIs type",
+    "user": "User-defined DOIs",
     "session": "Session DOIs",
     "open_button_group_aria_label": "Select Session DOIs which are",
     "open_or_closed": "Open or Closed",
     "open": "Open",
     "closed": "Closed",
-    "all": "All"
+    "all": "All",
+    "pi_button_group_aria_label": "Select DOIs which roles are",
+    "pi_or_any": "All",
+    "pi": "Principal Investigator",
+    "any": "Others"
+  },
+  "all_doi_table": {
+    "type_button_group_aria_label": "Select DOI type",
+    "user": "User-defined DOIs",
+    "session": "Session DOIs",
+    "open_button_group_aria_label": "Select Session DOIs which are",
+    "open_or_closed": "Open or Closed",
+    "open": "Open",
+    "closed": "Closed",
+    "all": "All",
+    "pi_button_group_aria_label": "Select DOIs which roles are",    
+    "pi_or_any": "All",
+    "pi": "Principal Investigator",
+    "any": "Others"
   },
   "loading": {
     "verifying": "Verifying URL",
--- a/packages/datagateway-dataview/public/res/default.json
+++ b/packages/datagateway-dataview/public/res/default.json
@@ -281,7 +281,7 @@
     "all_roles": "All"
   },
   "my_doi_table": {
-    "type_button_group_aria_label": "Select DOIs type",
+    "type_button_group_aria_label": "Select DOI type",
     "user": "User-defined DOIs",
     "session": "Session DOIs",
     "open_button_group_aria_label": "Select Session DOIs which are",
@@ -302,11 +302,7 @@
     "open_or_closed": "Open or Closed",
     "open": "Open",
     "closed": "Closed",
-    "all": "All",
-    "pi_button_group_aria_label": "Select DOIs which roles are",    
-    "pi_or_any": "All",
-    "pi": "Principal Investigator",
-    "any": "Others"
+    "all": "All"
   },
   "loading": {
     "verifying": "Verifying URL",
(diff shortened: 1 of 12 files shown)


packages/datagateway-dataview/src/views/doiTypeSelector.component.tsx
```diff
@@ -139,6 +147,46 @@ const DOITypeSelector = (props: DOITypeSelectorProps): React.ReactElement => {
...
+                  : t('all_doi_table.pi')}
+              </ToggleButton>
+              <ToggleButton value={false} sx={{ p: '3px 7px' }}>
+                {type === 'myDOIs'
+                  ? t('my_doi_table.any')
+                  : t('all_doi_table.any')}
+              </ToggleButton>
+            </ToggleButtonGroup>
+          </Grid>
+        </Grid>
```
louise-davies: I think we should only have the PI buttons on the `myDOIs` table as that's the one where a user looking for their own data should go to, and the `allDOIs` table is a more "public-facing" table where we don't expect users to be looking for their own stuff there

packages/datagateway-dataview/public/res/default.json
```diff
@@ -281,25 +281,32 @@
     "all_roles": "All"
   },
   "my_doi_table": {
-    "type_button_group_aria_label": "Select DOIs where I",
-    "minter": "Minted the user-created DOI",
-    "user": "Am listed on the user-created DOI",
-    "session": "Am listed on the session DOI",
+    "type_button_group_aria_label": "Select DOIs type",
```
louise-davies:
```suggestion
    "type_button_group_aria_label": "Select DOI type",
```

packages/datagateway-dataview/src/views/table/dls/dlsDOITables.component.tsx
```diff
@@ -261,5 +259,36 @@ export const DLSAllDOIsTable = (): React.ReactElement => {
...
+        }),
+      },
+      {
+        filterType: 'where',
+        filterValue: JSON.stringify({
+          'users.user.name': { eq: username },
+        }),
+      }
+    );
+
```
louise-davies: Delete this code as this is never used in the All DOIs table - this will also resolve the remaining test coverage issue

packages/datagateway-dataview/public/res/default.json
```diff
@@ -281,19 +281,22 @@
...
   },
   "my_doi_table": {
-    "type_button_group_aria_label": "Select DOIs where I",
-    "minter": "Minted the user-created DOI",
-    "user": "Am listed on the user-created DOI",
-    "session": "Am listed on the session DOI",
+    "type_button_group_aria_label": "Select DOI type",
+    "user": "User-defined DOIs",
+    "session": "Session DOIs",
     "open_button_group_aria_label": "Select Session DOIs which are",
```
louise-davies:
```suggestion
    "open_button_group_aria_label": "Select DOIs which are",
```

packages/datagateway-dataview/public/res/default.json
```diff
@@ -281,19 +281,22 @@
...
     "open_or_closed": "Open or Closed",
     "open": "Open",
     "closed": "Closed",
-    "all": "All"
+    "all": "All",
+    "pi_button_group_aria_label": "Select DOIs which roles are",
+    "pi_or_any": "All",
+    "pi": "Principal Investigator",
+    "any": "Others"
```
louise-davies: `any` is not really an accurate descriptor - it's more `other` and `pi_or_other` right?

louise-davies: I think I would prefer different labels here as the current ones don't read clearly. Maybe:

```
"pi_button_group_aria_label": "Select DOIs where I am",
"pi_or_any": "Any role",
"pi": "Principal Investigator",
"any": "Other role"
```

DLS may want to change this further which I can change in the translation files, but I think the above is a better "default"

packages/datagateway-dataview/src/views/doiTypeSelector.component.tsx
```diff
@@ -139,6 +147,39 @@ const DOITypeSelector = (props: DOITypeSelectorProps): React.ReactElement => {
           </Grid>
         </Grid>
       )}
+
+      {type === 'myDOIs' &&
+        (doiType === null ||
+          doiType.view === 'all' ||
+          doiType.view === 'session' ||
+          doiType.view === 'user') && (
```
louise-davies: We can get rid of these conditionals checking the `view` type as these buttons are displayed on every view on the `myDOIs` table right?

packages/datagateway-dataview/src/views/doiTypeSelector.component.tsx
```diff
@@ -139,6 +147,39 @@ const DOITypeSelector = (props: DOITypeSelectorProps): React.ReactElement => {
           </Grid>
         </Grid>
       )}
+
```
louise-davies: Also, I think it would be better to move the code for the PI button above the open/closed buttons - as currently in the `myDOIs` table, selecting user-defined DOIs removes the open/closed buttons, which as a consequence moves the PI buttons. Whereas if the PI buttons are first, then the appearance/disappearance of the open/close buttons does not affect the positioning of the PI buttons.

packages/datagateway-dataview/public/res/default.json
```diff
@@ -281,21 +281,24 @@
...
+    "type_button_group_aria_label": "Select DOI type",
+    "user": "User-defined DOIs",
+    "session": "Session DOIs",
+    "open_button_group_aria_label": "Select DOIs which are",
     "open_or_closed": "Open or Closed",
     "open": "Open",
     "closed": "Closed",
-    "all": "All"
+    "all": "All",
+    "pi_button_group_aria_label": "Select DOIs which I am",
```
louise-davies:
```suggestion
    "pi_button_group_aria_label": "Select DOIs where I am",
```

louise-davies (approved):
Looks good to me (LGTM) - works well and my comments have been addressed

PR #1980, opened 2026-07-23
Fixed Datetime parameter formatting

### LINKED ISSUE #1972: Fix Datetime parameter formatting
Sam Hough has started populating the `noticeSent` parameter, which when you try and view in a details panel causes a crash due to the date formatting not working (e.g. see https://datagateway-preprod.diamond.ac.uk/browse/proposal/MG44507/investigation?sort=%7B%22startDate%22%3A%22desc%22%7D, and click on the dropdown arrow next to the visit name)

The relevant code is located here: https://github.com/ral-facilities/datagateway/blob/d87046a56764d59f36f127a2ce50bae20d9ad42b/packages/datagateway-common/src/detailsPanels/dls/visitDetailsPanel.component.tsx#L88

Sam Hough's parameter has sub-second precision (e.g. `2026-07-15 13:43:42.309000+01:00`) which the parsing code doesn't expect. It would need to be something like `'yyyy-MM-dd HH:mm:ss.SSSSSSXXX'`

We should probably try and parse both with and without sub-second information. `date-fns` has an `isMatch` function which tests if a string matches a format string, which can be used to determine which format string to use before running the `parse`

Additionally, we should probably fix that it throws an error - we can use the `isValid` function from `date-fns` to test if it's a valid date before passing to the `format` function, e.g. `isValid(date) ? format(date, 'yyyy-MM-dd') : date.toString()` (`date.toString()` will return `Invalid Date`)

xkyue:
## Description
Fixed Datetime parameter formatting, try and parse both with and without sub-second information

## Testing instructions
- [ ] Review code
- [ ] Check Actions build

## Agile board tracking
closes #1972

### DIFF
--- a/packages/datagateway-common/src/detailsPanels/dls/visitDetailsPanel.component.tsx
+++ b/packages/datagateway-common/src/detailsPanels/dls/visitDetailsPanel.component.tsx
@@ -1,5 +1,5 @@
 import { Divider, Grid, styled, Tab, Tabs, Typography } from '@mui/material';
-import { format, parse } from 'date-fns';
+import { format, isValid, isMatch, parse } from 'date-fns';
 import React from 'react';
 import { useTranslation } from 'react-i18next';
 import { useDispatch, useSelector } from 'react-redux';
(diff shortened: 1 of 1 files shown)


louise-davies (approved):
LGTM

